PR #13, opened 2016-10-21
Remove duplicate emit calls and fix event handler function declaratio…

### LINKED ISSUE #12: Tweaks for collapse to work.
I was able to get collapse working with the following tweaks:
In collapse.vue change:
`hub.$on('toggled::collapse', function (data) {`
to:
`hub.$on('toggled::collapse', (data) => {`

That way this.id returns the expected value.

Also in collapse-toggle.vue there are two calls to emit the "toggled::collapse" events.  I had to remove one of the calls.

Similar changes will be needed to the analogous accordion code that seems to be stubbed in.

michaelwoodson:
…n for collapse, fixes #12.


pi0:
Thanks

PR #26, opened 2016-11-18
Export individual components

nddery:
This would allow us to import individual components instead of "using" the plugin on the main `Vue` instance.

So, in addition of being able to use the whole plugin like before:

```js
// index.js
import Vue from 'vue'
import BootstrapVue from 'bootstrap-vue'

Vue.use('BootstrapVue')
```
We could also instead import and use components individually.

```js
// components/Hello.vue
<template>
  <b-button-group vertical>
    <b-button>Left</b-button>
    <b-button>Middle</b-button>
    <b-button>Right</b-button>
  </b-button-group>
</template>

<script>
import { bButtonGroup, bButton } from 'bootstrap-vue'

export default {
  components: {
    bButtonGroup,
    bButton
  }
}
</script>
```

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -95,4 +95,5 @@ if (typeof window !== 'undefined') {
   //require('../styles/style.css');
 }
 
-module.exports = plugin;
+module.exports = components;
+module.exports.default = plugin;


pi0:
@nddery Thanks :)

nddery:
@pi0 any idea when we could expect an update to npm ?

sourcec0de:
@nddery @pi0 

This implementation breaks the current API and results in `Uncaught TypeError: plugin.apply is not a function(…)`

```js
import Vue from 'vue'
import BootstrapVue from 'bootstrap-vue'

Vue.use(BootstrapVue);
```

To make it work users will have to call it like this now.

```js
import Vue from 'vue'
import BootstrapVue from 'bootstrap-vue'

Vue.use(BootstrapVue.default);
```

pi0:
@sourcec0de @nddery would you please check 94794f188d5a3e59674017f29c3017a85af3e7dc (v0.3.8) it may possibly fix both issues.

sourcec0de:
@pi0 thats essentially the same thing as before.

The only way this will work as expected is if you are exporting using this syntax.

```js
export {bBtn, bTabs};
export default plugin;
```

This should created named exports for the individual components.
With the addition of a fallback export for the plugin.

Then when they are imported it would look something like this.
```js
import { bBtn } from 'bootstrap-vue'; // this will pull named export
import BootstrapVue from 'bootstrap-vue'; // this will pull default export
```

pi0:
@sourcec0de Would you please make a PR?

sourcec0de:
@pi0 yea, I'm headed to bed for this evening but I'll submit one tomorrow morning.

~ Thanks for the hard work btw

sourcec0de:
@pi0 @nddery created a pull request that implements what I mentioned above

PR #27, opened 2016-11-20
Fixed default export of plugin

nddery:
Fixes the issues reported in https://github.com/bootstrap-vue/bootstrap-vue/pull/26#issuecomment-261690614.

Moved all components export into it's own `components.js` file because it was a lot of duplication otherwise, not the cleanest solution (having a file that just import/export) but it works.

@pi0 @sourcec0de this would result in `v0.3.9` if your good with these changes, or do you have another approach to suggest ?

### DIFF
--- a/components/components.js
+++ b/components/components.js
@@ -1,0 +1,74 @@
+import bAlert from './alert'
+import bBreadcrumb from './breadcrumb'
+import bButtonCheckbox from './button-checkbox'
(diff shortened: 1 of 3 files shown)


sourcec0de:
@nddery this looks great

pi0:
LGTM

nddery:
Just released `v.0.3.9`, sorry for the troubles and thanks!

Issue #34, opened 2016-12-04
tabs/tab: Uncaught TypeError: Cannot create property 'true' on string 'active'


MehrCurry:
After Upgrading to 0.3.10 i get execptions when i click a tab to activate it:

```
TypeError: Cannot create property 'true' on string 'active'
    at VueComponent.set$1 [as $set] (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:825:14)
    at VueComponent.setActive (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:71:1), <anonymous>:1:28372)
    at VueComponent.boundFn [as setActive] (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:126:14)
    at VueComponent.mounted (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:71:1), <anonymous>:1:28596)
    at callHook (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:2479:19)
    at Object.insert (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:2671:5)
    at invokeInsertHook (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:4326:28)
    at Vue$3.patch [as __patch__] (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:4473:5)
    at Vue$3.Vue._update (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:2359:19)
    at Vue$3.eval (eval at <anonymous> (http://localhost:63342/hello/dist/build.js:84:1), <anonymous>:2333:10)
```

JanBurp:
+1

pi0:
@MehrCurry @JanBurp I'll investigate problem soon. thanks for reports.

mathlet0x:
@MehrCurry @JanBurp this issue has been solved, i close it.

JanBurp:
thanks

PR #37, opened 2016-12-07
Fixed issues on tabs/tab #34

mathlet0x:
[https://vuejs.org/v2/guide/migration.html#vm-set-changed](https://vuejs.org/v2/guide/migration.html#vm-set-changed)

I also moved `active` from props to data because : 

> [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "active" (found in component <bTab>)

### DIFF
--- a/components/tab.vue
+++ b/components/tab.vue
@@ -11,6 +11,7 @@ export default {
       return {
         fade: this.$parent.fade,
         animate: false,
+        active: false
       }
     },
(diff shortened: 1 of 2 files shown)


pi0:
@mathlet0x Awesome 👍

PR #42, opened 2016-12-13
Fixed b-button-radio issues

mathlet0x:
I fixed `:class` issue, `btnVariant` and `btnSize` were not interpreted. 

I also moved `model` into data because : 
>[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value.


pi0:
@mathlet0x Thanks dude 👍

Issue #24, opened 2016-11-13
Unable to get various components to work


irelandpaul:
I just copy and pasted your examples. I could not get Alert, Tabs, or Dropdown to work. Alert says propty or method "shoVariable" and "state" are not defined on the instances but referenced during render. Tabs just comes up as a straight line. Dropdown does nothing and gives no feedback.

Accordion, Jumbotron, and a few others worked. For accordion, how can one start out expanded?

KarimJedda:
I agree, a lot seems not to work with Vue 2...

pi0:
@irelandpaul @KarimJedda Would you please confirm including bootstrap4 css in your templates? it is not bundled with bootstrap-vue and should be imported.

KarimJedda:
I confirm including it, but it did not work. 
Console errors and components are not rendered.

pi0:
Changes planned in upcoming #50 (0.5.0) release, expect everything working again :)

brandonlamb:
👍 i also have been trying to find a VueJS 2.0 + Bootstrap integration library, the others seem to be for VueJS 1.0 and dont work, some are for bootstrap 3.x. I also get errors when trying to copy/paste the example and use the dropdown.

```
vue.common.js?e881:519 [Vue warn]: Property or method "arrow" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option. 
(found in component <app-nav> at /Users/brlamb/Documents/src/spa-dashboard/src/components/app-nav.vue)
warn @ vue.common.js?e881:519
warnNonPresent @ vue.common.js?e881:1434
get @ vue.common.js?e881:1475
render @ app-nav.vue?4db4:26
Vue._render @ vue.common.js?e881:2952
(anonymous) @ vue.common.js?e881:2189
get @ vue.common.js?e881:1652
Watcher @ vue.common.js?e881:1644
Vue._mount @ vue.common.js?e881:2188
Vue$3.$mount @ vue.common.js?e881:5978
Vue$3.$mount @ vue.common.js?e881:8305
init @ vue.common.js?e881:2500
createComponent @ vue.common.js?e881:4052
createElm @ vue.common.js?e881:3995
createChildren @ vue.common.js?e881:4103
createElm @ vue.common.js?e881:4028
createChildren @ vue.common.js?e881:4103
createElm @ vue.common.js?e881:4028
createChildren @ vue.common.js?e881:4103
createElm @ vue.common.js?e881:4028
[...]

PR #46, opened 2016-12-16
Fixed dropdown issue #24

mathlet0x:
I fixed the dropdown issue and i adapted the template to the actual bootstrap 4 documentation.

Before merging, my question is about `$emit` methods in this component, i'm not really sure that they are usefull here, but if i'm wrong, please tell me why and i'll adapt my merge request.

Edit: If this pull request is valid, i'll update dropdows documention to : 

`
<b-dropdown text="action" size="md" variant="primary" :caret="true">
	<a class="dropdown-item" href="#">Action</a>
	<a class="dropdown-item" href="#">Another action</a>
	<a class="dropdown-item" href="#">Something else here</a>
	<div class="dropdown-divider"></div>
	<a class="dropdown-item" href="#">Separated link</a>
	</b-dropdown>
`

Thanks by advance

### DIFF
--- a/components/dropdown.vue
+++ b/components/dropdown.vue
@@ -62,8 +62,23 @@
     },
     methods: {
       toggle(e) {
+        // hide an alert
         this.show = !this.show;
+        // Dispatch an event from the current vm that propagates all the way up to its $root
+        if (this.show) {
+          this.$root.$emit('shown::dropdown');
+          e.stopPropagation()
+        } else {
+          this.$root.$emit('hidden::dropdown');
+        }
       }
     },
+    created: function () {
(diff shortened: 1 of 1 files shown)


mathlet0x:
@pi0 any feedback ?

nddery:
I don't have much time right now to look into this but I would tend to say that we should stick as close as what Bootstrap does. So regarding emitting events when the dropdown open/close, I'd tend to keep them ?

mathlet0x:
Okay, so i restore them, i just wonder why emit an event without specify which component is used.

So let's merge it, thanks for your feedback ;)

PR #62, opened 2017-01-06
support for vue router link

qwexvf:
Hi i modified the onclick method to use router with.

this.$emit('click', this.link)

is this for debugging?

### DIFF
--- a/components/nav-item.vue
+++ b/components/nav-item.vue
@@ -31,6 +31,11 @@
         methods: {
             onclick: function () {
                 this.$emit('click', this.link);
+                try {
+                    this.$router.push(this.link);
+                } catch (e) {
+                    console.log(e);
+                }
             }
         }
     }


pi0:
@qwexvf Thanks for your contribution :)
I totally agree with that is needed. But we cannot consider every user has and wants using `vue-router`, maybe we can add something like `useRouter` prop to let user enable/disable this feature. Also please note that if we don't use `<router-link>` tags we are missing lot's of things like adding `active` class to current route. Isn't it better to conditionally use `<router-link>` on update when `useRouter` is active ?

qwexvf:
@pi0 thank you!
i agree that we need to use <router-link> tag i totally forgot about it. i think we don't need that change! sorry about this :(

pi0:
@qwexvf  Thanks you back dude :) I'll keep this PR so we can make further works on that.

pi0:
@qwexvf support for vue router link is now partially available. i'll close this PR for housekeeping. Really thanks for your contributions and idea.

PR #66, opened 2017-01-10
Added ability to choose form layout for input and select

asxcandrew:
Now user can choose between horizontal and vertical form layouts
Like in this article [http://stacktips.com/tutorials/bootstrap/vertical-horizontal-and-inline-form-example-in-bootstrap](url)


pi0:
@asxcandrew nice!

PR #67, opened 2017-01-10
Popover re-design

SirLamer:
This is a proposed major re-write for Popover. This makes breaking changes but given that it seems there has been little work on it since early on, it seemed worthwhile.

I tried to adopt as much of the official Bootstrap functionality as makes sense in a Vue framework context. This includes making changes to some of the property names to align with Bootstrap.

Previously, if both a property and a slot of the same name (either 'title' or 'content') is provided, then displaying the slot takes priority. This is more consistent with typical default VueJS behavior.

External control of the popover is provided, and appropriate events are provided parent code to respond to changes.

Determination of how to respond to events is simplified. When multiple event types are being monitored, a simple 'debounce' check is employed to make sure two quick switches don't occur in succession. In general, it's better to just rely on one trigger type except for 'click' and 'focus' which work pretty well together.

### DIFF
--- a/components/popover.vue
+++ b/components/popover.vue
@@ -205,7 +205,7 @@
 
       // display popover if prop is set on load
       if (this.showState) {
-        this.showPopup();
+        this.showPopover();
       }
     },
 
--- a/components/popover.vue
+++ b/components/popover.vue
@@ -23,16 +23,6 @@
 
   export default {
     replace: true,
-    data() {
-      return {
-        show: false,
-      }
-    },
-    computed: {
-      popoverAlignment() {
-        return !this.position || this.position === `default` ? `popover-top` : `popover-${this.position}`
-      }
-    },
     props: {
       position: {
         type: String,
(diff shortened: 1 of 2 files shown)


pi0:
@SirLamer Locally tested your PR & seems stable.Thanks:) Any chance adding simple docs ? So we can officially add this component to list of supported ones.

mathlet0x:
@pi0 do you use a specific tool to manage your documentation like before ?

pi0:
@mathlet0x Docs are simply nuxt pages powered by vue it self :)) Please see [components docs](https://github.com/bootstrap-vue/bootstrap-vue/tree/master/docs/pages/docs/components). they will be published statically by CI as soon as you push changes to `master`.

SirLamer:
Yes I will prepare a documentation page. I didn't want to assume that you'd want to add this to documentation right away.

SirLamer:
For what it's worth, I am using this popover component as the base for an inline confirmation dialogue feature meant to be wrapped around buttons, to confirm actions as intended before executing.

Link to the repository is here, though I just set up the project template today so the component is not available from here yet:
https://github.com/SirLamer/click-confirmation

SirLamer:
I can't figure out how to get this Nuxt thing to work. I also installed and used Yarn for the first time, I think that worked correctly and in fact I will probably adopt Yarn as well.

I am developing in Windows 10 and using Git Bash as a console. Npm, Yarn and Node are installed. "yarn install" has been run.

I get this error while test-running against the stock documents as they currently exist in the package, without any changes. On the surface it seems that Object.keys() is not available, which is called early on within index.js as well as docs\layouts\components.vue. I suspect there are bigger issues, it just happens that Objects.keys is called early within these files.

```
Vue.js error

ReferenceError: _Object$keys is not defined
    at Object.module.exports.Object.defineProperty.value (__vue_ssr_bundle__:2142:1)
    at __webpack_require__ (__vue_ssr_bundle__:21:30)
    at Object.<anonymous> (__vue_ssr_bundle__:2052:9)
    at __webpack_require__ (__vue_ssr_bundle__:21:30)
    at Object.module.exports.Object.defineProperty.value (__vue_ssr_bundle__:1216:1)
    at __webpack_require__ (__vue_ssr_bundle__:21:30)
    at Object.<anonymous> (__vue_ssr_bundle__:30392:14)
    at __webpack_require__ (__vue_ssr_bundle__:21:30)
    at module.exports.__vue_styles__ (__vue_ssr_bundle__:67:18)
    at Object.<anonymous> (__vue_ssr_bundle__:70:10)
```

SirLamer:
I have prepared a documentation page but I cannot test until the above Nuxt dev thing is fixed so I am just sitting on it.

pi0:
@SirLamer Would you please commit your works on this pr branch? so i can work on that problem

SirLamer:
Sure, note I was test-running the docs-dev script on your code without having added anything for Popover and I am getting exactly that error. To me it seems something like Nuxt doesn't have access to ES5 methods. I suspect it's something to do with my operating environment and not your code but, still, I'm stuck.

I will push my docs changes and will keep on it as well. I might try to run it in a Linux VM environment, though that should not be necessary.

SirLamer:
Note I took some liberties and cleaned up Readme.md as well just because I saw value in it.

pi0:
Ok, i'll merge this so users can use it earlier we may add docs later as it is taking too long :))

SirLamer:
Okay, well don't forget I did prepare a docs page for popover, it has to be added to the nav and then tested, though.

PR #69, opened 2017-01-12
Added active status handling for collapse-trigger

Michelemassari:
Hi, I needed an "active" class toggle when the trigger for collapse is clicked.

Thought it may be helpful to keep track of open/close status there as well as in the actual collapsible item.


pi0:
@Michelemassari Thanks

PR #70, opened 2017-01-13
Move production dependencies out of devDependencies

SirLamer:
As far as I understand, the Vue, Tether and Bootstrap dependencies should be in the "dependencies" group, not the "devDependencies" group. The dev group is meant for dependencies required only when developing on THIS package. When another package pulls this one in to extend it or use it in its own project, as far as this package is concerned that is a production environment. If these compilation-required dependencies are in the "dev" packages group then NPM or Yarn will not pull them down to the parent project as required and expected. In my particular experience, I found I had to add Tether as a dependency in my own packages.json, which shouldn't be required.

I may misunderstand your intention here, and I'm comparatively novice so feel to ignore this if you disagree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,19 +73,21 @@
       ]
     ]
   },
+  "dependencies": {
+    "bootstrap": "^4.0.0-alpha.5",
+    "tether": "latest",
+    "vue": "2.1.7"
+  },
   "devDependencies": {
     "babel-core": "latest",
     "babel-loader": "latest",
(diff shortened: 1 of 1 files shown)


pi0:
@SirLamer The main intention was that sometimes developers may want to use their own packages, for example at fandogh we have our ready to use UI packages and don't directly require and recompile bootstrap stuff.  Anyway i agree this would add a lot of user ease of use with little cost. :)

PR #71, opened 2017-01-13
Remove component page links which don't work with Vue routing

SirLamer:
I tried to be clever and add links to the component pages on the readme, however with Vue handling routing it does not work. I don't know if Github allows this but they could be made to work if the server knows not to handle the page request past /docs/components/.


pi0:
That's ok. i think it should be a problem with vue router

SirLamer:
Okay I trust you'll fix it then, right now none of the links work.

PR #74, opened 2017-01-17
Add modal id to shown::modal and hidden::modal event payload

xiaoyufan:
I've used multiple modals in one page and need listen to their shown and hidden events. Wondering if the specific modal id could be added to payload so that I can respond to these events dispatched from different modals respectively.

Thanks for your time.


pi0:
Thanks :)

PR #83, opened 2017-01-26
Fix carousel indicators

antonstefanov:
Carousel indicators are not updating properly, because the same variable "index" is used in the loop and in the data property


pi0:
Thanks :)

PR #97, opened 2017-02-09
Format form input on change

riker09:
When formatting on input event, the input field loses focus and makes it really difficult to enter the desired value. Formatting on change feels much more natural.

### DIFF
--- a/components/form-input.vue
+++ b/components/form-input.vue
@@ -8,6 +8,7 @@
             :placeholder="placeholder"
             :value="value"
             @input="onInput($event.target.value)"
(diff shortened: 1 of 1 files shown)


riker09:
I just noticed that with this patch the formatting does not initially kick in. The input event is triggered. Is there an easy way to trigger the change event as well? After all, the form is just being populated with values, so the input field value is technically changing from "" to "WHATEVER".

pi0:
@riker09 That's true. I've added `lazyFormatter` option so we can choose best behavior :)

Issue #110, opened 2017-02-17
Popover does not work in IE 11


javiercbk:
Popover uses the `includes` array method which is not supported in IE 11.

For browser compatibility it should be replaced by `indexOf`

javiercbk:
I might be able to make a PR by monday, is it good enough?

SirLamer:
Hi, I wrote this one so I'll address it.

pi0 - Is the base project compiler configured to perform a polyfill step? Or is ES6 not allowed in the project? Note I am aware (as of today) that Array.prototype.includes is from ES2016 and probably should be avoided in either case.

SirLamer:
Fixed, pi0 when decide when to publish the next release but in the meantime if you want, you could use this specific commit instead: https://github.com/bootstrap-vue/bootstrap-vue/commit/d1c75acec68dd403a625de654f0c375075991554

PR #111, opened 2017-02-18
Replace Array.prototype.includes with Array.prototype.indexOf

SirLamer:
Fixes a browser compatibility issue as suggested here: https://github.com/bootstrap-vue/bootstrap-vue/issues/110

Let's take this moment to remind ourselves of how Microsoft set the Internet back by a decade because they keep insisting on integrating their OS functionality with the web browser. Bravo!


pi0:
@SirLamer  Thanks :) 0.8.1 is released

PR #116, opened 2017-02-23
Add option to get class onto td of table

PatrickEriksen:
As an extension of #115 we needed to be able to access the `<td>`, so this uses the `field.class`.

Just a quick question, potentially this should have been a ternary that checks if field.class exists? Or is this simplistic method ok?

### DIFF
--- a/components/table.vue
+++ b/components/table.vue
@@ -12,7 +12,7 @@
         </thead>
         <tbody>
         <tr v-for="item in _items" :key="items_key" :class="[item.state?'table-'+item.state:null]">
-            <td v-for="(field,key) in fields">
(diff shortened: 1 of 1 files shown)


riker09:
@PatrickEriksen Yes, I believe you are right. This should check if field.class exists before hand. It didn't throw any error when I tested this. :blush:

PatrickEriksen:
@riker09 cool - made both of the field.class outputs ternary checks.

It works with both scenarios in my environment as well, so I'm still not 100% convinced its necessary. Would appreciate you giving it a test in your environment as well.

Otherwise, this should be good to go.

PR #119, opened 2017-02-26
[table component] Pass current row index to slot

riker09:
When rendering a row with the table component the current index is now available in `props.index`


pi0:
Thanks :)

PR #124, opened 2017-03-01
[Tabs component] updates

dleesalestreamsoft:
made the tabs be able to be set using :value and it will emit a input that can be used for vuex and vue-router


pi0:
@dleesalestreamsoft Really thanks :) Just please fix linter errors (`npm run fix` will fix most of them).

dleesalestreamsoft:
@pi0 when i tried running xo --fix i got so many lint errors that it is complaining about.  not even in just the tabs/tab file. so I am guessing you something is different about the configuration of xo.

pi0:
@dleesalestreamsoft ok :) i'll fix them later

PR #127, opened 2017-03-03
href prop isnt needed if object in to prop

jjonas-gcs:
If you use an object in to prop without using the href prop (for example if you use named routes) you will get href="[object Object]" in your html.
My pull request will fix this.


pi0:
@jjonas-gcs  thnks :)

Issue #98, opened 2017-02-09
Nested pagination on table component does not working when importing only table component


thomasfowler:
I think this is a bug but I'm not 100% sure.....

What we are doing is importing each component individually on each of our Vue templates instead of a global import on, eg. main.js.

By way of example, using the table example in the docs:

```javascript
<template>
  <div id="devTable">
    <b-table :items="items" :fields="fields" pagination :perPage="3">

      <template slot="name" scope="item">
        <div style="font-weight:600">{{item.value.first}} {{item.value.last}}</div>
      </template>
      
      <template slot="actions" scope="item">
        <b-btn size="sm" @click="details(item.item)">Details</b-btn>
      </template>

    </b-table>
  </div>
</template>

<script lang="js">
  import bButton from 'bootstrap-vue/components/button.vue'
  import bTable from 'bootstrap-vue/components/table.vue'

  export default {
    name: 'demoTable',
    components: {
      bBtn: bButton,
      bTable: bTable
    },
    data () {
      return {
        items: [
          {
            'isActive': true,
            'age': 40,
            'name': {
              'first': 'Dickerson',
              'last': 'Macdonald'
            }
          }
        ],
        fields: {
          name: {label: 'Person Full name'},
          age: {label: 'Person age'},
          isActive: {label: 'is Active'},
          actions: {label: 'Actions'}
        }
      }
    }
  }
</script>
<style lang="css">
  #devTable{
    width: 80%
  }
</style>
<style lang="scss">
  @import "../../node_modules/bootstrap/scss/buttons";
  @import "../../node_modules/bootstrap/scss/tables";
</style>
```

So, if we import the two components used individually, namely button and table, i get the following two errors on the javascript console:

```
[Vue warn]: Unknown custom element: <b-pagination> - did you register the component correctly? For recursive components, make sure to provide the "name" option. 
(found in component <b-table> at /blah/blah/node_modules/bootstrap-vue/components/table.vue)
```

and

```
[Vue warn]: v-model is not supported on element type: <b-pagination>. If you are working with contenteditable, it's recommended to wrap a library dedicated for that purpose inside a custom component. 
(found in component <b-table> at /blah/blah/node_modules/bootstrap-vue/components/table.vue)
```

So then, my assumption was that I simply needed to manually import the dependent component. Also, based on what is in `index.js` in the components dir, I made sure I was importing in the same order (in case there was some kind of import order dependency going on... there doesn't seem to be....):

```javascript
<template>
[...]

PR #136, opened 2017-03-09
added explicit import of button to dropdown.vue

PatrickEriksen:
another instance where we need to import the dependent components directly. similar to #99 and #98


pi0:
Thanks @PatricEriken

PR #137, opened 2017-03-09
Visual tabs bug fixed

nwerr:
- Fixed visual bug with button class
- Changed HTML tags to the offical bootstrap syntax
- Removed size property
- Added the possibility to render small tabs


pi0:
Thanks @nwerr :)

PR #139, opened 2017-03-10
fix dropdown.vue import button issue

storyn26383:
Fix dropdown.vue import button issue.


pi0:
@storyn26383 Eh Thanks :)

PR #142, opened 2017-03-13
Update modal.vue 

mosinve:
Add .bind(this) to hub.$on event listeners because without this modification at the moment of receving event "this" points to root element instead of current.

If I'm wrong, tell me how to use modal without his mod, pls :)

### DIFF
--- a/components/modal.vue
+++ b/components/modal.vue
@@ -110,8 +110,8 @@ export default {
     },
     created() {
         const hub = this.$root;
-        hub.$on('show::modal', id => id === this.id && this.show());
-        hub.$on('hide::modal', id => id === this.id && this.hide());
+        hub.$on('show::modal', id => id === this.id && this.show()).bind(this);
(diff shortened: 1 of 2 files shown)


components/modal.vue
```diff
@@ -110,8 +110,8 @@ export default {
     },
     created() {
         const hub = this.$root;
```
pi0: Maybe we can your slef trick :)) Something like this: 
```js
 const self=this;
 hub.$on('show::modal', id => id === self.id && self.show());
``

mosinve: but it's the same as usind .bind(this) and with extra var

pi0:
@mosinve Thanks for your contribution :) please say when your work was done on this component and i will merge it soon :)

mosinve:
It' works now) at least at this point))) 
Somehow another commit on dropdown added to this PR....

PR #154, opened 2017-03-20
Navbar styles fix

mosinve:
- updated "fixed-" styles according actual Bootstrap stylesheet
- added sticky prop to toggle "sticky-top" class
- removed full prop as "navbar-full" class does not exist in Bootstrap 4 stylesheet


pi0:
Thanks :)

PR #163, opened 2017-03-22
Popover - react to 'content' and 'title' property changes

SirLamer:
This is an attempt to make the positioning and sizing reactive to content changes. I chose to hook the property inputs only in order to use built-in Vue watchers. Vue does not natively support hooking changes to slots so for now it is not going to respond to slot changes until the user scrolls. I think this is okay, as the property inputs will accept HTML when desired.

In truth, this is such a niche issue that I don't want to bloat the package with a MutationObserver polyfill in order to use native DOM code to watch for changes in general.


SirLamer:
I have not self-committed this PR because I am interested in feedback on the implementation decisions.

pi0:
@SirLamer Positive improvement at least we can immediately respond to content changes. Anyway besides of this i think Using [Vue.nexttick](https://vuejs.org/v2/api/#Vue-nextTick) (runs after dom updates) and a simple Interval timer during show state would resolve almost all conditions.

pi0:
@SirLamer  Can we just merge it and having more improvements later?

SirLamer:
I will add the nextTick. I'm not sure about the timer, though... after all, Tether itself also self-updates every time the page is scrolled.

pi0:
Check ([this](https://jsfiddle.net/pi0/Lweyrv47/) ) fiddle. Tether won't be noticed of style changes (also all vue css transitions)

SirLamer:
In JSfiddle you can't scroll however if you were to drag the window border back and forth, you'll affect the same thing and see the tooltip follow the button.

By time interval I thought you were thinking every 2 seconds or something. We really can't have all of the popovers spam the position() method at 120hz because someone might make a button animate itself, it would be a massive waste of resources. We could maybe provide programmatic access so that such a program could command position() to run.

pi0:
> We could maybe provide programmatic access so that such a program could command position() to run.

It's reasonable :) 

Anyway we don't spam at all by timers . Only popovers in visible state(most of time they are invisible or at most one of them is visible)

PR #165, opened 2017-03-22
Added maxwidth property to popover

BramVer:
I used this library to show small forms on the click of a button and
overwriting the max-width that was set via CSS does not work for me, 
so I forked the repo and added these small changes. 

The only change is a small property 'maxwidth' that accepts the values 'v-sm', 'sm', 'md', 'lg' and 'v-lg'.
It's still only the max-width that changes and not the width itself and the default is still set to 276px, as per constraint, so the impact on users who do not want anything changed, will be nonexistant.

### DIFF
--- a/lib/components/popover.vue
+++ b/lib/components/popover.vue
@@ -174,10 +174,10 @@
                     targetAttachment: this.placementParameters.targetAttachment
                 };
             },
-            
+
             setPopoverStyle() {
                 return this.popoverStyle;
-            },
+            }
 
         },
 
--- a/lib/components/popover.vue
+++ b/lib/components/popover.vue
@@ -177,7 +177,7 @@
             popoverMaxWidth() {
                 var sizing = {
                     'v-sm': '75px',
-                    'sm': '250px',
+                    'sm': '276px',
                     'md': '450px',
                     'lg': '575px',
                     'v-lg': '750px',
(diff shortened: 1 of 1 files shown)


pi0:
Generally hard coding constants is not a good idea! It is accepted that we need ability giving custom styles to popover such max-width. Why not just providing an `popover-class`  or `popover-style` props?

BramVer:
That's a much better approach indeed, I'll implement it now.

lib/components/popover.vue
```diff
@@ -103,6 +102,15 @@
                 validator(value) {
                     return value >= 0;
                 }
+            },
+            popoverStyle: {
+                type: Object,
+                default() {
```
pi0: Default can be simply null not a function

BramVer: Right, I was setting it to {} before instead of null, and then I had to return it as a function. This was a leftover of that apparently.

lib/components/popover.vue
```diff
@@ -165,7 +173,12 @@
                     attachment: this.placementParameters.attachment,
                     targetAttachment: this.placementParameters.targetAttachment
                 };
+            },
+
+            setPopoverStyle() {
```
BramVer: Absolutely not needed at all. Currently in the habit of using setters instead of the vars directly, because of the project I'm working on, didn't think about it.

Issue #167, opened 2017-03-23
Input directory


boshevski:
The form-file component works great for files. How do we go about choosing a directory instead of a file?

pi0:
@boshevski Multiple file uploading is now implement :) Will come to new release soon ..
![image](https://cloud.githubusercontent.com/assets/5158436/24239677/8848722a-0fcc-11e7-810d-8fc2b466e467.png)

mosinve:
@pi0 it's not a directory feature ;)
There is experimental param to <input file>  - **webkitdirectory**
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory
http://jsfiddle.net/addyo/Ha979/
https://plus.google.com/+AddyOsmani/posts/Dk5UhZ6zfF3

pi0:
@mosinve Thanks for tip i'm going to dig into them for more information :) 
> The HTMLInputElement.webkitdirectory is a property that reflects the webkitdirectory HTML attribute and indicates that the <input> element should let the user select directories instead of files. When a directory is selected, the directory and its entire hierarchy of contents are included in the set of selected 

As mentioned in MDN, it just allows user select entire directory instead of individual files. Seems handling both multiple and directory is same. The only different would be that each file in the array has `File.webkitRelativePath`  attr: 

> After the user makes a selection, each File object in files has its File.webkitRelativePath property set to the relative path within the selected directory at which the file is located

So I'll add another prop `directory` to allow enabling such behavour.

mosinve:
Yep.... but i think we need to make notice in docs, that this feature has limited support in browsers.
In firefox we must set `allowdir=true `additional to `webkitdirectory` param

pi0:
Sure I'll add notice. Will inform you after merge so we can check it before release :)

boshevski:
That was quick! Thank you @pi0 and @mosinve 👍

pi0:
@boshevski  0.10.2 released :)

PR #169, opened 2017-03-23
[file-input] drop and multi files support + improvements #167

pi0:


pi0:
@mosinve ping!

mosinve:
Seems to work.... but feels like it's a one big cheat)))) 
found some more info about webkitdir
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitEntries
https://developer.mozilla.org/en-US/docs/Web/API/File/webkitRelativePath

pi0:
Yes somehow workaround but at least we keep API consitansy and users can use easier. The only problem as you mentioned in links is when user drags directory in directory mode. This case was not handled...

mosinve:
yeah... there are left things to work on.
one more addition - IMHO we need to add most of the props of standart input file el.
i found that missing accept and disabled props...

mosinve:
if listen to native dragover event on label, then drag-n-drop directories works, but doesn't work 'drop here target' ;(

pi0:
I'm going to start working on it in few minutes later :)

pi0:
@mosinve Finally working :)) Also there is an almost rewrite for all input elements. props are refactored to a single mixin. so we can maintain it better in future especially on v4 release.

PR #182, opened 2017-03-28
Tabs refactoring

mosinve:
Hi @pi0, there is my refactor attempt of Tabs. Please check code b4 merge :) I've tested already,  but just for sure :)

### DIFF
--- a/docs/components/index.js
+++ b/docs/components/index.js
@@ -24,5 +24,6 @@ export default {
     'popover': require('./popover').default,
     'progress': require('./progress').default,
     'table': require('./table').default,
+    'tabs': require('./tabs').default,
     'tooltip': require('./tooltip').default
 };
--- a/docs/components/tabs/README.md
+++ b/docs/components/tabs/README.md
@@ -1,0 +1,42 @@
+# Tabs
+
+**Basic usage**
+
+```html
+<b-tabs>
+  <b-tab title="first" active>
+    I'm the first fading tab
+  </b-tab>
+  <b-tab title="second" >
+    I'm the second tab content
+  </b-tab>
+  <b-tab title="disabled" disabled>
+    <b-card>I'm the card in tab</b-card>
+  </b-tab>
+</b-tabs>
+```
+
+**Cards Integration**
+
+Tabs support integrating with cards. Just add `card` prop!
+
+Note that you should add `no-block` prop on `<b-card>` element in order to decorate header.
+
+```html
+<b-card no-block>
+    <b-tabs ref="tabs" v-model="tabIndex" card>
+        <b-tab title="Tab 1" active>
+            Tab Contents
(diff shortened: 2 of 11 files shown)


pi0:
@mosinve Thnks :) Preparing docs. If ESLint fails i'll fix it too

pi0:
Change list : 
- docs page
- updated example
- fix bug when we initially set tab to something not 0
- changed fade to no-fade for easier usage
- cards integration ability

@mosinve Would you please check changes and also what is your idea changing `navStyle` to just a `pills` prop? Is it intended for special usages or we may have only two possible values for that?

mosinve:
About navstyle,  I left it as it was. In bootstrap there're 2 options: tabs or pills. So this prop intended to handle this. 
Quickly looked at changes from phone ,  and don't understand about cards... Check this today later from pc.

mosinve:
Figured about cards - cool feature :)
One little notation: "card?'card-header':null" -  {'card-header': card} looks better IMHO, and according Vue's guide it's the same ;)

pi0:
@mosinve I was just used to `card?` syntax :)) Yep object style is much better. 
`v-show` may help a little. But a real case: we have a blogging platform (like instagram) with 3 tabs editor, preview and comments. All of them are heavy enough ckeditor needs a big js chunk, comments are heavy and if we just use v-show we can't lazy load tabs. We may combine both methods with a `lazy` prop on each tab-item : 
```html 
<tab-item v-if="localshow || !lazy" v-show="localshow || lazy">```

mosinve:
I like lazy variant - it's more flexible. 
Pills prop - i think about it too, so i agree ))

And one more thing, when i tested tabs on local playground, i catch a bug with parent/child assigment. Because of it i moved tabs init from tab (this.$parent) to tabs (this.$slots) -  in playground all components are childen of this.vm, but in real life - tab are child of tabs so old version worked well, but gave an error in playground.

pi0:
@mosinve  I didn't get your meaning about $slots . Do you mean it won't work outside of playground?

mosinve:
slots work everywhere
$parent/$child - not work in playground

pi0:
I think i got it :)) @mosinve Would you please commit final fixes for that? so we merge & release it sooner :)

mosinve:
I'll do

mosinve:
funny thing:
if I write in tab item 
`v-if="localActive" v-show="localActive"`
transition begin to work smooth ))) even without lazy prop....

pi0:
😂 It seems being related to <transition> wrapper.

pi0:
@mosinve Merge it when done :) We can be ready for next release

pi0:
поздравления :)

mosinve:
متشکرم ;) Hope google trans make it correct)))))

PR #187, opened 2017-03-29
fixed carousel controls display 

Phinome:
i used `<b-carousel>` component. the controls display error, so i compare bootstrap v4 `Carousel Controls html & css` to take it being same. now , it looks to work good.

oh, the jsFiddle url , i do not know , so it equal ''.

Lucky

### DIFF
--- a/examples/carousel/demo.details
+++ b/examples/carousel/demo.details
@@ -1,14 +1,13 @@
  /*
  ---
- name: Cards - BootstrapVue
- description: https://bootstrap-vue.github.io/docs/components/card
+ name: Carousel - BootstrapVue
+ description: https://bootstrap-vue.github.io/docs/components/carousel
  authors:
-   - Pooya Parsa
+   - Pooya Parsa Phinome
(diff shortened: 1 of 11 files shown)


examples/carousel/demo.details
```diff
@@ -0,0 +1,14 @@
+ /*
+ ---
+ name: Carousel - BootstrapVue
+ description: https://bootstrap-vue.github.io/docs/components/carousel
+ authors:
+   - Pooya Parsa Phinome
```
pi0: I think you can just remove my name from beginning :))

pi0:
@Phinome Thanks 👍  So we can finally release carousel as an official component and have full bootstrap-4 coverage. Only one thing remains before that we may need to refactor transitions new vue.js `<transition>` system. If you are willing to complete that i would happy be available until then to help. Else just we can merge this PR and postpone things to later.

pi0:
Merging to go with second option :)) Thnaks

PR #198, opened 2017-04-02
[Table] Fixed pagination when filtered table

mosinve:
Fixed pagination by adding v-model to table, so it can throw outside count of filtered items, then we can pass this data to pagination as totalrows.

The another approach may be using `this.$root` as event bus for `table.emit('filtered', items.count)` and `pagination.on('filtered', function(totalrows){....})`.

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -119,9 +119,9 @@
                             regex = new RegExp('.*' + this.filter + '.*', 'ig');
                         }
                         items = items.filter(item => {
-                            let test = regex.test(toString(item));
+                            const test = regex.test(toString(item));
                             regex.lastIndex = 0;
-                            return test
+                            return test;
                         });
                     }
                 }
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -55,8 +55,7 @@
             },
             fields: {
                 type: Object,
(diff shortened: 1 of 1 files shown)


lib/components/table.vue
```diff
@@ -127,12 +131,11 @@
                         return this.sortDesc ? r : r * -1;
                     });
                 }
-
+                this.$emit('input', items.length);
```
pi0: We should keep tables more abstract. If only binds to length we for example can't receive new data. Why not simply binding model to `items` itself? Also please consider that this emit is in a computed property! we should somehow keep latest value and compare before this event to prevent duplicate ones :)

mosinve:
OK, I'll think about it further tomorrow)) this was a quick fix approach)))

2 апр. 2017 г. 23:00 пользователь "Pooya Parsa" <notifications@github.com>
написал:

mosinve:
@pi0, canimerge? :) 
I thought about moving filtering out, but it won't make a difference too much. IMHO where emit now,  is the most suitable place :) 
Also i include regex bugfix here.

pi0:
@mosinve Thanks Seems nice :) Yep I'll merge it now. adding some improvements later for ajax support

PR #174, opened 2017-03-25
Couple of fixes

gpyh:
I had to make some fixes for a project I'm working on. I'm sharing them back upstream.
In details :

- Uses directly the bLink components by importing it. I'm using Vue without the template compiler and it rightly complained about not knowing `b-link`. This fixes that issue.
- ~~Slight code improvement for bFormInput: it uses `is` instead of `v-if` to switch between input and textarea. If found this cleaner.~~
- Blur and focus events from input and textarea and propagated up to chain in bFormInput.

PR #201, opened 2017-04-04
b-button: Remove warning by importing b-link

gpyh:
Same drill as last time. (#174)

### DIFF
--- a/lib/components/button.vue
+++ b/lib/components/button.vue
@@ -5,7 +5,12 @@
 </template>
 
 <script>
+    import bLink from './link.vue';
+
     export default {
+        components: {
+            bLink
+        },
         computed: {
             classObject() {
                 return [
@@ -18,7 +23,7 @@
                 ];
             },
             componentType() {
-                return (this.href || this.to) ? 'b-link' : 'button';
+                return (this.href || this.to) ? bLink : 'button';
             },
             btnBlock() {
                 return this.block ? `btn-block` : '';


lib/components/button.vue
```diff
@@ -18,7 +20,7 @@
                 ];
             },
             componentType() {
-                return (this.href || this.to) ? 'b-link' : 'button';
+                return (this.href || this.to) ? bLink : 'button';
```
pi0: i think you have to keep this as is (b-link) as kebab case is preferred.

lib/components/button.vue
```diff
@@ -5,6 +5,8 @@
 </template>
 
 <script>
+    import bLink from './link.vue';
+
     export default {
```
pi0: We need to define imported module like this:
```js
components: {bLink}
```

lib/components/button.vue
gpyh: This is not a string. It's a variable name.

gpyh: I didn't take time to explain the reasoning as it's the same as #174

mosinve: @gpyh, 
comparing to #174, this peace specially omitted or forgotten?

        components:{
            bLink
        },

gpyh: It actually doesn't need the component to be registered as it uses directly its description.
It's still a good idea to add it though. I'll do that.

gpyh:
Is something still blocking this PR?

I realize I didn't motivate the change properly. Allow me to fix this:

I use bootstrap-vue by importing each component manually instead of registering them in the global scope. It makes the final bundle smaller and the code more explicit.
The problem I encounter is the following:

Say I have a component A in my app that depends on a bootstrap-vue component called B, which itself depends on another bootstrap-vue component named C. The dependency graph looks roughly like this:

```
A (my own) -> B (bootstrap-vue) -> C (bootstrap-vue)
```

The code for A would be roughly this:

```
import { B } from 'boostrap-vue/lib/components'

export default {
  name: 'A',
  components: {
    B
  }
}
```

My app has no way of knowing that B depends on C. It can only work with the tag name. At this point that tag name is simply a string unrelated to any actual component.

Registering this transitive dependency in my components would look like this:

```
import { B, C } from 'boostrap-vue/lib/components'

export default {
  name: 'A',
  components: {
    B, C
  }
}
```

With this I can use C in A but the C in B still doesn't render. It's because the components are registered locally ; Vue doesn't walk up the components hierarchy. A component is either global or strictly local.

To fix this, bootstrap-vue needs to be explicit in the dependency it uses. It means importing and registering the appropriate components.

For #174 this was what I did. Here, the component is dynamically determined.
In that situation, you can either assign the component with its tag name, or directly with its description object.
That's why importing the component is needed, but not registering it.

I only made those fixes as they appeared in my code, but in my opinion a better course of action would be to enforce this pattern across every component. I can carry out that task of reviewing all the components and making this fix if you want me too.

pi0:
> but in my opinion a better course of action would be to enforce this pattern across every component.

@gpyh Thanks for your explanations and we totally agree about this to be done in entire components, The only question I had was that why we should point to reference instead of string `b-link` form? (We are already registring tag in `components` section :)

pi0:
Long time since it is open. Merging this as a starting point. Will discuss before migration other components.

PR #204, opened 2017-04-05
Adds "clicked" prop that run when row is clicked

tithamane:
When the row of a table is clicked, we would like to perform a redirect.

Allow the user to pass in a function that will be executed when the row
is clicked. It's up to the user to add the redirection functionality within the method
as the method is passed the current item and its index.

This way a developer is not can do whatever they want within this function(we go to a different URL).

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -10,7 +10,7 @@
         </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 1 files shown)


mosinve:
you can use `.native` event modifier to listen to native browser events

tithamane:
Hmmm... How would I pass that to b-table?

PR #206, opened 2017-04-05
Add an option for a custom tab-head

valexiev:
There use cases when you need custom tab heads that are not <a>.

### DIFF
--- a/lib/components/tabs.vue
+++ b/lib/components/tabs.vue
@@ -6,8 +6,11 @@
                     <a :class="['nav-link',{small: small, active: tab.localActive, disabled: tab.disabled}]"
                        :href="tab.href"
                        @click.prevent.stop="setTab(index)"
-                       v-html="tab.title"
-                    ></a>
+                       v-if="!tab.headHtml"
+                       >{{ tab.title }}</a>
+                    <div :class="['tab-head',{small: small, active: tab.localActive, disabled: tab.disabled}]"
+                         v-else
+                         v-html="tab.headHtml"></div>
                 </li>
                 <slot name="tabs"></slot>
             </ul>
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "rollup-plugin-uglify": "^1.0.1",
     "rollup-plugin-vue": "^2.3.1",
     "rollup-watch": "^3.2.2",
+    "vue": "2.2.5",
     "xo": "^0.18.1"
   },
   "xo": {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -57,7 +57,6 @@
     "rollup-plugin-uglify": "^1.0.1",
     "rollup-plugin-vue": "^2.3.1",
     "rollup-watch": "^3.2.2",
-    "vue": "2.2.5",
```
pi0: This is intentionally for nuxt.js compatibility  :))

lib/components/tabs.vue
```diff
@@ -6,8 +6,11 @@
                     <a :class="['nav-link',{small: small, active: tab.localActive, disabled: tab.disabled}]"
                        :href="tab.href"
                        @click.prevent.stop="setTab(index)"
-                       v-html="tab.title"
-                    ></a>
+                       v-if="!tab.headHtml"
+                       >{{ tab.title }}</a>
+                    <div :class="['tab-head',{small: small, active: tab.localActive, disabled: tab.disabled}]"
```
pi0: If the difference is only changing tag type to div we can do this using `is` directive like this: `<a :is="headHtml?'a':'div'>`

valexiev: Your suggestion looks much better : )

package.json
```diff
@@ -57,6 +57,7 @@
     "rollup-plugin-uglify": "^1.0.1",
     "rollup-plugin-vue": "^2.3.1",
     "rollup-watch": "^3.2.2",
+    "vue": "2.2.5",
```
mosinve: Is there a reason to have vue in package.json twice and different versions?
I had npm complains about it.

valexiev: @pi0 said this is intentional for nuxt.js compatibility.

pi0: nuxt depends on an older vue-template compiler with lower semver/minor version & Vue requires exact matching version to work, however this is just for development and docs generation. Thia was the only workaround i had found.

lib/components/tabs.vue
pi0: @valexiev Please let me know about updates :) So we can Merge this ASAP :)

valexiev: I've tested this construction `<a :is="headHtml?'a':'div'>`, but it doesn't work.
The 2 reasons for using the second case with <div> are:
1) to be able to customize the tab head
2) to be able to make the tab head not clickable

If <a :is="headHtml?'a':'div'> works the tab head still have @click.

Because of the reasons above i will keep the change as i've proposed it.

pi0: Seams reasonable :)

Issue #207, opened 2017-04-05
Validating fields inside tabs


milenon:
When I try to validate a field inside a tab using vee-validate it is beeing validated and right afterwards the tab is being hidden, because its 'show' class is being removed from 

`<div role="tabpanel" class="tab-pane fade active show">`

Here is the sample code

> 
                <b-tab title="Test">
                    <div>
                        <label for="key">Key</label>
                        <div class="col-xs-12 col-sm-9">
                            <input type="text" id="key" name="key" v-validate="'required'" class="form-control" v-model="key">
                            <small v-show="errors.has('key')">
                                Please enter the Key.
                            </small>
                        </div>
                    </div>
                </b-tab>

If you then go to different tab and return to this one, you will see the field with the error message and it will not disappear again.

PR #209, opened 2017-04-06
Fix bug of tab component removing 'show' css class prematurely

valexiev:
Fixing issue #207.


valexiev:
The case:
A text input in a tab. When you start typing in the input, the tab content disappears because the CSS class "show" is removed.

This happens because after typing in the text input Vue reapplies classes, but uses only the initial class ("tab-pane") and the value of :class. Vue doesn't know anything about "show" and doesn't reapply it together with the other classes.

pi0:
@mosinve Please merge if you approve. LGTM :)

mosinve:
@valexiev Have you tested? Tabs transition works?

mosinve:
Merged

PR #213, opened 2017-04-07
[WIP] Better Tables

pi0:
This is work-in-progress (and still not working!) attempt to decouple table data from its presentation component. I've just opened this so if anyone has ideas on this we can work/discuss here.
This allows users to have complete control over data and make it possible to server side operations.

### DIFF
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -2,7 +2,7 @@
 
 > For tabular data. Tables support pagination and custom rendering.
 
-### `fields`
+### Fields
 Fields prop is used to display table columns. 
 keys are used to extract real value from each raw.
 Example format:
@@ -17,7 +17,7 @@ age: {
 }
 ```
 
-### `items`
+### Items
 Items are real table data records. Example format:
 
 ```js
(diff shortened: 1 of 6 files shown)


mosinve:
Just thoughts...
What if we expose events to allow user bind callbacks to them, providing custom contol of getting data, filtering, sorting, pagination control...
and we can provide built-in out-of-the-box table controller just to make component work.

pi0:
Using events callback is a good idea! :) Biggest problem is with async data, we need to somehow update component after pagination and other operations and Vue don't seem to support this. Also table component is getting bigger and bigger as we are adding more features to it.
If you see changes I've moved pagination and other props to the adapter. You think keeping them here would be better?

mosinve:
I just think that we build an UI element, which doesn't need to know where data come from and how it processed, it's job is to display that data the way we need. 
So table will need some support code to make it work.

pi0:
@mosinve That is the exact reason I am extracting that process logic outside of component, so we have tow solutions: 
- Implement and keep a bare-bone event based API inside component which sends callbacks when needs new data. (Component requests to be controlled from outside world via events)
- Allow users passing an `Adapter` which provides anything that component needs. (Outside world provides  an api that component can use to retrieve data)

mosinve:
I think that event based API will be more flexible.
And i agree that table component getting more complex.  It's already not just a table from bootstrap, it's on the way to be like jQuery Datatable.

PR #214, opened 2017-04-07
Two minor fixes

gpyh:
I'm submitting two fixes:

- ~~Added the `id` prop to b-form-input. This is not necessary but it prevents Vue from complaining that the property is not defined~~
- I changed the behavior of b-modal's click events so that `closeOnBackdrop` has the intended behaviour. In particular:
  * I removed the `id` check is I want to use modals without setting any id
  * I capture click events on the modal's content. That way the modal only closes when the user clicks on the backdrop; not on the modal itself

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -15,7 +15,7 @@
             >
 
                 <div :class="['modal-dialog','modal-'+size]">
-                    <div class="modal-content">
+                    <div class="modal-content" @click.stop>
 
                         <div class="modal-header" v-if="!hideHeader">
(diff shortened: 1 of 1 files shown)


pi0:
@gpyh About id field, is it already added form mixin. Did you checked latest release?

gpyh:
I updated and the warning isn't being raised anymore, so fix 1 is now unneeded.

As for the linting, I forgot to run it again. I really should add a git hook. Sorry.

pi0:
Thanks @gpyh :)

PR #217, opened 2017-04-10
Add clickable table row

tithamane:
We wanted to have the ability to redirect when clicking a row instead of using a button. We opted to go with passing a function to the component and have that function execute when the row is clicked in which we redirect.

Using a function will give us some flexibility in the case that we might want to do something else.

If there is a builtin/easier way to do this then we'd like to know about and forget about this pull request.

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -143,7 +143,7 @@
         methods: {
             rowClicked(item, index) {
                 if (this.rowClick) {
-                    this.rowClick(item, index)
+                    this.rowClick(item, index);
                 }
             },
             headClick(field, key) {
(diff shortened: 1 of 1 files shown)


pi0:
@tithamane Thanks for PR. I was thinking of adding this option too. However, i think using events would be a better idea instead of passing a function and then calling it. `this.$emit('rowClicked',item, index);`

tithamane:
@pi0 I like the idea. So rowClicked would be a boolean?

pi0:
@tithamane We can then remove that property :) Usage is like normal events `<b-table @row-clicked="myFunction">`

tithamane:
Is the current change fine?

pi0:
Yep :) Thanks

PR #223, opened 2017-04-12
better collapse animation

ohmygod12121212:
Since css animation does not support 0 to auto, using max-height does not look perfectly, the transition time seems less than expected.
I add a wrapper div to detect the height of content, and add the height dynamically to the transition element's style.


pi0:
@ohmygod12121212 THanks :)

PR #230, opened 2017-04-13
Popover: Force hide with prop, improved context reactivity

SirLamer:
Changes:
- The property 'show' now defaults to null. When 'false', the Popover is forbidden from showing. This overrides all other triggers.
- Like before, when the 'show' property is true it will trigger Popover display. Unlike before, so long as the property value remains true, the Popover will continue to display despite any DOM events or external global event calls
- Leave 'show' = null to maintain dynamic responsive show behavior (default)
- Display reactivity is improved for all universal cases. The Popover will be hidden along with its anchor element whenever that anchor element is no longer being rendered (examples: multi-tab page, modal hidden)
- Removed reactivity to Tab and Modal events, since they are no longer needed for responsive contextual display of the Popover dialogue.
- General refactor to improve readability, reduce compile size and reduce run-time memory usage


SirLamer:
I am leaving this parked here for the moment to welcome input/feedback. I am not merging myself yet.

SirLamer:
All also removed all of the event hooks and things that were being used as a "hack" means to affect reactivity. Since they 'over-reached' a bit, meaning they are reacting to all events event when external to the context of the given component, it needed to be revisited.

pi0:
@SirLamer Seems nice +1 ... I just didn't notified how are we checking popoever is in viewport or not?

SirLamer:
Tether takes care of it, we just have to ask it to check. So we hook Vue's 'updated()' lifecycle event, so every the context for the given Popover component changes it re-fires 'Position'. If the anchor element is not in view then the Popover is not displayed. Note this means that if the Popover is currently in a state where it is expected to be shown, it remains loaded and waiting in "standby" to be shown again should it re-appear. So it behaves more like expected. With the previous implementation, where the Tab event just destroyed the Tether object, that was actually break functionality in the case where the Popover-target element was later made to be shown again.

My test code:
```html
<input v-model="input"></input>
<br>
<br>

<b-tabs>
  <b-tab title="first" active>
    I'm the first fading tab
    <br>
    <br>
    <br>
    <b-popover :content="this.input" triggers="click">
  <b-btn>Hi</b-btn>
</b-popover>
  </b-tab>
  <b-tab title="second" >
    I'm the second tab content
  </b-tab>
  <b-tab title="disabled" disabled>
    <b-card>I'm the card in tab</b-card>
  </b-tab>
</b-tabs>
```

```js

data: {
      input: 'Oh hi'
}
```

pi0:
hmm.. Using updated() hook was great idea :) It would be so useful for other users in different use cases if you had time to write a blog post about that.

pi0:
Anyway LGTM. We may merge it when your work was done and be ready for next release :)

SirLamer:
Okay but... where? I don't have a blog. :-P

My real job is mechanical engineer / elevator inspector. I am self-taught as far as programming is concerned and I am working on a CRM for the company I work for. I don't have enough real discipline or experience to pretend that I should have a programming blog. ;-)

pi0:
@SirLamer Really ? :)) But you are way better in solving programming problems from many programmers i know. Blogs are dead! Just head to [Medium](medium.com) and post it as a story. We can publish it via twitter :)

SirLamer:
By the way, the next thing that is coming... harmonizing the code base of Popover with Tooltip. Because Popover is really just a special case of Tooltip.

But that'll have to be another day.

PR #233, opened 2017-04-15
Popover: Transition animal, CSS cleanup on dismiss

SirLamer:
This implements two features:
- Smooth fade transition on show and fade. This is identical to that in native Bootstrap 4
- The CSS classes added by Tether to the target element and now cleaned up on unloading. Note the current implementation looks for all classes with the prefix "tether-" and removes them, so any custom classes should not also start with "tether-". I consider this to be very unlikely and therefore safe.

### DIFF
--- a/lib/components/popover.vue
+++ b/lib/components/popover.vue
@@ -46,6 +46,8 @@
         }
     };
 
+    const TETHER_CLASS_PREFIX = 'tether-';
+
     const TRANSITION_DURATION = 150;
 
     export default {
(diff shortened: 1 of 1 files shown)


SirLamer:
For those wondering on the animation, as other components will benefit from the same, here's how it works:

Bootstrap v4 ships with CSS transitions ready to go for these animations. For Tooltip and Popover, it includes a base "startup" class "fade" which sets up the transition. Later, when the class "show" is added this starts the animation.

If you look at the code, you'll notice that the "show" class is added after initial load using the "$nextTick"  method. Because of the way Vue works regarding DOM manipulation, Vue will resolve everything internal to itself before pushing any changes to the DOM. If the "show" class it just made to be included in the frame that the Popover is created on the first load, this means that the element is created with "fade" and "show" classes already in there together. As a result, the CSS transition will not occur. By adding "show" next frame, it does occur. This only matters on the first instance of use but since it's simpler and more resilient to future changes I just left it this way for every use.

PR #250, opened 2017-04-19
Use default id (b_UID) for checkbox, file and select.

charkins:
This behavior already exists in the form-input and will ensure
a valid id exists for the label to reference when using the
form-fieldset component.


pi0:
@charkins Thanks :)

Issue #195, opened 2017-04-01
addons support for input groups


pixelbucket:
I've modified my form-input.vue locally to allow input groups. No real testing has been done outside of my personal use case so no pull request.  I'm also pretty sure doing this with named slots would be better to allow for things like checkboxes. I only needed it to add some dollar signs, so this way works for me.

**form-input.vue template becomes:**

```html
<template>
<div class="input-group">
    <div v-if="addonLeft" v-html="addonLeft" class="input-group-addon"></div>
    <input :type="type" :value="value" :name="name" :id="id || ('b_'+_uid)" :disabled="disabled" ref="input" :is="textarea?'textarea':'input'" :class="['form-control',inputClass]"
        :rows="rows || rowsCount" :placeholder="placeholder" @input="onInput($event.target.value)" @change="onChange($event.target.value)" @keyup="onKeyUp($event)"
        @focus="$emit('focus')" @blur="onBlur()" />
    <div v-if="addonRight" v-html="addonRight" class="input-group-addon"></div>
</div>
</template>
```

**Add props for addonRight and addonLeft...**
```javascript
        addonRight: {
          type: String,
          default: null
        },
        addonLeft: {
          type: String,
          default: null
        },
```

**And then use like this:**
```html
<b-form-fieldset label="Price" :label-size="5" horizontal>
  <b-form-input v-model="item.price"addon-left="$"  />
</b-form-fieldset>
```

pi0:
@pixelbucket Hi and thanks for suggestion, About slots yes i think it would be a better approach. We will add addons support soon in next releases :)

tmorehouse:
V4 input groups allow for multiple left or right addons now.  So maybe a slot for left and slot for right, and introduce a new component for addon (which would use a default slot for the addon content)

tmorehouse:
PR #266

tmorehouse:
This PR created new components for `input-group`s and `input-group-addon`s

pi0:
Thanks to @thomasfowler . Will be available soon.

tmorehouse:
PR #267 adds support for `input-group-button` (although dropdown button is not supported)

pi0:
Added in 0.14.0 :)

MarZab:
Getting a warning while using the left slot:

> Duplicate presence of slot "left" found in the same render tree - this will likely cause render errors. 
(found in <BInputGroup>)

pi0:
[...]

PR #266, opened 2017-04-21
input-group and input-group-addon support

tmorehouse:
Adds two new components for handling input groups and input-group-addon

```html
<b-input-group size="lg" state="warning">
  <b-input-group-addon slot="left">$</b-input-group-addon>
  <b-form-input type="number" v-model="value"></b-form-input>
  <b-input-group-addon slot="right">.00</b-input-group-addon>
<b-input-group>
```


tmorehouse:
Addresses issue #195

pi0:
@tmorehouse Awesome thanks. To keep consistent addons with other components usage, i suggest adding `left` and `right` props to to `<b-input-group>` so new usage would be like this:
```html
<b-input-group size="lg" state="warning" left="$" right=".00">
  <b-form-input type="number" v-model="value"></b-form-input>
<b-input-group>
```

lib/components/input-group.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+    <div :class="classObject">
+        <slot name="left"></slot>
```
pi0: Can be simply rewritten to:
```html
<slot name="left">
  <span class="input-group-addon" v-if="left" v-html="left"></span>
</slot>
```

tmorehouse: Input groups can have multiple left and right addons under bootstrap V4 though, and this would limit it to only a single left and right addon.

https://v4-alpha.getbootstrap.com/components/input-group/

tmorehouse: Or I suppose it could be combined for both.  simple usage by the left/right propes, or slots for more complex add-ons.

pi0: Making api complex for just special use cases is not also a good idea. just consider how much simpler is current usage than using bootstrap html markups directly. For multiply addons, we can support array values and rendering internally using `v-for`.

pi0: You are true, we can provide both methods and let users choose best one that fits to their usages :)

lib/components/input-group-addon.vue
```diff
@@ -0,0 +1,17 @@
...
+        <slot></slot>
+    </span>
+</template>
+
+<script>
+    export default {
+        props: {
+            id: {
+                type: String,
+                default: 'b_input_geoup_addon_' + this._uid
```
pi0: I think this won't work correctly as `this` is not available in props. would be really need that id?

tmorehouse: Ech component has a unique ID on the vue VM instance as _uid

It will be needed if one wants to use aria-labeledby when referencing the addon(s).

But I can remove it and make the user enter an ID if needed.

tmorehouse: Ah, I could male a computed property `thisId' that generates the ID if one doesn't exist.

pi0: Yep i think computed property would be better :)

lib/components/input-group-addon.vue
```diff
@@ -0,0 +1,22 @@
...
+    <span class="input-group-addon" :id="thisId">
+        <slot></slot>
+    </span>
+</template>
+
+<script>
+    export default {
+        computed: {
+            thisId() {
+                return: 'b_input_group_addon_' + this._uid;
```
pi0:
```js
return this.id || 'b_input_group_addon_' + this._uid;
```

tmorehouse: Oops. I was typing too fast ;-)

pi0:
Looks well. Will supply some docs and release soon :)

PR #269, opened 2017-04-22
Disabled  dropdown item fix

tmorehouse:
I noticed that `dropdown-item`s were not reflecting the disabled state. See this fiddle for an example: https://jsfiddle.net/oeotafg1/2/

This PR adds in disabled handling in `b-dropdown-item`, `b-link` and `b-button`.

For items rendered as an `a` tag, it also prevents the click handler from firing, making it truly disabled.

Check it out and see if it makes sense.

### DIFF
--- a/lib/components/button.vue
+++ b/lib/components/button.vue
@@ -76,7 +76,12 @@
         },
         methods: {
             onclick(e) {
-                this.$emit('click', e);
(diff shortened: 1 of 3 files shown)


pi0:
Thanks @tmorehouse for this.

Issue #271, opened 2017-04-22
Popover is being rendered on the DOM as invisible


SirLamer:
The Popover object is being rendered but is invisible, depending on its location it can block other UI elements.

This is for me to fix! I am just listing this here so that others know I am on it.

PR #272, opened 2017-04-23
Refactoring, fix element rendering as invisible when hidden

SirLamer:
Per https://github.com/bootstrap-vue/bootstrap-vue/issues/271

### DIFF
--- a/lib/components/popover.vue
+++ b/lib/components/popover.vue
@@ -27,26 +27,18 @@
     };
 
     const PLACEMENT_PARAMS = {
-        top: {
-            attachment: 'bottom center',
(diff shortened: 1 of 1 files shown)


SirLamer:
@pi0 Can you please check the test failure? That looks like something external to this PR. I am not merging yet.

pi0:
@SirLamer unfortunately this is happening during some package installation which fails as it downloads from github in build step. Will try to upgrade yarn to see if problem happens or not.

pi0:
Not merging. Merge when you where ready :)

SirLamer:
OK it passes now.

PR #275, opened 2017-04-24
b-link: Fix typo

gpyh:
The bane of dynamic typing :P


gpyh:
Impressive reaction time!

PR #278, opened 2017-04-24
b-carousel-slide: Trim down classes

gpyh:
Two fixes here for b-carousel-slide:

- The class carousel-caption was used even when caption was not set. Now it checks for it.
- The association of d-none and d-md-block used to make the content disappear on small viewport. This is taken from bootstrap's documentation and was intentional in there. The intent was to hide the *caption* on small viewport. As it's very specific and also affects the content I removed it. If you want the intended behaviour back I can work on it.

### DIFF
--- a/lib/components/carousel-slide.vue
+++ b/lib/components/carousel-slide.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="carousel-item" :style="{background,height}">
         <img class="d-block img-fluid" v-if="img" :src="img" :alt="imgAlt">
-        <div class="carousel-caption d-none d-md-block">
+        <div :class="{ 'carousel-caption': !!caption }">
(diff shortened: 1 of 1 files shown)


pi0:
@gpyh  What if we add an option like `visible-size` (or anything similar) which conditionally adds this class and is null by default?

gpyh:
I'm OK to implement that. But should this affect only the caption or the whole content?

pi0:
I think content only just like bootstrap example for smaller screens....

pi0:
Will merge and improve for faster delivery :)

gpyh:
I'll submit a new PR in the following 48 hours (likely this evening).

pi0:
Nice, thanks :)

Issue #277, opened 2017-04-24
BFormFieldset - TypeError: Cannot read property 'id' of null


AndrewBarba:
After upgrading 0.12.x to 0.13.x and now 0.14.x I am seeing this error on all of my `<b-form-fieldset>` components

<img width="664" alt="screen shot 2017-04-24 at 8 55 04 am" src="https://cloud.githubusercontent.com/assets/1316152/25337988/bc5134e0-28cb-11e7-9c80-541815f325a0.png">

AndrewBarba:
Okay I found the issue, I was using a `b-dropdown` inside the field set which doesn't have an input component. I guess this is incorrect usage

<img width="269" alt="screen shot 2017-04-24 at 10 28 31 am" src="https://cloud.githubusercontent.com/assets/1316152/25342096/e88ccdbe-28d8-11e7-9fcf-03c6c6b490a9.png">

tmorehouse:
The fieldset looks for the first input/select/textarea to grab an ID for associated the `label`.

This PR #279 checks for the existence of the inputs before checking the ID.

tmorehouse:
Also, If you know that your dropdown has an ID (if you assign one), then you can change the prop `inputSelector` on the fieldset to create a CSS selector that matches your dropdown (i.e. `'div.dropdown'`, or `'div#myid'`) to associate the label to your form control.

AndrewBarba:
Ah makes sense, thanks!

PR #279, opened 2017-04-24
Fix for #277

tmorehouse:


lib/components/form-fieldset.vue
```diff
@@ -27,12 +38,15 @@
                 return this.horizontal ? ('col-sm-' + (12 - this.labelSize)) : 'col-12';
             }
         },
-        mounted() {
-            const content = this.$refs.content;
-            if (!content) {
-                return;
+        methods: {
+            target() {
```
pi0: Can't we implement and use target as a computed property?

tmorehouse: Hmmm, would the this.$refs.content exist when the prop is calculated?

tmorehouse: We can try as a computed property, and see how it works.

tmorehouse: I could also alter the default `inputSelector` value to include `.dropdown` and the check for its ID existance. And in dropdown, we could generate a default `b_UID` if an ID is not given.

tmorehouse: Would maybe placing the target computation in `mounted()` and `updated()` be safer (and have target as a data prop)?

pi0: The thing I'm sure $refs is only accessible​ after mounted() so we may keep it as a method but just call it inside them :)

tmorehouse: OK so leave it as a method, and then have `mounted()` and `updated()`call the method and update a data value, which is then referenced by the the `<label>`'s `:for`

tmorehouse:
Once this is merged, I will add a default ID to the `.dropdown` if one is not provided to the dropdown component.

lib/components/form-fieldset.vue
```diff
@@ -27,12 +39,21 @@
...
-            if (!content) {
-                return;
+        methods: {
+            updateTarget() {
+                const content = this.$refs.content;
+                if (!content) {
+                    return null;
+                }
+                const input = content.querySelector(this.inputSelector);
+                return (input && input.id) ? input.id : null;
```
pi0: Shouldn't we update some data here? :))

tmorehouse: Ah crap.  My fingers are too fast and not listening to my head :p

tmorehouse:
So if server side rendering is used, can we actually use  the DOM function `this.$refs.content.querySelector(...)`?  I haven't played much with server side rendering.

pi0:
@thomasfowler As long as it is not happening during first render (computed and data bindings) SSR should be safe. I'll check that before release too.

PR #280, opened 2017-04-24
Add default b_UID to dropdown

tmorehouse:
Assigns a default `b_UID` to the dropdown if `id` not specified in props.

### DIFF
--- a/lib/components/dropdown.vue
+++ b/lib/components/dropdown.vue
@@ -74,7 +74,7 @@
                 default: false
             },
             id: {
-                type String,
+                type: String,
                 default: null
             }
         },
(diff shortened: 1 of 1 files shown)


SirLamer:
Solid suggestion, I will add the same to Popover

tmorehouse:
I wonder if most components should have a default b_UID added to the outer container element?

SirLamer:
Yeah I agree, this is a really clean solution, however Popover specifically is my pet. ;-)

tmorehouse:
@pi0,  what do you think about adding default `b_UID`s (if no ID provided) to all components?

tmorehouse:
Regarding default b_UIDs, I have created issue #281 for discussion.

tmorehouse:
I'm going to switch how the IDs are assigned in this, to make sure they are rendered after mounted/updated.

tmorehouse:
I will close this and refactor when in the lifecycle the ID is assigned

PR #282, opened 2017-04-25
Allow container-fluid inside jumbotron

tmorehouse:
Adds Boolean prop `container-fluid` to allow for fluid container inside jumbotron

http://v4-alpha.getbootstrap.com/components/jumbotron/

### DIFF
--- a/lib/components/jumbotron.vue
+++ b/lib/components/jumbotron.vue
@@ -1,6 +1,6 @@
 <template>
     <div :class="['jumbotron',fluid?'jumbotron-fluid':null]">
-        <div class="container">
+        <div :class="containerFluid ? 'container-fluid' : 'container'">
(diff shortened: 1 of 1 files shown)


lib/components/jumbotron.vue
```diff
@@ -16,6 +16,10 @@
                 type: Boolean,
                 default: false
             },
+            containerFluid: {
```
pi0: please rename it to just `fluid` to keep consistent with other components :)

tmorehouse: There is already another `fluid` prop for the main jumbotron div though.

The jumbotron itself can be fluid (`jumbotron-fluid`), and the container inside can also be fluid (`continer-fluid`).

pi0: Oops ;) You are right :)

PR #285, opened 2017-04-25
Iterate over body.children for event listeners

tmorehouse:
My mistake,  Fixing how event listeners are attached/detached from `document.body.children`


pi0:
Merging however it may be subject to change in future, as relying on global state is not just a good idea for library components.

tmorehouse:
Yeah, it is not ideal, but looks like the only fix for IOS devices at the moment (based on Bootstrap's code)

Issue #284, opened 2017-04-25
Dropdowns not closed


milenon:
Dropdowns are not closed if dropdown item with 'to' properties are used:

```
                <b-dropdown text="Dropdown Button" variant="success" class="m-md-2">
                    <b-dropdown-item to="#">Not closed after click</b-dropdown-item>
                    <b-dropdown-item to="/real/uri">Not closed after click</b-dropdown-item>
                    <b-dropdown-item href="/some/link">Closed after click</b-dropdown-item>
                </b-dropdown>
```

tmorehouse:
Are you using the vue-router with this?

milenon:
Yes

tmorehouse:
I've created a fiddle https://jsfiddle.net/tmorehouse/obagdgLg/1/ that can be played with, which reproduces the issue.

tmorehouse:
So it appears for router-links (specified by the `to` prop), the click event is not propagating up to the `router-link`.  The resultant HTML looks fine though

```html
<div class="m-md-2 dropdown btn-group">
  <button class="btn btn-success dropdown-toggle" aria-haspopup="true">Dropdown Button</button>  
  <!---->
  <div tabindex="-1" class="dropdown-menu">
    <a href="#" class="dropdown-item" tabindex="0">to #</a>
    <a href="#" tabindex="0" class="dropdown-item">href #</a>
    <a href="/foo" class="dropdown-item" tabindex="0">to /foo</a>
    <a href="/bar" class="dropdown-item" tabindex="0">to /bar</a>
    <a href="/foo" tabindex="0" class="dropdown-item">href /foo</a>
    <a href="/bar" tabindex="0" class="dropdown-item">href /bar</a>
    <a href="/real/uri" class="dropdown-item" tabindex="0">Not closed after click</a>
    <a href="/some/link" tabindex="0" class="dropdown-item">/some/link</a>
  </div>
</div>
```

tmorehouse:
looking in the latest `b-link` code, I think this might be fixed:

```js
        methods: {
            click(e) {
                if (this.disabled || this.href === '#') {
                    e.preventDefault();
                    e.stopPropagation();
                }
                if (!this.disabled) {
                    this.$emit('click', e);
                    this.$root.$emit('shown::dropdown', this);
                }
            }
        }
```

tmorehouse:
Hmmm looking at this a little deeper, it appears he above b-link code doesn't fix this.

[...]

PR #286, opened 2017-04-26
Router-link dropdown workaround for #284

tmorehouse:
This PR is a bit hacky, but it does get around the dropdown not closing for router-link issue #284 

It works by watching the `$route` value for changes and closes any open dropdowns (which would only be the one with the current `router-link` clicked) by emitting an event. **Caveat**: if you click a dropdown-item that is the current route, the menu will not close, unless you click outside the menu, hit esc, or chose another menu item.

Also, since `router-link` can't be disabled, this changes the `router-link` to an `a` via the computed `componentType` prop if `disabled` is true.

Short and sweet, but not perfect nor ideal, and might not be the best solution.

The watcher is in the dropdown mixin for now, but the watcher could be in `dropdown-item`, where it would have access to the `computedType` property and conditionally trigger the dropdown close emit only if it is indeed a `router-link`. _Although in the dropdown mixin we could just do a `this.visible = false` rather than emit the event._

Here is a rough hacked together example: https://jsfiddle.net/tmorehouse/0xLpLq2g/

Feedback is welcome. :)

### DIFF
--- a/lib/components/link.vue
+++ b/lib/components/link.vue
@@ -16,7 +16,7 @@
     export default {
         computed: {
             componentType() {
-                return (this.$router && this.to) ? 'router-link' : 'a';
+                return (this.$router && this.to && !this.disabled) ? 'router-link' : 'a';
             },
             hrefString() {
                 if (this.to) {
--- a/lib/components/link.vue
+++ b/lib/components/link.vue
@@ -16,7 +16,7 @@
     export default {
         computed: {
             componentType() {
-                return (this.$router && this.to && !this.disabled) ? 'router-link' : 'a';
+                return (this.$router && this.to) ? 'router-link' : 'a';
             },
             hrefString() {
                 if (this.to) {
--- a/lib/mixins/dropdown.js
+++ b/lib/mixins/dropdown.js
@@ -32,6 +32,10 @@ export default {
         });
     },
     watch: {
+        $route() {
+            // Handle lack of router-link event bubbling
+            this.$root.$emit('shown::dropdown', false);
+        },
         visible(state, old) {
             if (state === old) {
                 return; // Avoid duplicated emits
--- a/lib/mixins/dropdown.js
+++ b/lib/mixins/dropdown.js
@@ -31,7 +31,39 @@ export default {
             }
         });
     },
+    watch: {
+        visible(state, old) {
+            if (state === old) {
+                return; // Avoid duplicated emits
+            }
+
+            if (state) {
+                this.$root.$emit('shown::dropdown', this);
+                /*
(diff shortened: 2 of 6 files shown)


tmorehouse:
Instead of  watcher, once could also use one of the router injected component options:
- `beforeRouteEnter`
- `beforeRouteUpdate` (added in 2.2)
- `beforeRouteLeave`

To emit the close trigger (either in `beforeRouteEnter` or `beforeRouteUpdate`). If vue-router is not installed then these callbacks wouldn't get called.

lib/mixins/dropdown.js
```diff
@@ -32,6 +32,10 @@ export default {
         });
     },
     watch: {
+        $route() {
+            // Handle lack of router-link event bubbling
+            this.$root.$emit('shown::dropdown', false);
```
tmorehouse: This could just be:

```js
  this.visible = false;
```

pi0: Yep it would be better

lib/components/link.vue
```diff
@@ -16,7 +16,7 @@
     export default {
         computed: {
             componentType() {
-                return (this.$router && this.to) ? 'router-link' : 'a';
+                return (this.$router && this.to && !this.disabled) ? 'router-link' : 'a';
```
pi0: It makes one potential bug. router-link automatically assigns active class to current route so when item is disabled it will loose active class.

tmorehouse: Possible,

Although the Bootstrap dropdown styling is also applied (making it look like menu items as well)

And I suppose the user could `v-if` the dropdown item to take it away if they are on the current route or want the item disabled. instead.  If the4 user uses `v-show="false"` I might have to alter the CSS selector for the keyboard navigation to include `[style*="display:none"]):not([style*="display: none"]` (or I could just document the ways to change the selector for these special cases).

I could remove this line for now, and just deal with the hiding of the menu.

pi0: Active class is added to router-link's attributes which highlights current item. so i think removing is just a good idea

pi0:
Thanks, seems good :)

tmorehouse:
Yeah, maybe we could just document some of the "known issue" caveats of using router links in dropdowns

tmorehouse:
After a bit more investigation, I may have found a better solution rather than watching $route for changes

The @click.native="..." will intercept the click on <router-link>, but will not for regular a or button tags.

Adding both @click="..." and @click.native="..." at the same time in the b-dropdown-item does appear to work.

https://jsfiddle.net/tmorehouse/fv6btm53/

Disabled to items still change the route though. disabled href do respect the disabled state.

PR #288, opened 2017-04-26
Fix a typo

DaleSeo:
Corrected a wrong property name. (`calss` => `class`)

### DIFF
--- a/examples/form-select/demo.html
+++ b/examples/form-select/demo.html
@@ -2,7 +2,7 @@
 
     <b-form-select v-model="selected"
                    :options="options"
-                   calss="mb-3"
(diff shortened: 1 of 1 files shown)


DaleSeo:
I have no idea how my little fix has caused the tests to fail.
Did I miss something?

SirLamer:
Probably a glitch, sometimes the unit tester fails on its own accord, don't sweat it.

tmorehouse:
Looks like a couple of minor things (dangling comma in dropdown.vue and two `let`s in dropdown mixin that should be `const`s)., So its not directly related to your PR.

PR #290, opened 2017-04-26
Better dropdown router-link click handling

tmorehouse:
Better workaround for #284, which handles closing the dropdown when the clicked `<router-link>` is also the current route.

Previous iteration of this fix used a `$route` watcher. This uses the `@click.native="click"` handler to listen for the click event.  `@click="click"` is still needed for non `<router-link>` dropdown items

This solution appears a bit cleaner,

Hacked together example: https://jsfiddle.net/tmorehouse/fv6btm53/

### DIFF
--- a/lib/components/dropdown-item.vue
+++ b/lib/components/dropdown-item.vue
@@ -7,6 +7,7 @@
        tabindex="-1"
        role="menuitem"
        @click="click"
+       @click.native="click"
     >
         <slot></slot>
     </a>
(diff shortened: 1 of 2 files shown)


lib/components/dropdown-item.vue
```diff
@@ -7,6 +7,7 @@
        tabindex="-1"
        role="menuitem"
        @click="click"
+       @click.native="click"
```
pi0: Nice idea :) Just as a question won't having both makes duplicate events? Why we kept both?

pi0:
Just seen you comment :)

PR #292, opened 2017-04-27
new b-form-input-static component

tmorehouse:
Added a standalone `<b-form-input-static>` component, and refactored `<b-form-input>` to use the new component when `static` prop is `true`

### DIFF
--- a/lib/components/form-input-static.vue
+++ b/lib/components/form-input-static.vue
@@ -13,7 +13,7 @@
         mixins: [formMixin, generateId],
         computed: {
             staticValue() {
-                return this.formatter ? this.formatter(value) : value;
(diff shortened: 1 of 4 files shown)


pi0:
I think now it is safe to remove `staticValue`  and just allow using `value` prop. Also we can add a default slot inside `<p>` like this: `><slot></slot></p>` so users have ability to more customizations

tmorehouse:
The formatter wouldn't get called on the static-input though, so the raw value would be shown.   But maybe it isn't needed in the static-input?

tmorehouse:
@pio, Just realized I was missing the `.vue` extension on the bFormInputStatic component import in form-input.vue

tmorehouse:
PR #293 fixes missing component extension

PR #296, opened 2017-04-27
Display modal as block when visible.

### LINKED ISSUE #276: Modal window does not work
When I copy example (https://jsfiddle.net/pi0/bofh9aaa) locally, it does not work on bootstrap 4.0.0-alpha.6 and bootstrap-vue 0.14.0. I get only a blackout of body.

charkins:
Commit b76a37a84045e39a392ad2f44b2b84fa3cc5d706 flagged the
modal as display block !important, which overrides the inline
display: none added by v-show. The modal is not visible because
opacity is set to 0, but the dialog prevents interaction with
components below the dialog.

If the goal is to display:block when visible (v-show will remove
the inline display:none style), then the CSS could just select
.modal.show.

It is possible that this will fix #276

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -70,8 +70,8 @@
     }
 
     /* Make modal display as block instead of inline style, and because Vue's v-show deletes inline "display" style*/
-    .modal {
-        display: block !important;
+    .modal.show {
+        display: block;
     }
 </style>
 


charkins:
Sounds good. I used .modal.show to be more restrictive as I wasn't sure what the exact issue was that led to !important being added in the first place.

Thanks!

charkins:
I think I misunderstood your comment above. You said "at current release .modal {display:block} isn't !important". I thought you meant that this was removed and no longer present in master, but it appears you meant that !important wasn't present in the current release (v0.14.0).

The !important is what appears to be breaking things for me (running bootstrap-vue from master). The problem is that .modal { display: block !important; } overrides the display:none inline style added by v-show. It is not visible because the opacity is 0, but it breaks everything on the page.

Here is an example using a fork of the jsfiddle mentioned in #276, with the bootstrap-vue.css and bootstrap-vue.js pointing to the version generated from current master (with the display: block  !important): https://jsfiddle.net/o0xb0pom/3/

The "Launch Demo Modal" button is unclickable. If you inspect the page, you'll see that it's being covered by the modal which has display:block and opacity:0.

mosinve:
Yeah, you right,  I see now)  Then, do you think adding show selector will fix this?

mosinve:
Playing a little with modal jsfiddle....  `.modal.show{display:block!important}` slightly changes fadein/fadeout transition. So modal need to be displayed as block before it gets show class to get transition work as expected. And i begin to thought, that adding !important wasn't needed, and the issue that lead to this change is result of improper use of bootstrap-vue: It's intended to be used instead bootstrap js & jquery, not with them. Bootstrap itself make changes to display state of modal.. and it may interfere withbootstrap-vue actions.

tmorehouse:
What about using the `modal-open` class on `body` (i.e. `body.modal-open modal.show` as a selector)?  I haven;t gone over the bootstrap-vue modal code, so I am not sure of how this implementation works.

mosinve:
I'll check it, thanks)

mosinve:
And. Just remember,  Bootstrap sets some class on body when showing a modal,

PR #297, opened 2017-04-27
Fix text slot on nav-item-dropdown.

charkins:
The use of v-html on the element wrapping the text slot prevents
it from being used, even when v-html is bound to null. Move the
v-html to a span inside the slot to fix.

### DIFF
--- a/lib/components/nav-item-dropdown.vue
+++ b/lib/components/nav-item-dropdown.vue
@@ -8,9 +8,8 @@
            :aria-expanded="visible"
            :disabled="disabled"
            @click.stop.prevent="toggle($event)"
-           v-html="text"
         >
-            <slot name="text"></slot>
+            <slot name="text"><span v-html="text"></span></slot>
         </a>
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks :)

tmorehouse:
could the `v-html` be placed on a `<template>` object instead (just to prevent extra markup if not required)?

```html
<a .....>
  <slot><template v-html="text"></template></slot>
</a>
```

or will `v-html` not work on `<template>` elements?

pi0:
@tmorehouse i don't think as template is just a placeholder.

tmorehouse:
I know `v-if` and `v-for` work on `<template>` to group a bunch of items together in a single conditional/loop, but wasn't sure about `v-html`

tmorehouse:
I just did a quick test, and unfortunately `v-html` on `<template>` doesn't work :(

charkins:
Yup, no v-html on template unfortunately and no way to interpolate html without escaping, so there isn't an easy way to avoid introducing an additional element here. Ideally, vue would leave innerHtml intact if v-html is null.

PR #298, opened 2017-04-28
b-carousel-slide: Hiding content for some viewport

gpyh:
I didn't meet the timeline I promised but here is the feature you requested!

I went for the most descriptive property name I could think of. If you want me to put `visible-size` as you suggested feel free to ask and I'll change it.

The setting is given for each slide. This may be unoptimal performance-wise but the implementation is very straightforward and, would a user want it for some reason, it allows to set different sizes for different slides in the same carousel.

Also, if you explain to me how to edit the documentation, I'd be happy to do so.

### DIFF
--- a/lib/components/carousel-slide.vue
+++ b/lib/components/carousel-slide.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="carousel-item" :style="{background,height}">
         <img class="d-block img-fluid" v-if="img" :src="img" :alt="imgAlt">
-        <div :class="{ 'carousel-caption': !!caption }">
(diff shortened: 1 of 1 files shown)


gpyh:
One this I forgot: if the size given isn't a bootstrap breakpoint, it will silently fail and *never* show the content. I don't if this is desirable or not. If you want the failing case to *always* show, feel free to ask and I'll change this.

pi0:
@gpyh Feel free opening a new PR if you wish. We can add a type validator for property. (type: ['md','sm'] ...)

PR #299, opened 2017-04-28
b-modal: Remove !important in display: block

gpyh:
Revert back the change from https://github.com/bootstrap-vue/bootstrap-vue/commit/eaf42348ea3808f290c75ef0a662d6685d55a598 and its previous commit.

I don't know what was the rationale for setting `!important` but it definitely caused a bug.
The modal's div was indeed hidden but overlayed the whole page content as `display` couldn't be set to `none`. In my particular case it caused every mouse input over it to be silently captured. This is one of the unintended behaviours it may have caused.

Paging @mosinve to discuss this.

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -71,7 +71,7 @@
 
     /* Make modal display as block instead of inline style, and because Vue's v-show deletes inline "display" style*/
     .modal {
-        display: block !important;
+        display: block;
     }
 </style>
 


pi0:
@mosinve Sorry bothering again 😁 Please make final decision on this. I'll release new patch as soon as this PR resolves :))

mosinve:
Maybe will be useful to write somewhere that bootstrap-vue is not recommended to work with BootstrapJS ? I think they may conflict somehow...

pi0:
@mosinve Agree. Both README.md and SETUP.md are good candidates.

tmorehouse:
Yeah, as well not many of the native BootstrapJS custom/name-spaced events are fired either, if external libraries are relying on that.

mosinve:
At least we have modals breakdown... and I really don't know what else may be broken

mosinve:
We should test somehow the case of simultaneous use of Bootstrap's JS and Bootstrap-vue.

Issue #307, opened 2017-04-30
<b-form-fieldset>Horizontal forms: option for label justify-right</b-form-fieldset>


alexsasharegan:
### Enhancement
When working with forms and the `b-form-fieldset` component, it would be nice to have a little more control over the label. At this point, all I really need is the ability to justify-right the label text for a tighter look on input labels.

Maybe there could be a prop: 
```js
props: {
  labelTextAlign: {
    type: String,
    required: false,
    default: 'left'
  }
}
```

and then a computed prop:
```js
computed: {
  labelAlignClass(){
    switch (this.labelTextAlign.toLowerCase()) {
      case 'center':
        return 'text-center';
      case 'right':
        return 'text-right';
      case 'left':
      default:
        return 'text-left';
    }
  }
}
```

Hopefully this could be useful for many!

PR #308, opened 2017-04-30
Enhancement for issue #307

alexsasharegan:
Support for text alignment classes on the form input label per issue #307.

Didn't see a test runner in the package.json besides linting, so I assume circle-ci will test this upon submission.

### DIFF
--- a/lib/components/form-fieldset.vue
+++ b/lib/components/form-fieldset.vue
@@ -2,7 +2,7 @@
     <div :class="['form-group','row',inputState]">
         <label :for="target"
                v-if="label"
-               :class="['col-form-label',labelLayout]"
+               :class="['col-form-label',labelLayout,labelAlignClass]"
                v-html="label"
         ></label>
         <div :class="inputLayout" ref="content">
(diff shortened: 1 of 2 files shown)


lib/components/form-fieldset.vue
```diff
@@ -68,6 +79,11 @@
                 type: Number,
                 default: 3
             },
+            labelTextAlign: {
+                type: String,
+                required: false,
+                default: 'left'
```
pi0: Please don't make left default, it breaks global alignment in RTL projects.

lib/components/form-fieldset.vue
```diff
@@ -35,6 +35,17 @@
...
             },
+            labelAlignClass() {
+                switch (this.labelTextAlign.toLowerCase()) {
+                    case 'center':
+                        return 'text-center';
+                    case 'right':
+                        return 'text-right';
+                    case 'left':
+                    default:
+                        return 'text-left';
```
pi0: Also default would be just null :)

pi0:
Thanks!

alexsasharegan:
Cheers! 🍻

lib/components/form-fieldset.vue
mosinve: Null сauses "TypeError: Cannot read property 'toLowerCase' of null" when labelTextAlign isn't specified at component level.....

mosinve:
This PR is a breaking change to my project.

PR #311, opened 2017-05-01
updated vue version to 2.3

bossbossk20:
updated vue version to 2.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "bootstrap": "^4.0.0-alpha.6",
     "tether": "latest",
     "uglify-js-harmony": "^2.7.5",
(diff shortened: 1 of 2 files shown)


pi0:
Thanks @bossbossk20. Actually i am aware of new release. But we may delay unfortunately this change for a while (maybe 2.4) as 2.3 release has breaking SSR changes and we can not just force users to upgrade and breaking their projects. Anyway 
- currently bootstrap-vue dist files are fully compatible with 2.3 and are independent of parent project vue version (you can see jsfiddles as they are using vue:latest) 
- If users really need Vue>2.3 they can easily add dependencies to their project's package.json, `^2.2.6` is semver compatible with `2.*` :)

PR #317, opened 2017-05-02
Fix: Issue 307

alexsasharegan:
Fixing the `toLowerCase` call on default value null. Reference issue #307

### DIFF
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -48,7 +48,7 @@ import {bAlert, bBtn} from 'bootstrap-vue/lib/components'
 <script src="//unpkg.com/bootstrap-vue@latest/dist/bootstrap-vue.js"></script>
 ```
 
-## Build variants 
+## Build variants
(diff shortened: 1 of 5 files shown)


mosinve:
Just curious, why just don't leave `''` ? As for me you do the same thing but from another side... ;) and with more code...

alexsasharegan:
@mosinve I was using an empty string by default before, but it seems that @pi0 has a preference for empty defaults being `null`. I see you already PRed this, so we don't need to use this PR. I'll let @pi0  decide what fits the lib best—I just wanted to respond to the breaking change.

tmorehouse:
As well, For attributes, using null actually will remove the attribute from the element, where as empty string `''` will leave the attribute on a node but without a value.

alexsasharegan:
Interesting, @tmorehouse . With the object syntax in class bindings, any falsy value will be omitted. Looks like you can nest style objects in the array as well though.

mosinve:
I didn't know this :) okay now i know ))))

PR #326, opened 2017-05-02
Various fixes for Modals

tmorehouse:
Various fixes & tweaks:

- Fixes `onEsc` issue where hitting `ESC` and the modal wasn't clicked initially (focused), the modal wouldn't close (see discussion at #247), by ensuring that an element within the modal (or the modal itself) is focused when modal opens (using `$nextTick`)
- Refactored the `enforceFocus` handler, and only instantiate it when the modal is open (removes it when closed),
- Added in option to restore focus (`returnFocus`) to the original element that triggered the modal opening.  Works with the `v-b-modal` directive automatically (i.e. the element with the `v-b-modal` directive), or can be specified as a prop `return-focus`.  Can be a CSS selector string (i.e. `'#mybutton'`or an `HTMLElement` reference).
- Added in `generate-id` mixin, in case the modal doesn't have an ID, for `aria-labeledby` and `aria-describedby` cross referencing.

For the initial focus on modal open, the modal content is searched for the first focusable element, and if not found defaults to the modal itself. When searching for focusable elements, the following order is used:
- Footer
- Body
- Header
- Modal itself.

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -172,6 +172,9 @@
                 this.body.classList.add('modal-open');
                 this.$emit('shown');
                 this.$emit('change', true);
+                if (typeof document !== 'undefined') {
+                    document.addEventListener('focusin', this.enforceFocus, false);
+                }
                 this.$nextTick(function () {
                     // Make sure DOM is updated before focusing
                     this.focusFirst();
@@ -203,6 +206,9 @@
 
                 // Hide if not canceled
                 if (!canceled) {
+                    if (typeof document !== 'undefined') {
+                        document.removeEventListener('focusin', this.enforceFocus, false);
+                    }
                     this.is_visible = false;
                     this.$root.$emit('hidden::modal', this._id);
                     this.body.classList.remove('modal-open');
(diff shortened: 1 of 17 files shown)


lib/components/modal.vue
```diff
@@ -69,20 +69,32 @@
...
+        'button:not([disabled])',
+        'input:not([disabled])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        'a:not([disabled]):not(.disabled)',
+        '[tabindex]:not([disabled]):not(.disabled)'
+    ].join(',');
+
     export default {
+        mixins: [generateId],
```
pi0: `generateId` usage should be limited to inputs only as is mentioned inside Mixin. Also as of vue 2.3.x we can not trust `_uid` because they are inconsistent between client server in SSR mode. See [here](https://github.com/vuejs/vue/issues/5573) 

> _uid is internal and should not be used as part of the application state

tmorehouse: Ah, yes.  I'll remove that, and make sure `aria-labeledby` and `aria-describedby` don't get set if the ID is missing.

pi0: Thanks @tmorehouse that's so sad that we currently can't have such behavior. I'll make an on/off switch for `generateId`s so non-ssr users can use them optionally until we can find better way generating consistent IDs :)

tmorehouse:
Take a new look over, I think I have removed all generate-id references (including `this._id`)

Issue #333, opened 2017-05-03
BS4/Bootstrap-Vue Accessibility recommendations


tmorehouse:
Bootstrap V4 is addressing some of the accessibility issues: https://github.com/twbs/bootstrap/issues/22549, Of which bootstrap-vue has started to already address.

Here is  little list of items that are complete or need work for aria and/or keyboard navigation (to be updated as we progress through them):

**Components - ARIA / Keyboard Navigation**
- [x] `b-form-radio` input wrapped in label (which negates need for `for` on label)
- [x] `b-form-checkbox` input wrapped in label (which negates need for `for` on label)
- [x] `b-progress` ARIA `progress` attributes
- [x] `b-modal` attributes `aria-labelledby` and `aria-describedby`. Section `roles` & semantic elements. (PR #247)
- [x] `b-modal` Focus first input/button on open (PR #247). Allow page author to specify input to be focused when opened via `@shown` event (PR #378)
- [x] `b-modal` focus the modal content when opened
- [x] `b-tabs` & `b-tab` attributes `aria-controls` and `aria-labelelledby`. Section `roles` (PR #339)
- [x] `b-tabs` & `b-tab` keyboard navigation (PR #339)
- [x] `b-alert` has aria `role="alert"`,  `aria-live` and `aria-atomic` (PR #340)
- [x] `b-form-file` Needs focus styling in custom-file-input mode for keyboard users (PR #1033)
- [x] `b-form-fieldset` attribute `for` when ID supplied on `,form-control`. Has `role="alert"`, `aria-live`, `aria-atomic` on feedback. and `aria-describedby` (PR #340)
- [x] `b-breadcrumb` ~should have role of navigation (PR #340)~ Add `aria-current` to active crumb (PR #526)
- [x] `b-input-group` needs `role="group"` (PR #340)
- [x] `b-button-group` `role="group"`, make new toolbar component  (PR #367)
- [x] `b-button-toolbar` `role="toolbar"` + optional keyboard navigation (PR #367)
- [x] `b-nav` & `b-nav-link` aria attributes (PR #358)
- [x] `b-dropdown` keyboard navigation (PR #274)
- [x] `b-nav-dropdown` `aria-` attributes (PR #358)
- [x] `b-nav-dropdown` open on ENTER or SPACE (see Issue #348) (PR #349)
- [x] `b-nav-toggle` `aria-` attributes  (PR #358, #410, #411, #412)
[...]

PR #339, opened 2017-05-04
[tabs] Add in ARIA and keyboard navigation support

tmorehouse:
Better ARIA referencing and roles for generated elements.

To enable full ARIA compliance,  each `b-tab` **must** have  a document unique `id` assigned to them. If `id`s are not supplied then only the `role`s and `aria-`states are assigned (partial ARIA compliance).  If an optional document unique ID is assigned to prop `buttonId` on `b-tab`, it will be used as the tab controller's `id`, otherwise one is generated based on the supplied `b-tab` `id` prop (or none if no `id`). These `id`s are required to let screen readers know which `tab` controller is controlling which tab-pane.

This PR also allows keyboard navigation LEFT/UP to move to previous non-disabled tab, and RIGHT/DOWN to get to next non-disabled tab. using key modifier SHIFT with LEFT/UP or RIGHT/DOWN will move to the first or last (respectively) non-disabled tab.

As a side effect, this fixes a bug when programmatically calling `vm.prev()` or `vm.next()` which, previously,  wouldn't let you skip past `disabled` tabs.

Address parts of #333

### DIFF
--- a/RELEASE_NOTES_TEMPLATE.ejs
+++ b/RELEASE_NOTES_TEMPLATE.ejs
@@ -4,5 +4,4 @@
 <% if(commit.title.toLowerCase().indexOf('eslint')!==-1) return; %>
 <% if(commit.title.toLowerCase().indexOf('update')==0) return; %>
 * __<%= commit.title %>__ (<%= commit.authorName %>) - <%= commit.sha1 %>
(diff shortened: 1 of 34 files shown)


pi0:
I did merge blind without local tests :)
@mosinve Your idea on this would be so much useful.

PR #349, opened 2017-05-06
[button-group] Toolbar keyboard navigation + new button-group-dropdown component

tmorehouse:
Addes ARIA and optional keyboard navigation support to `toolbar` variant. **By default keyboard navigation is disabled**, and must be explicitly enabled by setting prop `toolbarKeyNav` to `true` (only applicable to `toolbar` variant), See caveat.

Tabbing into toolbar will focus the first non-disabled visible item. pressing LEFT or UP will focus previous non-disabled item in toolbar. pressing RIGHT or DOWN will focus next non-disabled item in toolbar.  SHIFT+LEFT/UP will focus the first non-disabled item in the toolbar, while SHIFT+RIGHT/DOWN will focus the last.

Also introduces new component `b-button-group-dropdown` for button group dropdowns (`split` button not supported)

**Caveat:** Toolbars with text or text-like inputs (via nested `input-group`s) introduce some issues with Keyboard navigation.  When a text (or text-like) input is focused, pressing LEFT/UP/RIGHT/DOWN will not move the text cursor, but instead will jump to previous or next item in the toolbar, making editing text slightly more difficult.  This is why keyboard navigation is disabled by default.  When enabling keyboard navigation, one should avoid using text or text-like inputs in toolbars.


pi0:
Thats so nice @tmorehouse 💯

PR #356, opened 2017-05-08
[scrollspy] Documentation

tmorehouse:


tmorehouse:
Feel free to edit or change to another directory location

tmorehouse:
@pio I'm not sure if a separate directory is needed for directives, or if this should be moved into the regular "components" directory.

PR #358, opened 2017-05-08
Additional ARIA on navs and dropdown

tmorehouse:
also removed dependency on generate-id

### DIFF
--- a/lib/components/collapse.vue
+++ b/lib/components/collapse.vue
@@ -81,7 +81,7 @@
                 el.style.height = null;
             },
             emitState() {
-                this.$root.emit('collapse::toggle::state', this.id, this.state);
+                this.$root.$emit('collapse::toggle::state', this.id, this.state);
             }
         },
         created() {
--- a/lib/components/collapse.vue
+++ b/lib/components/collapse.vue
@@ -91,6 +91,7 @@
                 }
                 this.toggle();
             });
+            this.emitState();
         },
         mounted() {
             this.emitState();
--- a/lib/components/collapse.vue
+++ b/lib/components/collapse.vue
@@ -91,7 +91,6 @@
                 }
                 this.toggle();
             });
-            this.emitState();
         },
         mounted() {
             this.emitState();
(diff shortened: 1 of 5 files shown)


lib/components/collapse.vue
```diff
@@ -78,16 +79,22 @@
...
         },
-
         created() {
             this.$root.$on('collapse::toggle', target => {
                 if (target !== this.id) {
                     return;
                 }
                 this.toggle();
             });
+            this.emitState();
```
pi0: We are emitting in mounted. So why this is needed ?

tmorehouse:
Hmmm 

```
rollup -c build/rollup.config.js 
  (vue plugin) Unexpected token (1:1156)
undefined (1:1156)
```

Not sure where this error is originating from.

lib/components/collapse.vue
tmorehouse: Just in case the receiving components that are already in document need to reflect the state (although it does default to false).

lib/components/dropdown.vue
```diff
@@ -30,7 +31,7 @@
         <div :class="['dropdown-menu',{'dropdown-menu-right': right}]"
              ref="menu"
              role="menu"
-             :aria-labelledby="split ? null : _id"
+             :aria-labelledby="id ? (id + (split ? '__BV_toggle_' : '__BV_button_')) : null)"
```
pi0: isn't this the problem?

lib/components/collapse.vue
tmorehouse: Although, since the toggle state does default to false, it can probably be removed from `create()` and left on `mount()`

lib/components/dropdown.vue
tmorehouse: I am think this might be where it is, although I have looked at it 20 times and don't see any typos.

Maybe if I move this to a computed property

tmorehouse: Arg.  now I see it :p

pi0: :D Look and extra closing parentheses

PR #361, opened 2017-05-08
[dropdown-item] Explicit component reference (fixes #360)

### LINKED ISSUE #360: unknown custom element <b-link> found in <DropdownItem>
version 0.15.6

I am using the b-dropdown-item component which throws a warning about an unknown element b-link. I am using it in the navbar. 

I see in the source you are trying to extend b-link

tmorehouse:
Addresses issue #360

### DIFF
--- a/lib/components/dropdown-item.vue
+++ b/lib/components/dropdown-item.vue
@@ -17,6 +17,7 @@
 
     export default {
         extends: bLink,
+        components: {bLink},
         methods: {
             dropdown_click(e) {
                 if (!this.disabled) {


lib/components/dropdown-item.vue
```diff
@@ -17,6 +17,7 @@
 
     export default {
         extends: bLink,
```
tmorehouse: Do we need `extends` here?

pi0: I think we can use a smarter refactoring approach by just extending and adding `dropdown-item` class to bLink.

Issue #363, opened 2017-05-08
b-nav-item: Option for plain link when using vue-router


alexsasharegan:
How do I use the navbar component's `<b-nav-item>` to generate a non-router-link component? Before I was able to use the `href` attr instead of the `to` attr to get plain links vs. router-links (respectively).

My project mixes router-controlled links with real links regularly, so this is a must for me. I'm pretty sure some of the last changes switched this up on me. I've been looking for a vue-router solution, but everyone just says use a normal `<a>` tag. I'm trying to avoid rewriting all the bootstrap markup that this lib does for me so well though!

Here's the gist of the navbar:

```vue
<template>
    <b-navbar toggleable fixed="top" type="inverse" variant="inverse">
        <b-nav-toggle target="hd-navbar-content"></b-nav-toggle>
        
        <b-link class="navbar-brand">
            <a href="https://artisanhd.com">
                <img src="/hd-assets/images/artisanhd-logo-light.svg" alt="ArtisanHD" height="40">
            </a>
        </b-link>
        
        <b-collapse is-nav id="hd-navbar-content">
            <b-nav is-nav-bar>
                <b-nav-item to="/customer/">My Account</b-nav-item>
                <!-- These links need to defer to the browser navigator -->
                <!-- ignore the url prefixing... it's for a reverse proxy, not a vue-router thing -->
                <b-nav-item :href="'/uploadconfigure/' | applyUrlPrefix">Start Order</b-nav-item>
                <b-nav-item :href="'/customer/login' | applyUrlPrefix">
                    <span v-if="isLoggedIn">Logout</span>
                    <span v-else>Login</span>
                </b-nav-item>
            </b-nav>
        </b-collapse>
    </b-navbar>
</template>
```

tmorehouse:
This looks like it might be related to #362

alexsasharegan:
Yeah, does look related. As the link component sits now, either `to || href` props and the presence of `this.$router` will turn it into a `<router-link>`.  There's no escape hatch for a standard link tag.

tmorehouse:
Yeah, that is what I was thinking.  It should only be a router link if `to` is present I would think.

And if neither are present it should not provide any href or two (although not sure how the browser will treat an anchor tag with out an href (as it becomes an anchor rather than a link, so it might require role="link" and explicit click/enter/space handlers when no href or to are supplied.).

alexsasharegan:
[...]

Issue #362, opened 2017-05-08
Dropdown regression on from version 0.15


MarZab:
I've got a bug on all my dropdowns from version 0.15; Clicking anything makes all the options active and does not complete the action. The dropdown is unresponsive afterwards.

![snip_20170508173419](https://cloud.githubusercontent.com/assets/1311249/25828764/553c93b6-3452-11e7-80a4-c5e9ca218c67.png)

tmorehouse:
Could you provide a bit of sample code, or a jsfiddle? Is this embedded in a nav? and which flavour of the dropdown are you using? `b-nav-dropdown-item`, `b-dropdown`, or `b-button-group-dropdown`

MarZab:
Latest packages, stock bootstrap latest 4 and this:

`// global css

import "./main.scss"

import Vue from 'vue'
import VueRouter from 'vue-router'
Vue.use(VueRouter);

// bootstrap
import BootstrapVue from 'bootstrap-vue';
Vue.use(BootstrapVue);

// router
let routes = [
    {
        name: 'dsf',
        path: '',
        component: {
            'template': `<b-dropdown text="Dropdown" variant="success" class="m-md-2">
        <b-dropdown-header>This is a heading</b-dropdown-header>
        <b-dropdown-item>Action</b-dropdown-item>
        <b-dropdown-item>Another action</b-dropdown-item>
        <b-dropdown-divider></b-dropdown-divider>
        <b-dropdown-item>Something else here...</b-dropdown-item>
    </b-dropdown>`
        }
    },

];
let router = new VueRouter({routes:routes});

/* eslint-disable no-new */
let vueinstance = new Vue({
    el: '#app',
    router: router,
});
`

removing vue-router fixes the issue.

MarZab:
Fiddle: https://jsfiddle.net/7htvb7gh/1/

tmorehouse:
In the fiddle I see the following menu generated:
```html
<div role="menu" class="dropdown-menu">
 <h6 tabindex="-1" class="dropdown-header">This is a heading</h6>
 <a href="/7htvb7gh/1/show/#" class="active dropdown-item" aria-disabled="false" tabindex="-1" role="menuitem">Action</a>
 <a href="/7htvb7gh/1/show/#" class="active dropdown-item" aria-disabled="false" tabindex="-1" role="menuitem">Another action</a>
 <div role="separator" class="dropdown-divider"></div>
[...]

PR #365, opened 2017-05-09
[link] Allow both router links and regular links to co-exist in same document

tmorehouse:
In some cases one would like to have both regular links (i.e. links to an in-page ID) and router links (routed by vue-router).

If `to` is not specified, then assume a regular local link (not handled by router).

Addresses #362 and #363  (I think... discussion is welcome)

### DIFF
--- a/lib/components/link.vue
+++ b/lib/components/link.vue
@@ -1,7 +1,6 @@
 <template>
     <!-- When VueRouter is available -->
-    <a v-if="routerAvailable"
-       is="router-link"
+    <a v-if="routerAvailable && to"
        :active-class="activeClass"
        :exact-active-class="exactActiveClass"
        :disabled="disabled"
@@ -52,15 +51,11 @@
             },
             _to() {
                 if (!this.routerAvailable || this.disabled) {
-                    return '#';
+                    return null;
                 }
 
                 if (this.to) {
                     return this.to;
-                }
-
-                if (this.href) {
-                    return this.href;
                 }
             }
         },
--- a/lib/components/link.vue
+++ b/lib/components/link.vue
@@ -1,6 +1,7 @@
 <template>
     <!-- When VueRouter is available -->
     <a v-if="routerAvailable && to"
+       is='router-link'
        :active-class="activeClass"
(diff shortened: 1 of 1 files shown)


lib/components/link.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <!-- When VueRouter is available -->
-    <a v-if="routerAvailable"
+    <a v-if="routerAvailable && to"
```
pi0: Can't we refactor `&& to` into computed property?

lib/components/link.vue
```diff
@@ -52,16 +52,12 @@
             },
             _to() {
                 if (!this.routerAvailable || this.disabled) {
```
pi0: Then this condition won't happen  as `routerAvailable` is false when no `to` prop exists :))

tmorehouse:
@pi0, check the new changes and see what you think?

pi0:
LGTM :)

tmorehouse:
Hopefully this doesn't break something else :P

pi0:
Something is breaking Nav links. (Documentation website is broken)

tmorehouse:
Is documentation site using latest release or master?

pi0:
Testing locally on master. I think condition should be `return this.$router && this.to` :))

tmorehouse:
Agh, lol..  an easy fix (I hope)

pi0:
Yep, i'll publish fix soon

MarZab:
Shouldn't in the case neither href or to provided, the link be rendered as `<a role="button>"` since bootstrap does support this troughout? Related to: https://github.com/bootstrap-vue/bootstrap-vue/issues/362

tmorehouse:
So in the case of a `dropdown-item`, having both HREF and TO blank should render a `button.dropdown-item` instead of a link.

tmorehouse:
BS V4.alpha.6 `.dropdown-items` support both using links or buttons (although there are slight styling issues between the two font-wise).  Maybe the dropdown-item component should create a `<button class="dropdown-item>` when both HREF and TO are falsy? rather than using a link tag?.

PR #366, opened 2017-05-09
[button-group] Toolbar keynav fixes

tmorehouse:
A couple of capslock typo fixes.

### DIFF
--- a/lib/components/button-group.vue
+++ b/lib/components/button-group.vue
@@ -36,7 +36,7 @@
                 ];
             },
             isKeyNav() {
-                return this.toolbar && this.toolbarkeyNav;
+                return this.toolbar && this.toolbarKeyNav;
             }
         },
         props: {
@@ -56,7 +56,7 @@
                 type: String,
                 default: null
             },
-            toobarKeyNav: {
+            toolbarKeyNav: {
                 type: Boolean,
                 default: false
             }


tmorehouse:
I am wondering if the toolbar feature should be moved into it's own component, since it is really a wrapper for groups of button-groups?

See BS V4 example:
http://v4-alpha.getbootstrap.com/components/button-group/#button-toolbar

pi0:
Do you mean for accessibility features?

tmorehouse:
Somewhat.

A toolbar will never have buttons as direct descendants. It will always have aa child button group(s) and maybe input group(s)

tmorehouse:
maybe a `b-button-toolbar` component, or `b-toolbar` component.

pi0:
Never personally needed that but this is a official Bootstrap component and It would be useful :)

tmorehouse:
Yeah.  It is part of the `button-group` documentation of BS V4, but really it is is't own wrapper/component that would have button-group and input-group's as children components.  And then remove the 'toolbar' prop option from the `button-group`

tmorehouse:
I think the component wrapper should be `b-button-toolbar`, to follow the BS4 class of `.btn-toolbar`

tmorehouse:
I can whip up a new component in a few minutes.

PR #367, opened 2017-05-09
New component b-button-toolbar

tmorehouse:
Wrapper component for creating toolbars comprised of button-groups and/or input-groups.

Direct children elements should be `b-button-group` and `b-input-group` components.

Allows for optional keyboard navigation by setting prop `keyNav` to `true` (defaults to `false`)

### DIFF
--- a/lib/components/button-group.vue
+++ b/lib/components/button-group.vue
@@ -1,50 +1,23 @@
 <template>
-    <div :class="classObject"
-         :role="toolbar ? 'toolbar' : 'group'"
-         :tabindex="isKeyNav ? '0' : null"
-         @focusin.self="focusFirst"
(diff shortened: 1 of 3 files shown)


tmorehouse:
Should I remove the keyboard navigation and toolbar option from b-button-group?

pi0:
I think so. As button-group is just a helper for styling. Also it would be great if you can provide some docs. (Simply copy paste docs/components/button-group as a starting point)

tmorehouse:
Will do.

I'll remove the toolbar option from the `b-button-group`, and all the extra fluff code for key-nav.

tmorehouse:
I've removed the toolbar features of `button-group`.  It should merge OK once the other lint bugs have been resolved.

PR #368, opened 2017-05-09
[button-toolbar] docs

tmorehouse:


docs/components/button-toolbar/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Button toolbar",
+  "component": "bButtonToolbar"
+}
```
pi0: you can also provide jsFiddle later after release :)

PR #372, opened 2017-05-09
Fix previousTab keybinding.

charkins:
The key bindings specified prevTab function, but the function is
actually named previousTab.


tmorehouse:
This should be good to merge.

PR #378, opened 2017-05-10
[modal] Do not change focus if activeComponent is in modal content.

charkins:
This allows focus to be set in @shown handler of a modal without
the focusFirst handler overriding.


tmorehouse:
@pi0 , I think this would be a good addition.

PR #379, opened 2017-05-11
allow Close button in Modal to be hidden

andrewharvey:


lib/components/modal.vue
```diff
@@ -48,7 +48,7 @@
 
                         <footer class="modal-footer" ref="footer" v-if="!hideFooter">
                             <slot name="modal-footer">
-                                <b-btn variant="secondary" @click="hide(false)">{{closeTitle}}</b-btn>
+                                <b-btn variant="secondary" @click="hide(false) v-if="!hideClose">{{closeTitle}}</b-btn>
```
tmorehouse: Missing a closing quote on `@click="hide(false)`

lib/components/modal.vue
```diff
@@ -161,6 +161,10 @@
                 type: Boolean,
                 default: false
             },
+            hideClose: {
```
tmorehouse: I wonder if this would be better called 'okOnly'?

andrewharvey:
> Are you thinking of this for use case as an alert (with only OK button)?

Exactly.

I'll make the 2 changes you've suggested. okOnly is probably better.

tmorehouse:
Cool.

I'll let @pi0 review. He should be back online in a few hours.

pi0:
Thanks!

PR #386, opened 2017-05-12
Allow Semantic tags

tmorehouse:
For some components, it is nice to be able to change the root element tag type from default of `div` (i.e. to `section`, `article`, `ul`, `li` etc.

Also, for components with ID prop, the ID should appear on the root element as well (as Vue strips out component props as attributes if not explicitly added to the root element)


pi0:
Maybe we can add semantic tags support to other components too.

tmorehouse:
Yeah... I don't think all components need it, but a few more might.

Issue #385, opened 2017-05-12
How to show empty state if table array is empty when filter?


lavezzi1:
Hi there!

In my case I want to show empty state if user search something with `input v-model="filter"` and found nothing. I see `_items` is private. Thanks!

tmorehouse:
I wonder if a named slot for `empty` and a slot for `emptyFiltered` could be created (and conditionally shown based on the length of `this._items` and `this.items` ?

The slot could be set to have a `td`  with `colspan` set to of the number of fields available.

It would also be nice to have a named slot for  `tfoot` element (or conditional `tfoot`)

lavezzi1:
> I wonder if a named slot for empty and a slot for emptyFiltered could be created (and conditionally shown based on the length of this._items and/or this.items ?

Sounds like a good idea!

tmorehouse:
I have a prototype that I will be testing.  and will generate a PR later.

tmorehouse:
Now available in the latest developments release.

lavezzi1:
@tmorehouse `:colspan="fields.length"` you can't get length of an object. You should do something like `:colspan="Object.keys(fields).length"`

tmorehouse:
Oops,  true, used to dealing with the items length.

tmorehouse:
I have fixed this in the PR #389

tmorehouse:
Should be all fixed is latest master branch.

PR #387, opened 2017-05-12
[tables] Allow slots for empty result sets

tmorehouse:
added two slots:
- `empty` to be shown if no items (`items.length === 0`)
- `emptyFiltered` to be shown if a filtered result has no records

and one prop
`showEmpty` to enable showing the empty slots

One method to address issue  #385

There might be better ways to do this though.  comments/changes welcome.


bastones:
It is not very clear [in the documentation](https://bootstrap-vue.js.org/docs/components/table/#items-record-data-) that the `showEmpty` prop MUST be set on the `b-table` component in order for the `empty` and `emptyFiltered` slots to work. Is this the case?

I'll be happy to update the documentation to make this clearer for others, but just want to double-check this is the intended behaviour.

emanuelmutschlechner:
I think documentation is fine. `show-empty` is listed in the "Component Reference" of `<b-table>` and it's usage is shown in the "Complete Example".

bastones:
Is it too much hard work to add a few words to the row in the table under the "Slots" heading to make developers aware the `empty` and `emptyfiltered` slots won't work if the `show-empty` prop is not set?

I will even do this myself if I know the pull request will be accepted, but I stand by what I said - it is not clear and good documentation should be clear about technical behaviour. It is only after using the component and wasting time tracking down why the slot wasn't working that I now know why this happens: it's because I can see there is a default message that gets displayed which the `empty-text` prop or `empty` can replace, but again only if the `show-empty` prop is set.

arippberger:
+1 for @bastones comment on adding info to the documentation. It doesn't make it clear `show-empty` needs to be set to true.

PR #389, opened 2017-05-12
[table] allow tfoot clone of thead + ARIA for sorting

tmorehouse:
Allow user to enable duplicating table thead in tfoot (prop `footClone`

ARIA attributes on header

Additional table style options, and header/footer style variants


pi0:
Sorry for delay

PR #390, opened 2017-05-13
[dropdown] support click event on dropdown menu item

makkrnic:


cynecx:
Wouldn't it be better if we could support the click event with every components which uses the b-link component? I am not sure but it worked before but upgrading to a newer release, everything broke, and not to support the click event with basic components seems ridiculous.

Is it possible for you, if you could also add this for b-nav-item component?

pi0:
It is already fixed in Blink (https://github.com/bootstrap-vue/bootstrap-vue/commit/3e9b860d05d8dec8b7f91eb530e9949866dc1966) Sorry for inconvenience releasing new changes soon.

cynecx:
@pi0 I am using master and it doesn't seem to work.

pi0:
@click event is emitted up so it should be working if you currectly build. Please tests against 0.15.7 and open a Issue with JSFiddle so i can check it.

pi0:
Should be fixed by 650799f @0.15.8

makkrnic:
Hello, sorry for not responding sooner, and thank you for such a prompt response (and fix).

PR #397, opened 2017-05-16
[table] missing this in computed style

tmorehouse:


tmorehouse:
Was missing "this" in the table computed style property

Issue #396, opened 2017-05-15
Carousel error


iiilock3iii:
Carousel keeps giving the "TypeError: Cannot read property 'id' of undefined" error on the github and jsfiddle sites

tmorehouse:
I will take a look at this shortly.

tmorehouse:
@iiilock3iii , this should be fixed in the latest master branch, and will be available in the next release.

PR #398, opened 2017-05-16
[carousel] Fixed this arg in forEach

tmorehouse:
Should address issue #396


pi0:
foreach is calling inside an arrow function so why we need this ? :))

tmorehouse:
Hmm, this is the only spot I could figure that the error "couldn't get property of Id from undefined" might be coming from.

pi0:
I think function should bound to this. But if this work around works feel free merging it  :)

tmorehouse:
Ah, I found root cause I believe. `querySelectorAll` returns a nodelist, which is not a true array (so it doesn't have all the nice sugar of an Array)

PR #402, opened 2017-05-16
added invisible for fields

happilymarrieddad:
Adding invisible as a property to fields so you can dynamically save fields per a user. This allows for a different data table per a user.


pi0:
Thnaks

tmorehouse:
Should this also be applied to the individual items on each row as well?  It appears to only apply to the header at the moment.

Also `visibility:hidden` visually hides the content of the cell, but leaves the cell "space" there.  Is this the intended result? or did you want the column to be removed from the document? (i.e. `display:none`)

happilymarrieddad:
I wanted the column completely removed from the table. I didn't realize I missed the column itself. Sorry man

tmorehouse:
I have a PR  that applies the invisible class to the rest of the column cells.

But note that bootstraps invisible class only sets the visibility to hidden (which is different than display none).

You can also physically remove a column by removing the field entry for that column, and you can also set the field's class property to any class that hides/styles a column. The fields class will get applied to all cells in the column.

Check out the reader in the /docs/components/table directory for updated information on table field and item properties.

Issue #400, opened 2017-05-16
Modal: dead code?


alexsasharegan:
I was tweaking the modal component to figure out if a bug was on my end or not, and I came across three sections of code that don't appear to run.

1. [modal.vue L99-102](https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/modal.vue#L99-L102)
2. [modal.vue L110-122](https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/modal.vue#L110-L122)
3. [modal.vue L306-308](https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/modal.vue#L306-L308)

I've tried inspecting a modal component at runtime, and it looks like the `model` attribute is just thrown away, cutting off the usage of this code. Is it dead code, or does serve a purpose?

PR #403, opened 2017-05-16
[Issue #400] Modal Dead Code removal

alexsasharegan:
If this is, in fact, dead code, here is a removal PR.


pi0:
@alexsasharegan `visible` is not a dead code! It is used to sync visiblity and user can toggle modals with changing `visible` model ;)

alexsasharegan:
Touché. I've been hunting down why the modal sometimes doesn't get the 'show' class added to the main `.model` element, and I couldn't figure out where this was used. Is there any Vue.js documentation on writing components that implement `v-model`? I looked, but came up 404.

alexsasharegan:
There's really no example on how to use v-model to show the modal. I'm hacking away, but no luck.

pi0:
https://vuejs.org/v2/guide/forms.html

alexsasharegan:
Yeah, I've looked that over. I'm not convinced this is implemented correctly. The example from the docs here looks very different: https://vuejs.org/v2/guide/components.html#Form-Input-Components-using-Custom-Events

alexsasharegan:
Is there any chance I could get a working example in a fiddle or something? I can't get this to work for the life of me. I use the crap out of v-model too!

alexsasharegan:
I've even tried watching the change event to check out how this works (since the model is set to emit a change event). @tmorehouse, maybe you can show me an example forked from the docs fiddle? I'm not sure if it's me, or the v-model feature isn't correct.

Issue #329, opened 2017-05-03
Popover doesn't close


lavezzi1:
Is this a bug or a feature? 

![kapture 2017-05-03 at 16 04 57](https://cloud.githubusercontent.com/assets/10686894/25652352/58a9fdfe-301a-11e7-9478-e927c78ac93b.gif)

pi0:
I think this problem is with debouncing period.

SirLamer:
Yes that's right. Trying to support two different event types at once is complicated.

If you force selection of only one event type (ex. triggers="hover") then the debounce feature is disabled and this problem will not occur.

I think I can affect a better solution for this, maybe track the state of each supported event type separate and apply a different filter to each one, rather than the global debounce.

I am planning to revisit Popover big time once Bootstrap 4 switches from Tether to Popper

alexsasharegan:
I'm having some similar troubles with the popover component. I switched to a single-event trigger, and while that helps, there are still problems.

- The tether node created by the trigger is orphaned when the creating component is hidden.
- If a component gets hidden, it doesn't hide its associated popover element.
  - Once the page is scrolled, tether has nothing to reference position, and the orphaned element hugs `0,0`
- Running bootstrap-vue v0.15.8

It's a little too hard for me to try and turn this into a fiddle, but here's the popover component code:
```vue
<template>
    <b-popover :show="address.isEmpty() ? false : null"
               :triggers="['click']"
               placement="bottom"
    >
        <b-btn variant="link" size="sm" class="p-0">
            {{ address.name || address.address_line1 | truncate(16) || 'Unavailable' }}
        </b-btn>
        <i class="fa fa-fw fa-caret-down"></i>
        <hd-plain-address slot="content"
                          :model="address"
                          class="mb-0"
        ></hd-plain-address>
    </b-popover>
</template>
```

![may-15-2017_sm](https://cloud.githubusercontent.com/assets/16269252/26077185/115a4ddc-3970-11e7-8665-1ac32c128c6e.gif)

alexsasharegan:
It looks like it isn't a Tether node that's getting orphaned, it's part of the popover template. Tether must move the node and make it a direct child of the `document.body`. Add a line to `beforeDestroy` and the orphaning problem goes away:

```js
{
  beforeDestroy() {
    this.normalizedTriggers.forEach(item => this.removeListener(item));
    clearTimeout(this._timeout);
    this.destroyTether();
[...]

PR #405, opened 2017-05-16
Issue 329 (partial fix)

alexsasharegan:
As I demonstrated in issue #329, the popover component leaks dom nodes out of Vue's control—presumably due to Tether's mechanics. This PR manages the leak by checking if the popover ref changed ownership to `document.body`, and removing it if so.


alexsasharegan:
Sorry about the git madness. Just trying to stay current with master.

pi0:
That's ok, i always squash merge :))

Issue #399, opened 2017-05-16
First item in dropdown is selected by default


paglias:
When you open the dropdown the first item is selected (or focused, I'm not sure) by default.

<img width="382" alt="schermata 2017-05-16 alle 11 43 05" src="https://cloud.githubusercontent.com/assets/589911/26100012/e32630ee-3a2c-11e7-8a94-b53b278d7f8f.png">

This doesn't seem to happen when an header is the first item in the dropdown
<img width="298" alt="schermata 2017-05-16 alle 11 45 00" src="https://cloud.githubusercontent.com/assets/589911/26100071/1b8b2c5a-3a2d-11e7-9df1-c1e8e82a9405.png">

Hovering other elements doesn't remove the active status but clicking and holding the dropdown toggle button does so my guess is that the first element in the dropdown gets incorrectly focused.

The screenshots are taken from the documentation site so I don't think it's necessary to provide a JSFiddle. Let me know if you need more informations

paglias:
Hmm it looks like it could be the expected behavior https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/mixins/dropdown.js#L89 but I have to say I find it a bit confusing. Especially if you have the same css style for the focus and active state of the dropdown item in which case the first element always look the active one

tmorehouse:
The heading is focus able,  to allow screen readers announce it, otherwise the heading is invisible to screen readers and the user may loose context of what the following items relate to.

tmorehouse:
Maybe if we add additional css to the component to style the headers slightly different when focused. Although, this might change based on when bootstrap v4 moves into beta stage from alpha.

mosinve:
maybe to force a focus only if there is a heading?

tmorehouse:
ARIA best practices states that first item in a dropdown menu should be focuses when the menu opens (and that all items should be accessible, otherwise they become invisible).

Imagine a case like this:

![image](https://cloud.githubusercontent.com/assets/2781561/26154105/91638500-3ae4-11e7-99a2-7ff31fb2a43a.png)

If headings were not "visible/readable" by screen reader, then user wouldn't know context of the "Add" and 'delete" menu actions.

tmorehouse:
The blue box around the focused item is the default for focus (which is different then the hover styling provided by bootstrap (and the hover styling is very subtle)

mosinve:
Ok, in case when headings used it's worth to be.
[...]

PR #408, opened 2017-05-17
[dropdowns] Allow hover to focus item

tmorehouse:
To emulate the way `<select>` focuses/highlights current menu item on mouse hover.

See discussion at #399


mosinve:
Seems working to me, did you test it locally?

tmorehouse:
I tested in a fiddle where I loaded the menu, and then manually added a listener event on the `dropdown-menu` div that grabbed the dropdown items and focused the hovered one.

tmorehouse:
Hmmm. Actually this conflicts with the hover styling, which hides to focus ring while the item is hovered, but does indeed focus the item.

tmorehouse:
I'm not sure if I like how this works visually, as the hover style overrides the focus style, when the item is hovered.

Moving the mouse out of the menu does indeed show that the hovered item was focused:
https://jsfiddle.net/ckrcbk5h/1/

tmorehouse:
If we were to set the `.active` class on the focused/hovered item (and remove it from the others), it actually emulates a select menu more closely.

tmorehouse:
Check out this fiddle using `active` on the focused item: https://jsfiddle.net/ckrcbk5h/5/

image

The item with the blue background has the active class when focused (note this is only on hover at the moment)

pi0:
Thats so nice @tmorehouse  🙌 Thanks @mosinve for helping :)

mosinve:
I'm glad I could be useful. ;)

mosinve:
I like dropdown now )) Thanks @tmorehouse 👍

PR #410, opened 2017-05-17
[collapse] render ID on root div

tmorehouse:
Plus removed a duplicate method from dropdown.js

### DIFF
--- a/lib/mixins/dropdown.js
+++ b/lib/mixins/dropdown.js
@@ -164,21 +164,6 @@ export default {
                 }
             });
         },
-        focusHovered(e) {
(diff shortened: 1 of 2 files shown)


lib/mixins/dropdown.js
```diff
@@ -164,21 +164,6 @@ export default {
                 }
             });
         },
-        focusHovered(e) {
-            if (!this.visible) {
```
pi0: Why this is deleted?

tmorehouse: It was in there twice by mistake (copy/paste error).

PR #411, opened 2017-05-17
[collapse] aria-expanded state

tmorehouse:
`aria-expanded` state should be reflected on control toggle button only. `b-nav-toggle` already reflects the `aria-expanded` state of the targeted collapse.


pi0:
Nice :)

Issue #414, opened 2017-05-18
Navbar: toggleable without breakpoint


alexsasharegan:
Bootstrap has a plain `navbar-toggleable` class, but it isn't supported here. Perhaps this should be the default or when a boolean `true` is passed, but a string passed will trigger the breakpoint string to be concatenated.

PR #415, opened 2017-05-18
Issue/414

alexsasharegan:
This should provide support for the basic `navbar-toggleable` class.

From issue #414


alexsasharegan:
Actually, I need to change this. Null will still trigger the join with a hanging hyphen.

alexsasharegan:
We'll call this WIP.

lib/components/navbar.vue
```diff
@@ -33,7 +36,7 @@
             },
             toggleBreakpoint: {
                 type: String,
-                default: 'sm'
```
pi0: What if `toggleable` is `true` and user not providing `toggleBreakpoint` which is null by default ? :))

lib/components/navbar.vue
```diff
@@ -15,8 +15,11 @@
                     this.variant ? `bg-${this.variant}` : null,
                     this.fixed ? `fixed-${this.fixed}` : null,
                     this.sticky ? 'sticky-top' : null,
-                    this.toggleable ? `navbar-toggleable-${this.toggleBreakpoint}` : null
+                    this.toggleable ? this.toggleClass : null
                 ];
+            },
+            toggleClass() {
+                return ['navbar-toggleable' + this.toggleBreakpoint].join('-');
```
pi0: if `toggleBreakpoint` is null it will return `nav-toggleable-` which is wrong

alexsasharegan: Yeah, I mentioned that above. I put this together a little too hastily between today's hectic schedule.

lib/components/navbar.vue
alexsasharegan: I believe that's the direction I ended up taking this. Let me know!

pi0:
@alexsasharegan Is your work done?

alexsasharegan:
Yes. Tested all the scenarios I can think of, and it looks good!

Issue #413, opened 2017-05-18
Link Support: target and rel attributes


alexsasharegan:
When working with some of the special components that take `to` or `href` attributes _(basically components using the link mixin)_, there isn't support for passing down the `target` and `rel` attributes.

I would love it if the default prop value for `rel` was `rel="noopener noreferrer"` to add a security-boosting default per this article: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/. Maybe that could be triggered with a computed property so it only kicks in if a target other than `_self` is specified.

```vue
<template>
    <b-navbar toggleable
              type="inverse"
              variant="primary">
        <b-nav is-nav-bar>
            <b-nav-item :href="FACEBOOK"
                        target="_blank"
                        rel="noopener noreferrer">
                <i class="fa fa-fw fa-facebook"></i>
            </b-nav-item>
        </b-nav>
    </b-navbar>
</template>
```

tmorehouse:
Rather than putting too much logic into the component, we could just provide new props (`target` and `rel`) to the link component(s), and rely on the user assigning values appropriately for their use cse.

alexsasharegan:
Definitely! Slim and trim!

tmorehouse:
I'll work on a PR for adding the two new props.

alexsasharegan:
Gracias, señor!

pi0:
I agree with troy. But only about `rel=noopener` it is REQUIRED for PWA checklist when `target=_blank` so what's your idea using a computed prop for rel and adding this automatically when no rel provided?

tmorehouse:
Hmmm... so create a computed prop `computedRel` that is based on both `this.rel` and `this.target` (depending on their values or lack of values)?

tmorehouse:
From https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
> Update: FF does not support "noopener" so add this.
> ```js
> rel="noopener noreferrer"
> ```

So should it default to both if `target="_blank"`?

What about cases where a user is linking to their own website (i.e. same domain) and wants `_blank`, but also needs the referer set?

Maybe if `this.rel` is `null` then default the `rel` for target `_blank`, but if `this.rel` is set to an empty string then remove the `rel` attribute?

tmorehouse:
I can see defaulting `rel` for user supplied link content, but if it is author supplied content, then it shouldn't be set (maybe)?

[...]

PR #418, opened 2017-05-18
[links] rel and target attributes

tmorehouse:
Adds `target` and `rel` attributes to all link-based components.

If `target === '_blank'` and `rel` is `null` then default `rel` to `'noopener'`

Addresses #413


tmorehouse:
Does the `computedRel` logic look OK to you @pi0 ?

pi0:
Yep it seems nice :)

alexsasharegan:
I think the `this.target === '_blank'` condition could be eliminated. It doesn't matter which target, you don't want to grant another domain context to your page, right? Or does it not matter because the window will be closed?

pi0:
Sorry i was merged fast :)) So you mean adding `noopener` even to internal links?

tmorehouse:
I didn't apply the rel/target to `breadcrumb` as typically breadcrumbs are in-domain (plus it would mean adding rel and target properties to each item in the list of items

tmorehouse:
The only think I would be timid of, is if someone does want the opener property not set to `noopener` (which they can override by specifying an empty string for `rel`).

Just concerned that it could break some user use cases.

alexsasharegan:
This would allow for any falsey value to drop the rel attribute, but it would probably break anyone depending on that communication as they will be forced to specify that prop now. But how many people use it that way?

```
props: {
        target: {
            type: String,
            default: null
        },
        rel: {
            type: String,
            default: 'noopener noreferrer'
        },
},
computed: {
    computedRel() {
        return this.target ? this.rel : null;
    }
}
```

tmorehouse:
Maybe it should be up to the page author to specify any values, leaving security up to them?

alexsasharegan:
So, I tested the `window.opener` prop on a link opened in the same window, and it is `null`. That makes sense because you just blew away the window by navigating to that link.

alexsasharegan:
So maybe that means your computedRel is perfect the way it is?

tmorehouse:
Note I defaulted `target` to `_self` as well.

alexsasharegan:
As it is now, if I use the conventional `_blank`, I get the `noopener`. If I specify a `rel` prop, I get my rel. If I use a non-standard target, I have to specify the rel to get the security. All seems like reasonable to me.

alexsasharegan:
Ship it! @pi0, you weren't premature on the merge after all.

tmorehouse:
Should `nofollow` also be added (for firefox)?

alexsasharegan:
Hmm... that's primarily an SEO concern, right? Now that we have a `rel` attribute, people have the ability to specify that at least.

tmorehouse:
@alexsasharegan Oops, I meant `noreferrer` not `nofollow`

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
> Update: FF does not support "noopener" so add this.
> ```js
> rel="noopener noreferrer"
> ```

alexsasharegan:
I know that article makes it sound necessary, but when I look at the MDN page for an anchor tag, that attribute doesn't seem necessary. It just wipes your url from the referrer header on the request. Not sure how that is a security risk.

tmorehouse:
True.  So we'll leave it as is for now :)

PR #425, opened 2017-05-19
[table] CSS refactor

tmorehouse:
When `b-table` had prop `small`, the arrows were not aligned top to bottom correctly (up to high).

Also removed erroneous `.` from `table-responsive` class


lib/components/table.vue
```diff
@@ -318,6 +320,29 @@
         opacity: 0.3;
     }
 
+    table.table-sm thead .sorting:before,
```
pi0: Can't we refactor it into a generalized prop selector? This is a LOT! :D

tmorehouse: Yeah, I was thinking the same.

Maybe making asc/desc their own class, and maybe making disabled its own class.

tmorehouse: so have the following classes:
- `sorting`
- `asc`
- `desc`
- `disabled`

tmorehouse: I can try and refactor before the merge.

tmorehouse: Also not sure why pagination is being imported, as it is not used directly by table.

tmorehouse:
CSS has been refactored for much smaller footprint

pi0:
Nice +1

Issue #423, opened 2017-05-19
Modal Property close-on-backdrop doesn't work.


Catrionaa:
When setting the modal property "close-on-backdrop" to false, When clicking on the backdrop it still closes.

tmorehouse:
THis might just be a simple fix in the handler.

tmorehouse:
Hmmm this is strange.  The logic is there to prevent clocking when `close-on-backdrop="false"`, but it is still closing.  And `close-on-esc="false"` is also not working (and it has the logic there as well).

tmorehouse:
Ah, even though the two props are set to `Boolean`, when setting the props to `false`, the actual value is being returned by `this.closeOnBackdrop` (and `this.closeOnEsc`) is the string `"false"`, rather than the Boolean value `false`.  It appears Vue.js is not converting the string values of the attributes to boolean.

alexsasharegan:
Oh, that makes sense. I was going to say that it's working for me. If you want to pass anything other than a string, you'll have to use the attribute binding syntax.

tmorehouse:
the `fade` prop is similar, it needs to be `v-bind` as well to disable the transition animation

tmorehouse:
So not really sure if this could be classified as a bug, but rather "needs clarity" in documentation.

I can add a few extra tests for prop `!== 'false'` to get around the need for binding.

tmorehouse:
Or we change the props to be `noCloseonEsc` and `noCloseOnBackdrop` (and default to `false`).  This might be a breaking change for some users though.

alexsasharegan:
For booleans, the presence or absence of the prop also generated the corresponding boolean.

tmorehouse:
Yeah, and with  default of `true` it is not possible to rely just on presence or absence.  I can add in the extra test for `!== 'false' as a simple workaround.

alexsasharegan:
Oh, true is the default! Yeah, I guess in hindsight, it would have been better to use a negative attribute to enable that syntax.

alexsasharegan:
Maybe pin that one for major release with breaking changes.

tmorehouse:
Would also have to change `fade` to `noFade` (defaulted to `false`).

@pi0 what are your thoughts on changing the boolean props (which default to `true`) to negative props (defaulting to `false`)?

tmorehouse:
[...]

PR #426, opened 2017-05-19
[modal] boolean prop handling

tmorehouse:
Changed `closeOn...` to `noCloseOn...` and `fade` to `noFade` (all defaulting to `false`) for better boolean prop handling *(this is a braking change compared to current usage)*

Tweaked computed ID's to align with other components

Addresses usage issue #423


tmorehouse:
We can reduce footprint a bit more by removing default of `false` from Boolean props, but would need patch to docs generation to handle showing `false` default for Booleans that have no default (i.e. `null`)

tmorehouse:
Is there a way to include a note in the release info about usage/breaking changes? i.e. that props have been changed?

pi0:
@tmorehouse After release i think you can edit release notes :) If not just provide me the notes.

tmorehouse:
OK Cool, will try that.

PR #432, opened 2017-05-20
Tests/card

alexsasharegan:
Adds testing for the card component.


alexsasharegan:
P.S. Should we disable linting on the tests? I didn't really adhere to the lint standards of the rest of the lib because all the test specs fail linting as well.

alexsasharegan:
P.P.S. Feedback welcome. I work in a shop that doesn't do **any** testing. I would love to get a little guidance on how to write good tests.

__tests__/components/card.spec.js
```diff
@@ -1,6 +1,83 @@
...
+        const { app: { $refs, $el } } = window
+
+        const refs = ['simple_card', 'standard_card', 'img_card', 'img_overlay_card']
+
+        refs.forEach((ref) => {
+            const childNodes = [...$refs[ref].$el.childNodes]
+            const cardBlock = childNodes
+                .find(el => el.classList && el.classList.contains('card-block'))
+
+            expect(cardBlock).toBeDefined()
```
pi0: Not sure if it covers `null` value (not.tobeNull) :))

__tests__/components/card.spec.js
```diff
@@ -1,6 +1,83 @@
...
+            .find(el => el.classList && el.classList.contains('card-block'))
+
+        expect(blockEl.classList.contains('card-img-overlay')).toBe(true)
+    })
+
+    it('should contain text content', async() => {
+        const { app: { $refs, $el } } = window
+
+        expect($refs.simple_card.$el.textContent).toContain('Simple Card')
+        expect($refs.standard_card.$el.textContent.trim()).toContain('Last updated 3 mins ago')
```
pi0: Why trim?

pi0:
Nice work @alexsasharegan :) I will merge it. Also Feel free accepting join invitation (https://github.com/bootstrap-vue) ;)

__tests__/components/card.spec.js
alexsasharegan: Yeah, wasn't sure how strict the `toContain` equality check was. A lot of those nodes have new-lines, etc. Noted that it isn't necessary!

__tests__/components/card.spec.js
alexsasharegan: Well, [Array.prototype.find returns the element or undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find?v=control)
> The find() method returns the value of the first element in the array that satisfies the provided testing function. Otherwise undefined is returned.

So that should still be good, yes?

Issue #394, opened 2017-05-14
Table Serveside fetch pagination


Diaskhan:
Hie
There have a server side pagination ??
I have tried items-provider function but it does n work for me !!
My ajax returns [{"id":2,"name":"When died napaleon","text":null,"parent":0,"type":1,"typename":"число"}]
If it works, Could U provide some example ?

        <b-table class="table table-striped table-bordered"  :items="items"  :items-provider="fetch" :fields="fields" :current-page="currentPage" :per-page="perPage" :filter="filter">
            <template slot="isfolder" scope="item">
                <i v-if="item.value==true" class="fa fa-folder"></i>
                <i v-if="item.value==false" class="fa fa-question-circle"></i>
            </template>
            <template slot="name" scope="item">
                <a v-on:click="navigate(item.item)">{{item.value}} </a></i>
            </template>
            <template slot="actions" scope="item">
                <b-btn size="sm" @@click="details(item.item)">Details</b-btn>
            </template>
        </b-table>

    <script>

    var appq = new Vue({
        el: '#apptable',
        data: {
            items: [],
            fields: {
                id: {
                    label: 'ID' //,                     sortable: true
                },
                name: {
                    label: 'Наименование',
                    sortable: true
                },
                actions: {
                    label: 'Действия'
                }
            },
            currentPage: 1,
            perPage: 10,
            filter: null
        },
        mounted: function () {
        },
        methods: {
            details(item) {
                alert(JSON.stringify(item));
            },

            fetch({ perPage, currentPage, filter, sortBy, sortDesc }) {
                var self = this;
                return axios.get('/question/getquestions/6')
                    .then(function (response) {
                        return response.data;
                   });
            },

        }
    })

    </script>

tmorehouse:
For server side pagination, it is possibly best to handle pagination outside of the `b-table` component (and pass the paging parameters to your provider function via variables scoped to your module or data on your app.

Then pass your apps method (which takes advantage of your app's paging state stored in your data) and pass it as the `itemsProvider`  function.

Diaskhan:
Actually I did it As U sad !!
Tnhks for feedback !!!
If I provide Some Example into documentation, Could U apply commit  ??

Diaskhan:
[Example.zip](https://github.com/bootstrap-vue/bootstrap-vue/files/1019227/Example.zip)
I did it that way ! 

But there are some ussue, 

[...]

Issue #131, opened 2017-03-05
[Tables] Pagination


yvandupre:
i would like to know if table will support pagin client, server?

pi0:
@yvandupre Unfortunately no as it may need external dependencies and we cannot add them to this repo.

yvandupre:
seem to be a local pagin....

so we feed the table (50 data), if we display only 10 data, when we on other, there are not another call to server...

seem a litle bit less powerfull then
https://github.com/wenzhixin/vue-bootstrap-table

pi0:
@yvandupre Thanks for your suggestion and also #158 table API is now exposing `items-provider` prop. With this one we can easily implement any ajax behavior. Currently i had not time for providing more guides but i'll add them soon. Functions blueprint is `provider(ctx)` which ctx is a reference to internal component. We have to apply pagination and sorting using `ctx.perPage` `ctx.currentPage`  `ctx.filter` `ctx.sortBy` and `ctx.sortDesc` then returning results items.

mosinve:
While checking fieldset noticed a table pagination bug. It appears when the filter is set.
steps to reproduce:
1. open table docs or playground
2. set filter to show less than min perpage value (<5)
3. set different values of rows per page
4. check that pagination changes according to rows per page value , not taking in account filtered amount of rows

tmorehouse:
For the provider routine, one could use promises within the `fetch(....)` to handle the async, and then return the result of the promise (say using a Promisifed `XmlHTTPRequest` (like axion or similar), although the user might need to load a Promise polyfill depending on the browser.

Sort of in this wein of pseudo code:

```js
methods: {
  fetch({perPage,currentPage,filter,sortBy,sortDesc}) {
    return get('/url',params).then(data => {
      // massage data and then return it
      return data;
    }
  }
}
```

PR #433, opened 2017-05-20
[table] Improved documentation

tmorehouse:
Added more detail about scoped slots and  supported `item` and `field` properties


alexsasharegan:
Is the `fields` prop supposed to be an object?

docs/components/table/README.md
```diff
@@ -17,28 +17,107 @@ age: {
...
+| state | String | Bootstrap contextual state applied to row (`active`, `success`, `info`, `warning`, `danger`)
+
+
 ### Custom rendering
 Custom rendering for each field is possible using **scoped slots**.
 If you want to add an extra field which does not exits on records, just add it to `fields` array.  Example:
- 
+
+```js
+ fields: [
```
alexsasharegan: Are the fields supposed to be an array of objects, or just an object?

tmorehouse:
Oops, yeah.  I'll fix that

docs/components/table/README.md
tmorehouse: Oops, yeah. I'll fix that.  Thanks for catching that! :+1:

alexsasharegan: Days since last Troy-o: `0`

tmorehouse:
I'm wondering if the item prop `state` should be `_state` to differentiate it from record values?

We could make `state` deprecated and start to use `_state`.

Also, it appears the default `toString` method for filtering/searching includes the `state` variable.  We probably should filter that out of of the item before converting it to a string for filtering.

As well, filtering only looks at the raw data, and doesn't include any virtual computed columns.  This might need to be noted in the documentation.

alexsasharegan:
Hmm... I was looking into what you mentioned, and I came across something else. Does the default function for the prop `field` actually return an object? Arrow function shorthand to return an object literal is ambiguous due to the code block syntax. Take a look here: https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/table.vue#L96-L97

image
image

alexsasharegan:
I'm kind of a constructor hater for Arrays and Objects, but it's probably better than this: 

image

alexsasharegan:
So I can see how state might collide with a user value. I'm not a fan of using private notation, but how about `bState` and the html equivalent `b-state`?

Also, I don't see how the toString method incorporates the `state` since it uses that `sortBy` variable to pluck of the appropriate item field.

Good call on the itemProvider. Looking through the source was the first I saw of it.

alexsasharegan:
P.S. `key` on line 249 is undefined

tmorehouse:
I'll do  quick commit to add key to the fieldClass method

tmorehouse:
fieldClass is now fixed with comit https://github.com/bootstrap-vue/bootstrap-vue/commit/2900e953a49b1941be8481e1bac89cf646249666

tmorehouse:
@alexsasharegan The filter function converts the item row object to a string at line 220 https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/table.vue#L220

So anything in the item object gets converted to a string representation (i.e. including the `state` property (if it exists).

alexsasharegan:
Ah, I see that now. Is there a way to confine that to the value prop name?

tmorehouse:
About the default for fields, yeah, the `{}` is ambiguous in `default: () => {}`

It might be better to make it better/clearer if it was:

 `default: () => { return {}; }`

 or

 `default: () => new Object()`

tmorehouse:
We could use the fields object keys to select the item keys to stringify (i.e. by making an intermediate object to stringify)

tmorehouse:
It mght also be nice to allow computed fields in the items array (i.e. an anonymous function), but that might be too much bloat on the items array (and the user could run their own massage routine on the items using the `itemsProvider` function.

tmorehouse:
also, trying to figure out what the prop `item_key` is for.  It is bound to each `tr` in the table body.  And it doesn't follow normal camelCase / kebab-csae naming.

alexsasharegan:
Lots of good ideas! And you've got your hands full with this one 😃 

Is this a good time to bring up that it's kinda weird to have the `<tfoot>` above the `<tbody>`?

tmorehouse:
Yeah, I used to always poot my `<tfoot>` at the bottom of the table, but based on the W3 html5 standards the `<tfoot>` follows directly below the `<thead>`.

I guess this is because, in some printing situation, the `<tfoot>` (and `<thead>`) can appear on multiple pages if the table is longer than a page. (although I haven't stumbled across any browser that does that).

alexsasharegan:
Welp, learn something new every day!

tmorehouse:
And regarding itemsProvider, it might be handy to add in a `Boolean` prop to signal that paging is done in the `itemProvider` function. maybe `providerPaging`.  This would allow `b-table` to maintain the state of paging, but not use it when creating `_items`.  As well, filtering could be done via the provider function as well (for server side filtering) if another `Boolean` prop. is set (say `providerFilter`).

A watch method could see if any of the paging or filter values have changed and then freshen the `items/_items` array.

Could address issues like #394 and #131

tmorehouse:
@pi0 & @alexsasharegan 

Just thinking, `state` prop on item is really a `variant` class for the `tr`.

maybe `state` could be set as `_variant` (semi-private item prop) to follow with the other places we use `variant` as a component prop.

`state` could be deprecated fro preference of the `_variant` prop:

```html
<tr v-for="(item,index) in _items"
    role="row"
    :key="items_key"
    :class="[(item._variant || item.state)?'table-'+(item._variant || item.state) : null]"
    @click="rowClicked(item, index)"
>
```

And then a few releases later we could remove `item.state` support.

The row's class could be made into a method for computing the class as well as `rowClass(item)`

tmorehouse:
Or make the `state` property `_rowVariant`, which better reflects that is a row level property

alexsasharegan:
Good catch. I think we should attempt to unify the API around using `variant`. I'm gonna vote for using plain `rowVariant` though just because I think it's super awkward having a public API prop use that notation. One of those _just doesn't feel right_ type things to me. With the extra specificity of the `row` prefix, I think does a good job of both avoiding collisions and improving semantics.

tmorehouse:
To make sure it is not a breaking change, I could use the logic above to prefer `_rowVariant` over `state`.

Issue #431, opened 2017-05-20
Cannot click multiple checkbox if using same v-model


yeungc:
Hello, I just found a problem, if I use same v-model for multiple b-form-checkbox, only one of it can be clicked, please see example:
https://jsfiddle.net/s0Lafhrk/

But if I use native HTML code instead of b-form-checkbox, the problem will be solved:
https://jsfiddle.net/yth2z0r3/

May I know how to fix it? Thanks.

tmorehouse:
I wonder if this is because of how the value is bound to the checkbox inside the component:

```html
<input
       type="checkbox"
       :id="_id"
       :name="name"
       :value="value"
       :disabled="disabled"
       :class="[custom?'custom-control-input':null]"
       :checked="checked===value"
       @change="$emit('change',$event.target.checked?value:uncheckedValue)"
>
```

It might need to be changed to:

```html
<input
       type="checkbox"
       :id="_id"
       :name="name"
       :value="value"
       v-model="value"
       :disabled="disabled"
       :class="[custom?'custom-control-input':null]"
       :checked="checked===value"
       @change="$emit('change',$event.target.checked?value:uncheckedValue)"
>
```

But not sure if that would affect how the `uncheckedValue` is applied

tmorehouse:
The other option might be to create a composite component that makes each individual `b-form-checkbox` value into an array result.

alexsasharegan:
### Vue Examples
[Basic Checkbox](https://vuejs.org/v2/guide/forms.html#Checkbox-1)
[Component implements v-model](https://vuejs.org/v2/guide/components.html#Customizing-Component-v-model)

alexsasharegan:
So I guess checkboxes are implemented as arrays under the hood. Check this craziness out:

https://jsfiddle.net/s0Lafhrk/3/

alexsasharegan:
Still some kinks to work out when unchecking the box.

alexsasharegan:
This works fairly well: https://jsfiddle.net/s0Lafhrk/4/

alexsasharegan:
Yeah, that appears to have full feature parity with the native Vue example @yeungc provided (https://jsfiddle.net/yth2z0r3/)

alexsasharegan:
So the question is, how does vue support v-model being any type?

alexsasharegan:
Alright, this will support the old style, and the array style v-model: https://jsfiddle.net/s0Lafhrk/5/

[...]

PR #434, opened 2017-05-20
Issue/431

alexsasharegan:
Add support for the v-model as an array with the checkbox component. I would say that supporting this is optional, but it does land directly in the vue docs here: https://vuejs.org/v2/guide/forms.html#Checkbox.


alexsasharegan:
This is for issue #431

tmorehouse:
Looks like it should work (I think)

alexsasharegan:
Here is the dev fiddle: https://jsfiddle.net/s0Lafhrk/5/

pi0:
Will test locally before next release ;) BTW seems awesome improvement.

PR #440, opened 2017-05-22
[table] contextual variant classes

tmorehouse:
Deprecated `item.state` for `item._rowVariant`

Added `field.variant` to allow contextual classes on columns.

Improved cell class computation.

Updated documentation to reflect changes/additions


tmorehouse:
Should we add in a filter on the item data (to remove `_rowVariant`) when stringifying the row for filtering?

tmorehouse:
I have added a routine to filter out `state` and `_rowVariant` when stringifying the record for filtering.

Issue #419, opened 2017-05-18
Navbar: supported content


alexsasharegan:
Just looking for feature parity on the BS4 docs for navbar supported content, I see this lib here:

- ✅  `.navbar-brand` for your company, product, or project name.
- ✅  `.navbar-nav` for a full-height and lightweight navigation (including support for dropdowns).
- ✅  `.navbar-toggler` for use with our collapse plugin and other navigation toggling behaviors.
- ❌  `.form-inline` for any form controls and actions.
- ❌  `.navbar-text` for adding vertically centered strings of text.
- ✅  `.collapse.navbar-collapse` for grouping and hiding navbar contents by a parent breakpoint.

Are there ways to implement the two content forms I don't see documented? I can hand-write the html, but it gets weird when you're trying to switch between components that build the BS4 nested structure and your hand-written markup.

tmorehouse:
Would you see these as two new `b-nav-` components?

alexsasharegan:
The navbar-text could be an addition to `b-nav-item` perhaps in the form of a `is-nav-text` attribute. The form might even be doable with the current components, but we should add an example to show the proper way to do it.

alexsasharegan:
Maybe just a simple `v-if="isNavText"` for a `div.navbar-text` and a slot, then `v-else` with the `b-link`?

PR #441, opened 2017-05-23
Navbar extra components

tmorehouse:
Added two new components for navbar usage:

- `b-nav-text` Creates  `.nav-text` container
- `b-nav-form`  Creates an `.inline-form` form container (using `b-form`)

Addresses components mentioned in #419


tmorehouse:
Yeah, they are pretty basic components. ;-)

All the content would be placed into the default slots of these components.

alexsasharegan:
Do you even need to pass the `id`? Shouldn't all the normal html props pass along just fine? Like so:

```html
<b-nav-form id="b-form" method="POST" action="https://example.com/api" />
<!-- should render -->
<b-form id="b-form" method="POST" action="https://example.com/api"></b-form>
<!-- and on down to a plain form element -->
```

alexsasharegan:
Take a look at what I mean here: https://jsfiddle.net/gfqucs0z/52/

I didn't even bother with the id prop declaration. Maybe that's bad because of the docs, but maybe not?

tmorehouse:
For things that need an ID for ARIA stuff, it should definitely be placed on the root element.  But not all components required IDs for ARIA.  In all cases, supplying an ID is optional.  ID prop isn't probably needed on the `b-nav-form` component through

alexsasharegan:
Right, all I was trying to demonstrate that when placing an id on the `b-nav-form`, it propagates to the root component. Thus making it superfluous.

PR #442, opened 2017-05-23
[docs] components - show if prop is required

tmorehouse:
Some components have props that are flagged as required.

This PR adds an extra column to props table to show if property is marked as required

Shows `true` is required, otherwise blank (to not clutter table output)


tmorehouse:
Option: Could show the word 'Yes' instead of `True` for `Required`

alexsasharegan:
Hmm... either is clear, but I think I lean towards "Yes".

tmorehouse:
And should we show 'No' if not required or just a blank?

alexsasharegan:
I like the no clutter approach, but I suppose having the table column empty would look silly. It would be ideal if the column only appears if there is a required prop (then it would show Yes/No).

tmorehouse:
There is the 'invisible' field prop now for `b-table`, which just hides the content of the column (but leaves the space of the column there).

Could get the fields to check if there are any `required`s and set the `invisible` property to `true`

alexsasharegan:
That sounds like cleanest, easiest to understand way.

tmorehouse:
Also switched from word `True` to word `Yes`

alexsasharegan:
Wait, is that... never mind! You just changed it!

alexsasharegan:
I'm gonna say you caught that before committing an official Troy-o

tmorehouse:
LOL

tmorehouse:
Hmmm. Now I am wondering if it might be better to just remove the `required` field from the list of fields rather than just making it invisible and have it take up blank space.

tmorehouse:
The `Required` column will only be rendered if the component has a prop that has a truthy `required`.  And if a prop is required it will show the word 'Yes', otherwise it will be blank.

PR #446, opened 2017-05-23
Bring navbar example up to date

makkrnic:
Dropdown should use `right` instead of `right-alignment`


tmorehouse:
@pi0 the embedded fiddle will need to be updated as well.

pi0:
I've updated embedded one

Issue #315, opened 2017-05-02
Form-file bug on Edge 14


hvalcourtSerdy:
Hi, yesterday our QA tested the bootstrap-vue components on https://bootstrap-vue.github.io/docs (we want to see if we will use bootstrap-vue).

We share the result with you in fews ticket, hope this will help you for your developpement.

a) Go to https://bootstrap-vue.github.io/docs/components/form-file on Edge (Windows 8.1 - Edge 14)
b) Select a file.
c) Bug : The windows always re-open when we click the "X", or "Cancel". The QA had to close Edge with ctrl-alt-delete.
d) Screenshot: 
![2_edge14](https://cloud.githubusercontent.com/assets/25179767/25619067/8363be76-2f17-11e7-9364-d1c65cd7660b.png)

pi0:
#322

hvalcourtSerdy:
Hi, this one is on Edge, not on android webview

pi0:
@hvalcourtSerdy Do you confirm that this one works correctly with pure bootstrap file inputs?

hvalcourtSerdy:
Edge is a supported browser for bootstrap 4 : https://v4-alpha.getbootstrap.com/getting-started/browsers-devices/#desktop-browsers

And the problem is not in Bootstrap wall of browser bug here : http://getbootstrap.com/browser-bugs/

The problem is this one :
"When using a <input type="file"> inside a <label> on MS Edge, the file explorer window keeps reopening. When this happens, the only way to close Edge is by killing the process."
as in https://github.com/facebook/react/issues/7683

And this one : https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8282613/

In the source code here, https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/form-file.vue , the input is inside label

tmorehouse:
This issue should e fixed in the next release.  the `label` wrapper has been changed to a `div`

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #449, opened 2017-05-23
Deprecate (remove) generate-id

tmorehouse:
Removed need for generate ID mixin.

Users must supply an ID on form input (file, radio,input, etc) components if they want ARIA compliance. IF they do not supply ID(s) then many of the aria attributes will not appear (but wont break functionality)

Allow users to override the `b-form-fieldset` `label`'s attribute `for` (needed in special cases like groups of checkboxes). If no `for` prop is supplied looks for first input to associated to.  For radio buttons, it points to the group of radios (`role="radiogroup"`) rather than the first radio button

Fixed bug where `form-file` with `plain` set to `true` wouldn't display the file input.

Fixed bug where MS Edge browser pukes on `<input type="file">` inside `<label>` (issue #315). Fixed by changing `label` to `div`


tmorehouse:
After looking at `form-radio` code, I wonder if a `form-checkbox-group` should be created (which takes an array/object of options)?

tmorehouse:
After this merge, I think we can safely delete `generate-id.js` mixin.

SirLamer:
I was requested to review, but as I have had nothing to do with the form elements to date, I will abstain unless called back explicitly.

tmorehouse:
@SirLamer I added you just because of our original discussion on using auto-ids (but unfortunately Vue SSR foobars it up for us :( ).

SirLamer:
Ohhh okay, well @pi0 said he doesn't want auto-IDs so I kinda forgot about that. ;-)

I think auto-assigning a unique ID for accessibility functionality in the case where an ID is not provided is a good idea, I do that with my own form elements I use in my application.

tmorehouse:
Went through a few more component and did a bit of cleanup.

lib/mixins/generate-id.js
```diff
@@ -4,12 +4,7 @@
  * it is better to prefer references/refs and events over relying on generated id.
  * @see https://github.com/bootstrap-vue/bootstrap-vue/issues/281
  *
- * Currently is being used in:
- *  - dropdown.vue
- *  - form-checkbox-vue
- *  - form-file.vue
- *  - form-input.vue
- *  - form-select.vue
+ * No longer used
```
pi0: So is it is safe to just remove this mixin file ? :D

tmorehouse: I think it is.  I went through most of the components to see if it was referenced anywhere else and didn't s it.

PR #450, opened 2017-05-24
[tooltip] fix for #435

### LINKED ISSUE #435: tooltip hidden but still in document
When I use tooltip, it works, but it cover on the input, make the input can not be focused. Guess it is cause by the style of tooltip
snip20170521_10
pop up
snip20170521_11
pop hidden but still in document
snip20170521_14
pop hidden style, display none didn't work

tmorehouse:
Remove `!important` from `display: block;` css

Fixes #435


tmorehouse:
It appears the `!important` was allowing the hidden tooltip element to "cover" focusable elements, preventing them from being clicked in some instances.

bibodeng:
OK, good job

Issue #452, opened 2017-05-24
Collapse component should have an option to stay open


rdgout:
Hi guys,

I noticed there is no way to make sure the collapse component is open on init. Could it be possible to implement this?
It would be even better if we could manage the initial state with a prop.

AFAIK with regular bootstrap, it should be possible to keep it open.

alexsasharegan:
After looking at the source, it looks like you might be right about not being able to set the initial state to open. For now, you could use the parent component's `mounted` hook to call `toggle` on the `collapse` child via a ref.

@tmorehouse @pi0 , could this be enabled via `v-model` kind of like the modal component?

tmorehouse:
@alexsasharegan that might be an option, or a Boolean prop (read once on init to set the collapsed state)

alexsasharegan:
It's certainly simpler. Maybe the prop name should convey that. Something like `showOnInit`? So many other components use `show` in a more interactive way, so it might be beneficial to distinguish things.

tmorehouse:
Yeah, `showOnInit` would be more clear (and non-reactive)

tmorehouse:
This would be as simple as:

```js
        data() {
            return {
                show: this.showOnInit
            };
        },
        props: {
            showOnInit: {
                type: Boolean,
                default: false
            },
            isNav: {
                type: Boolean,
                default: false
            },
            id: {
                type: String,
                required: true
            }
        },

alexsasharegan:
Vue won't baulk at that right? You know the message about not manipulating props? That should just copy by value plain and simple.

tmorehouse:
Yeah, since `data()` is a function, it is only returned once when the component is instantiated.

tmorehouse:
@alexsasharegan do you want to take a stab at a PR for this?

alexsasharegan:
Sure! I'll take some low-hanging fruit 😉

alexsasharegan:
How come the example file looks different than the fiddle on the docs site?

tmorehouse:
The examples currently do not propagate to JSfiddle, which is why it probably looks different.

The docs are based on current release as well, rather than master.

alexsasharegan:
@rdgout This should be solved in the next release via `v-model` (v0.16.2). There will be an example in the component doc page.

rdgout:
@alexsasharegan Looks good!

rdgout:
[...]

PR #453, opened 2017-05-24
Issue/452

alexsasharegan:
Fix for issue #452 

Allows for a `showOnInit` prop to be passed that will determine the initial collapsed state of the component.

My working "test" here: https://jsfiddle.net/f0qsxezp/14/


pi0:
@alexsasharegan Thanks. Isn't it possible to make a two way bounded model just like modals? This has two advantages that it is possible to control open state using vue reactive data and also checking if collapse is visible or not. Also you can use playground on docs for testing instead if jsfiddle this is much easier to test locally :)

alexsasharegan:
If you want to go with `v-model` here, I'll cook that up! There was some discussion with @tmorehouse about just keeping it simple, but I don't think anybody was dead set on one way vs. another.

alexsasharegan:
So I've got the v-model working, but it creates an interesting edge case.

I wired up the canonical button trigger using the `v-b-toggle` directive, and then wired up another button that toggles the collapse `v-model`. They both work side-by-side, but the `v-model` can become stale. Currently, when you update the `v-model` I don't just toggle the collapse state, but rather check against the current state of collapse and only change if it's different.

I know it's kind of hard to explain this, but does this make sense? Imagine this: you're `v-model` is set to `true` at mounted. Someone clicks a toggle directive button, now the collapse state is collapsed. If you change your `v-model` attribute to `false`, nothing will change. This seems right to me, but also means you can't rely on the `v-model` for the state of the collapse component.

I can't for the life of me think of why you would use both methods _(directive and v-model)_ except setting an initial open state with `v-model` and then let the button do the toggling. The way I have it set up now is the same as the modal `v-model`.

alexsasharegan:
P.S. Got the playground working. Sweeeeeeeet

lib/components/collapse.vue
```diff
@@ -24,19 +24,32 @@
...
         },
 
+        model: {
+            prop: 'visible',
+        },
+
+        watch: {
+            visible(newVal, oldVal){
+                if (newVal !== this.show) {
+                    this.toggle()
```
pi0: I think it would be better to explicitly set `newVal` instead of just toggling. This may help prevent problems you mentioned using both `v-b-toggle` and `v-model` :)

alexsasharegan: I thought about that, but then I would have to re-implement the emitState logic. Thoughts?

pi0: You right. We also need to emit a `change(this.show)` event on emitState too. then calling emitState here. So both methods work like each other.

lib/components/collapse.vue
```diff
@@ -43,10 +43,9 @@
         },
 
         watch: {
-            visible(newVal, oldVal){
-                if (newVal !== this.show) {
-                    this.toggle()
-                }
+            visible(newVal) {
+                this.show = newVal;
```
pi0: Why removed condition guard? (newVal !== oldVal) this helps prevent emitting duplicate events

alexsasharegan: Ah, my bad. Good catch.

alexsasharegan: Not to be picky, but then we have exactly what I had before where I just called `vm.toggle()`

pi0: Yep :D It can be just swapped with this.show = ... . Also we need emitting `change` event on emitState

alexsasharegan: - The watcher will only be called once the value of `visible` has changed
- If we don't want to emit unnecessary events, we check against `show` for a change
- At this point we're performing the same actions in the toggle method

Why not leave it calling toggle to keep the logic consolidated?

pi0: Consider this case:
- v-model is `true`
- collapse closes using `v-b-toggle`  (v-model is still `false`)
- v-model sets to `false` => this makes collapse opening again as we are calling toggle :))

alexsasharegan:
Should the change event just emit the show state? There wouldn't be any reason to emit the id like the internal state event, right?

pi0:
Consider this case:
- v-model is `true` so `visible` is true on mount
- collapse closes using `v-b-toggle` so visibility is `false` now (v-model is still `true`)
- v-model sets to `false` => this makes collapse opening again as we are calling toggle :))

pi0:
> Should the change event just emit the show state? There wouldn't be any reason to emit the id like the internal state event, right?
Agree

alexsasharegan:
I think I'm following. So this is part of the `v-model` magic I've been trying to grasp. When you emit the change event with the new value of the `v-model` prop, Vue takes care of mutating it for you. So I don't "get in trouble" for trying to mutate `visible`, but now we keep it in sync.

Did I get that right?

pi0:
@alexsasharegan `v-model` magic is nothing other than a syntax sugar. so `<v-model=foo` translates to `<value=foo @input='foo=value'` this keeps model being in sync.

tmorehouse:
An aside: The emit state is used to trigger the calling button to set it's 'aria-expanded' state.  Since it is emitted on $root, the ID is used to let the button determine if the event is for itself or not.

alexsasharegan:
Awesome. I get it now. It's always made sense from the parent component's perspective, but I was overcomplicating it from the emitter's perspective.

alexsasharegan:
Hopefully now we got everything covered. Thanks for the patience and assistance, @pi0 !!!

alexsasharegan:
...Okay, I mean NOW. Haha got git confused on that merge

PR #454, opened 2017-05-25
modal: change event order

youaresofunny:
There are no issue about it (but i think we can discus this little fix here)

The chain of events in hide method of modal now looks like this:
`click smth (ok / cancel) -> trigger(change, hidden) -> trigger(ok/cancel)`
It means that 'hidden' (or 'change') is triggered even before we start to process the 'ok' or 'cancel' methods. (so if we e.cancel() in ok, we will anyway see hidden trigger)

After commit it will be work in this way:
`click smth (ok / cancel ) -> trigger(ok / cancel) -> trigger(change) -> trigger(hidden) (if event not canceled)`

Also, can i add "beforechange" emit, that triggers before "ok" / "cancel" ?

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -214,15 +214,15 @@
                     }
                 };
 
-                // Emit events
-                this.$emit('change', false);
-                this.$emit('hidden', e);
-
+                // Emit `ok` and `cancel` events
                 if (isOK === true) {
                     this.$emit('ok', e);
                 } else if (isOK === false) {
                     this.$emit('cancel', e);
                 }
+
+                // Call `change` event after we know what we did
+                this.$emit('change', false);
 
                 // Hide if not canceled
                 if (!canceled) {
@@ -234,6 +234,7 @@
                     }
                     this.is_visible = false;
                     this.$root.$emit('hidden::modal', this.id);
+                    this.$emit('hidden', e);
                     this.body.classList.remove('modal-open');
                 }
             },
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -188,6 +188,7 @@
                 if (this.is_visible) {
                     return;
                 }
+                this.$emit('show');
                 this.is_visible = true;
                 this.$root.$emit('shown::modal', this.id);
                 this.body.classList.add('modal-open');
@@ -214,15 +215,15 @@
                     }
                 };
 
-                // Emit `ok` and `cancel` events
+                // Emit events
+                this.$emit('change', false);
+                this.$emit('hide',e);
+
                 if (isOK === true) {
                     this.$emit('ok', e);
                 } else if (isOK === false) {
                     this.$emit('cancel', e);
                 }
-
-                // Call `change` event after we know what we did
-                this.$emit('change', false);
 
                 // Hide if not canceled
                 if (!canceled) {


alexsasharegan:
So I think there's been some discussion on this in the past. I'm in favor of complying with the existing BS4 events emitted by the modal: 
image

It is important to note that this can't be done with just calling these events in a different order—transition-group hooks have to be implemented. An added difficulty bonus here is that the transition group hooks are called once per transition element. We currently have two transition elements.

youaresofunny:
Well, I'll just add BS4 event system there

tmorehouse:
Also, Vue.js (and boostrap-vue) is reactive based, whereas Bootstrap V4 native is event based.  so there are a few differences in how things can be handled.

Although I do think having some of the added events fire on $root would be handy... The biggest issue is providing contest to the event when it is fired (i.e. providing the ID of the component, a reference to the component, or an event object (with target and maybe relatedTarget attached.

alexsasharegan:
@youaresofunny You can see part of the long-running questions on issue #276

youaresofunny:
Yep, but this commit is not about $root emits :walking_man:

alexsasharegan:
You're right, @youaresofunny , I got carried away on another topic. The order of events seems more sensible with your changes. They don't take into account the asynchronous nature of the transitions, but that is probably best dealt with in a later, breaking change. Now with your event renaming, I don't believe we've made any breaking changes at all, and you should have your event flow worked out.

Any extra thoughts, @tmorehouse ?

tmorehouse:
I wonder if we should standardize the format / calling signature of all component emits on $root?

i.e. this or something like this:
```js
this.$root.$emit('event::name', this.id, this, other, args, here)

this.$root.$on('event::name', id, ctx, other , args, here)
```
Although not all components have an id, so maybe just `this` and the receiving handler could look for a `ctx.id`?

tmorehouse:
For local `vm.$emits()` I don't see too much of an issue with the proposed order.

alexsasharegan:
@tmorehouse, with all those options, why not make a class and just pass that? I've got some sweet class patterns you guys might like, so pass me the prop names and I'll PR the internal event object so we can define an interface and avoid too many parameters.

tmorehouse:
The modal click handler (where it can be canceled) is kind of nifty, although it is binding to a native event structure ($event) which we don't have the luxury of in `$emit`s.  So maybe modeling after the native event might be handy, or we could just create an event from `new Event` and take advantage of that.  although it might have too many props we don't need.

I can see possibly providing 'target' being the VM of the component, and `relatedTarget` could be a reference, if needed to, another element.

Maybe `event.target` (native element ref), `event.$target` being the component VM reference, etc.

alexsasharegan:
Yeah, I was thinking of more of a hybrid event. A class that uses a lot of the same Event names, but it more scoped to the internal needs of the BV components. For example, the emitting component would be placed on `target`, and the modal canceler would be done with a closure trick using a `preventDefault` callback on the SyntheticEvent.

alexsasharegan:
If you've played with react, they have done some cool things with their synthetic event to make it familiar, but ergonomic. We have very small needs, but I thought that was a good target to emulate.

tmorehouse:
Cool.  I think it should have props similar to `cancellable` and `defaultPrevented`, etc.

`relatedTarget`, could be good as well (and optionally set, null otherwise)

PR #455, opened 2017-05-25
Enhancement/button group shorthand

alexsasharegan:
So this is certainly open for discussion, but why not export all the button components using the same shorthand as the button component itself? This PR does just that.


lib/components/index.js
```diff
@@ -54,7 +54,9 @@ export {
     bButton as bBtn,
     bButtonToolbar,
```
tmorehouse: What about adding bBtnToolbar ?

alexsasharegan: Good catch! I'll get that one. What about `bInputGroupButton`?

tmorehouse: Yeah that one is bit long too.   It wouldn't hurt to add it I would think.

PR #456, opened 2017-05-26
[pagination] disabled prop

tmorehouse:
Adds a `disabled` prop to pagination component, which will disable the buttons (links) when set to `true`


lib/components/input-group-addon.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="input-group-addon" :id="id">
+    <div class="input-group-addon" :id="id || null">
```
tmorehouse: Not sure how this got in this PR, but it is OK.

lib/components/pagination.vue
```diff
@@ -230,6 +241,10 @@
             }
         },
         props: {
+            disabled: {
+                type: Boolean,
+                default: 3
```
pi0: Why this is 3 ? :))

tmorehouse: Oops, Typo

PR #457, opened 2017-05-26
components-extra does not work

skyway777:
I've clonned repository, tried to build project with extra-components and realized that extra-components does not work.
Fixed this.


pi0:
@skyway777 Thanks. Anyway there is currently no extra component available in the stock :))

PR #460, opened 2017-05-27
[table] Async providers, events, and enhancements

tmorehouse:
Adds async provider function support (callback and Promise based)

Deprecates `items-provider` in favour of passing function to `items` prop

Adds scoped slot support for table header and footers

Added `busy` prop for triggering table "busy" state (via `aria-busy="true" attribute).

Emits `head-clicked` event when header cells are clicked (when not busy), and a `refreshed` event when the provider function has completed/resolved.

Added `sort-changed` event when sorting criteria has changed

Added extra documentation.

A working prototype example can be found here: https://jsfiddle.net/dyzs5txk/ (complete with header/footer scoped slots). You can select how the items are provided to `b-table` (view the console for the various stages of data retrieval). The example requires a browser that supports Promises (just to emulate the delay in fetching data).

The updated documentation provides greater detail on the new features

image

### DIFF
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -307,6 +307,7 @@ trigger the calling of the provider function.  So be sure to bind to the `per-pa
 **Event based refreshing of data:**
 You may also trigger the refresh of the provider function by emitting the 
 event `table::refresh` on `$oot` with the single argument being the `id` of your `b-table`.
+You must have a unique ID on your table for this to work.
 
 ```js
     this.$root.$emit('table::refresh', 'my-table');
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -35,12 +35,15 @@ Supported field properties:
 ```js
 [
     {
-        _rowVariant: 'success', // Displays record green
+        _rowVariant: 'success', // Displays record row green
         age: 27,
         name: 'Havij'
     },
     {
-        _rowVariant: 'danger', // Displays record red
+        _cellVariants: { 
+            age: 'danger',  // Displayes cell for field 'age' red
+            name: 'success' // Displayes cell for field 'name' green
+        },
         age: 42,
         name: 'Robert'
     }
@@ -52,6 +55,7 @@ Supported optional item record modifier properties (make sure your field keys do
 | Property | Type | Description
 | ---------| ---- | -----------
 | `_rowVariant` | String | Bootstrap contextual state applied to row (`active`, `success`, `info`, `warning`, `danger`)
+| `_cellVariants` | Object | Bootstrap contextual state applied to individual cells. Keyed by field (`active`, `success`, `info`, `warning`, `danger`)
 | `state` | String | **deprecated** in favour of `_rowVariant`
 
 **Note** `state` is deprecated. `_rowVariant`, if present in the record, will be prefered.
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -48,8 +48,9 @@
                 role="row"
                 :class="rowClass(item)"
                 @click="rowClicked($event,item,index)"
+                @hover="rowHovered($event,item,index)"
             >
-                <td v-for="(field,key) in fields" :class="tdClass(field)">
+                <td v-for="(field,key) in fields" :class="tdClass(field, item, key)">
                     <slot :name="key" :value="item[key]" :item="item" :index="index">{{item[key]}}</slot>
                 </td>
             </tr>
@@ -386,9 +387,14 @@
                     field.class ? field.class : ''
                 ];
             },
-            tdClass(field) {
+            tdClass(field, item, key) {
+                let cellVariant = '';
+                if (item._cellVariants && item._cellVariants[key]) {
+                    cellVariant = (this.inverse ? 'bg-' : 'table-') + item._cellVariants[key];
+                }
                 return [
-                    field.variant ? ((this.inverse ? 'bg-' : 'table-') + field.variant) : '',
+                    (field.variant && !cellVariant) ? ((this.inverse ? 'bg-' : 'table-') + field.variant) : '',
+                    cellVariant,
                     field.class ? field.class : ''
                 ];
             },
@@ -407,6 +413,15 @@
                     return;
                 }
                 this.$emit('row-clicked', item, index);
+            },
+            rowHovered(e, item, index) {
+                if (this.busy) {
+                    // If table is busy (via provider) then don't propagate
+                    e.preventDefault();
+                    e.stopPropagation();
+                    return;
+                }
+                this.$emit('row-hovered', item, index);
             },
             headClicked(e, field, key) {
                 if (this.busy) {
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -91,7 +91,7 @@
         const exclude = { state: true, _rowVariant: true };
 
         return toString(Object.keys(obj).reduce((o, k) => {
-          if(!exclude[k]) {
+          if (!exclude[k]) {
             o[k] = obj[k];
           }
           return o;
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -99,9 +99,18 @@
     };
 
     const defaultSortCompare = (a, b, sortBy) => {
-        return toString(a[sortBy]).localeCompare(toString(b[sortBy]), undefined, {
-            numeric: true
-        });
+        if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
+            if (a[sortBy] < b[sortBy]) {
+                return -1;
+            } else if (a[sortBy] > b[sortBy]) {
+                return 1;
+            }
+            return 0;
+        } else {
+            return toString(a[sortBy]).localeCompare(toString(b[sortBy]), undefined, {
+                numeric: true
+            });
+        }
     };
 
     export default {
--- a/__tests__/components/table.spec.js
+++ b/__tests__/components/table.spec.js
@@ -416,10 +416,10 @@ describe('table', async() => {
                         if (sortBy) {
                             expect(spy).toHaveBeenCalledWith(vm.context)
                             expect(vm.context.sortBy).toBe(null)
-                            sortBy = vm.context.sortBy
+                            sortBy = null
                         } else {
                             expect(spy).not.toHaveBeenCalled()
-                            expect(vm.context.sortBy).toBe(sortBy)
+                            expect(vm.context.sortBy).toBe(null)
                         }
                     }
                 })
@@ -453,10 +453,10 @@ describe('table', async() => {
                         if (sortBy) {
                             expect(spy).toHaveBeenCalledWith(vm.context)
                             expect(vm.context.sortBy).toBe(null)
-                            sortBy = vm.context.sortBy
+                            sortBy = null
                         } else {
                             expect(spy).not.toHaveBeenCalled()
-                            expect(vm.context.sortBy).toBe(sortBy)
+                            expect(vm.context.sortBy).toBe(null)
                         }
                     }
                 })
--- a/__tests__/components/table.spec.js
+++ b/__tests__/components/table.spec.js
@@ -598,4 +598,18 @@ describe('table', async() => {
             expect(spy).toHaveBeenCalled();
         }
     })
+
+    it('table_provider should emit a refreshed event for providerArray', async() => {
+        const { app: { $refs, $el } } = window
+        const vm = $refs.table_provider
+        const spy = jest.fn()
+
+        vm.$on('refreshed', spy)
+        await setData(vm,'items', app.providerArray);
+        await nextTick()
+        await sleep(50)
+        expect(spy).toHaveBeenCalled()
+        expect(vm.value.length).toBe(app.items.length)
+    })
+
 });
--- a/__tests__/components/table.spec.js
+++ b/__tests__/components/table.spec.js
@@ -479,15 +479,15 @@ describe('table', async() => {
             vm.$on('input', spy)
 
             // Set page size to be 1 less then number of items
-            setData(app, 'currentPage', 1)
-            setData(app, 'perPage', app.items.length - 1)
-            nextTick()
+            await setData(app, 'currentPage', 1)
+            await setData(app, 'perPage', app.items.length - 1)
+            await nextTick()
             expect(tbody.children.length).toBe(vm.perPage)
             expect(vm.value.length).toBe(vm.perPage)
 
             // Goto page 2, should have length 1
-            setData(app, 'currentPage', 2)
-            nextTick()
+            await setData(app, 'currentPage', 2)
+            await nextTick()
             expect(tbody.children.length).toBe(1)
             expect(vm.value.length).toBe(1)
 
@@ -511,21 +511,21 @@ describe('table', async() => {
             vm.$on('input', spy)
 
             // Set page size to number of items
-            setData(app, 'currentPage', 1)
-            setData(app, 'perPage', app.items.length)
-            nextTick()
+            await setData(app, 'currentPage', 1)
+            await setData(app, 'perPage', app.items.length)
+            await nextTick()
             expect(tbody.children.length).toBe(app.items.length)
             expect(vm.value.length).toBe(app.items.length)
 
             // Apply Fiter
-            setData(app, 'filter', String(app.items[0].name.last))
-            nextTick()
+            await setData(app, 'filter', String(app.items[0].name.last))
+            await nextTick()
             expect(tbody.children.length < app.items.length).toBe(true)
             expect(vm.value.length < app.items.length).toBe(true)
             
             // Empty filter alert
-            setData(app, 'filter', 'ZZZZZZZZZZZZZZZZZzzzzzzzzzzzzzzzzz........')
-            nextTick()
+            await setData(app, 'filter', 'ZZZZZZZZZZZZZZZZZzzzzzzzzzzzzzzzzz........')
+            await nextTick()
             expect(tbody.children.length).toBe(1)
             expect(vm.value.length).toBe(0)
             expect(tbody.children[0].children[0].textContent).toContain(vm.emptyFilteredText)
@@ -550,15 +550,15 @@ describe('table', async() => {
             vm.$on('input', spy)
 
             // Set page size to number of items
-            setData(app, 'currentPage', 1)
-            setData(app, 'perPage', app.items.length)
-            nextTick()
+            await setData(app, 'currentPage', 1)
+            await setData(app, 'perPage', app.items.length)
+            await nextTick()
             expect(tbody.children.length).toBe(app.items.length)
             expect(vm.value.length).toBe(app.items.length)
 
             // Set items to empty list
-            setData(app, 'items', [])
-            nextTick()
+            await setData(app, 'items', [])
+            await nextTick()
             expect(app.items.length).toBe(0)
             expect(vm.value.length).toBe(0)
             expect(tbody.children.length).toBe(1)
--- a/__tests__/components/table.spec.js
+++ b/__tests__/components/table.spec.js
@@ -345,7 +345,7 @@ describe('table', async() => {
                 tr.click()
                 expect(spy).toHaveBeenCalledWith(vm.value[idx], idx)
             })
-        })
+        }
     })
 
 });
--- a/__tests__/components/table.spec.js
+++ b/__tests__/components/table.spec.js
@@ -413,7 +413,7 @@ describe('table', async() => {
                         expect(vm.context.sortBy).toBe(fieldKeys[idx])
                         sortBy = vm.context.sortBy
                     } else {
-                        if (vm.context.sortBy) {
+                        if (sortBy) {
                             expect(spy).toHaveBeenCalledWith(vm.context)
                             expect(vm.context.sortBy).toBe(null)
                             sortBy = vm.context.sortBy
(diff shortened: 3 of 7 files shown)


alexsasharegan:
So this is my first deep dive into the table component—she's a big one! One initial thought: do you think it bears mentioning that if `items` returns a promise, it should not be resolved with the `ctx.callback` as this would lead to some to an unnecessary refresh? Or maybe just a couple tweaks for clarity (the example for the promise ignores the items variable):

```js
function myProvider(ctx) {
     let items = [];
 
     let promise = axios.get('/some/url');
 
     return promise.then((data) => {
         // pluck the array of items off our axios response
        items = data.items

         // must return an array
         return items;
     });
 }
```

Also, I feel like passing the callback as second parameter is more inline with traditional patterns.

lib/components/table.vue
```diff
@@ -107,13 +115,12 @@
                 default: ''
             },
             items: {
-                type: Array,
-                default: () => []
+                type: [Array, Function],
+                default: () => this.itemsProvider ? this.itemsProvider : [] // Deprecate itemsProvider
```
pi0: Maybe we can also throw a console.warn before returning `this.itemsProvider` :)

lib/components/table.vue
```diff
@@ -213,27 +278,49 @@
...
+            },
+            hasProvider() {
+                return this.items instanceof Function;
+            },
             _items() {
-                if (!this.items) {
-                    return [];
-                }
+                // Conditionals
+                const providerFiltering = Boolean(this.hasProvider && !this.noProviderFiltering);
```
pi0: We may convert this to computed props.

lib/components/table.vue
```diff
@@ -283,27 +370,82 @@
...
+                    this._providerSetLocal(data);
+                }
             }
         }
     };
 </script>
 
-
-<style>
+<style scoped>
```
pi0: Scopes are inconsistent in  bundlers. We may simply prefix root div and selectors with something like `.bs-table` class. (Or maybe using [css-modules](https://vue-loader.vuejs.org/en/features/css-modules.html))

lib/components/table.vue
```diff
@@ -348,4 +490,14 @@
...
     }
+
+    /* Busy table styling */
+
+    table[aria-busy="false"] {
+        opacity: 1;
+    }
+
+    table[aria-busy="true"] {
+        opacity: .6;
```
pi0: Nice trick :D

docs/components/table/README.md
```diff
@@ -2,58 +2,73 @@
 
 > For tabular data. Tables support pagination and custom rendering.
 
-### fields
-Fields prop is used to display table columns. 
+### `fields` prop
+Thee `fields` prop is used to display table columns.
```
pi0: Troyo :

lib/components/table.vue
tmorehouse: Sure that would be a good idea.

tmorehouse:
@alexsasharegan I'll change the provider signature to the 2 parameter mode `provider(ctx, cb)`

And will tweak the examples for the promise method

tmorehouse:
Take a look now and see what you guys think.,

alexsasharegan:
There are calls to `_updateProvider` on lines 320 and 416, but I think that's supposed to call `_providerUpdate`, right?

alexsasharegan:
So maybe this is a micro-optimization, but I try to avoid duplicate iteration whenever possible. The record toString setup you have does a filter, then does a reduce. Maybe just put a guard in and only call reduce like so:

```js
return toString(Object.keys(obj).reduce((o, k) => {
  if(!exclude[k]) {
    o[k] = obj[k];
  }
  return o;
}, {}));
```

alexsasharegan:
Is there any reason `item_key` is the only snake case prop instead of camel case? I didn't find `itemKey` used elsewhere.

alexsasharegan:
What exactly is going on between [lines 309 - 315](https://github.com/tmorehouse/bootstrap-vue/blob/table-enhanced/lib/components/table.vue#L309-L315)? I don't think `sortBy` or `sortDesc` ever gets used. We could probably trim off some extra characters with a destructuring statement too.

tmorehouse:
I have no clue what `items_key` is used for. It's a leftover from days gone by

alexsasharegan:
That's weird, it's used in a `v-for` over `<tr>`'s in the `<tbody>`. Correct me if I'm wrong, but it isn't necessary to provide a unique key for non-vue instances.

tmorehouse:
@alexsasharegan lines 309 to 315 ensure that changes to any of those props will trigger a refresh of the _items computed prop (registration of reactivity is triggered by the first get of the property). They are used in conditionals later down in the computed property (in the sorting routines).

tmorehouse:
If no one is using the `items_key` prop, I think we can safely remove it.

alexsasharegan:
Yeah, line 49 is the only one I found. Blast 'em! :boom:

tmorehouse:
Blasted :boom:

docs/components/table/README.md
tmorehouse: Fixed

lib/components/table.vue
tmorehouse: I've used `b-table` to match with the components name

lib/components/table.vue
tmorehouse: Changed to computed props.

lib/components/table.vue
tmorehouse: A deprecation console.warn has been added.

alexsasharegan:
@tmorehouse, I thought registration of reactivity can only happen at initialization (when vue redefines properties with getters and setters) or when a totally new instance is set (and the process is repeated).

This is probably more for my edification, but are you sure getting a reactive prop triggers a refresh? Vue is smart enough to know that when the properties don't change, nothing needs to happen. I was under the impression that a get potentially adds that accessor to a dependency list so it can be notified when a set happens.

Also, when copying non-object properties (things like perPage), a non-reactive primitive is returned. So making a change to that would not have a reactive effect like setting it from the perspective of the object. Like this:

```js
let numberCopy = this.perPage
numberCopy *= 2
// this.perPage is the same, and no reactivity was triggered

// vs. ->
this.perPage *= 2
```

Like I said, this is probably more for me catching up than it is you guys. I'm detail oriented though, so I gotta get 'em straight!

tmorehouse:
@alexsasharegan , Getting a prop doesn't trigger reactivity/refresh, but actually registers the calling function as a "dependant".

This only occurs when the requested prop is **first** accessed by a get.

Take this example:
```js
props: {
  foo: { type: string, default: '' },
  bar: { type: string, default: '' }
},
computed: {
  baz() {
    if (this.foo === '' || this.bar === '') {
       return 'BAZ'
    }
    if (this.bar !== '') {
       return 'BAZ-BAR';
    }
}
```

Now remember that computed property results are cached once evaluated (http://vuejs.org/v2/guide/computed.html#Computed-Caching-vs-Methods) And wont be refreshed until one of the dependants have changed.

How does Viue determine what properties a computed property depends on?  It has to evaluate the computed property at least once (each prop is depends with a getter and setter routine.  each getter registers the calling function in a list of dependents).

So imagine the situation, using the above example code, and none of the props are changed from their defaults of `''` (empty string).

You call/reference the `baz` computed prop.  It runs and returns `'BAZ'` (and caches the result).  It returns `'BAZ`' because `this.foo === ''` is true first, before `this.bar === ''` evaluates to `true` (Javascript in a conditional; `||`, will stop evaluating the conditions at the first `true` result)..  It doesn't touch the prop `this.bar` at this point.!!!

So Vue has no clue that computed prop `baz` is dependant upon both `this.foo` and `this.bar`... only `this.bar` at the moment (because that was the only prop it accessed at this point).

Then you change `bar` prop to the string `'ZZZ'`.  And then you call the computed prop `baz`.  What result do you get? you would expect `'BAZ-BAR'`, but in reality you get the cached response `BAZ`, as Vue doesn't know at this point that `baz` depends on `this.bar`, not until the first test `this.foo === ''` evaluates to false before it accesses `this.bar` for the first time (and registers it's dependency).

pi0:
@mosinve Your final approve is needed :)

alexsasharegan:
@tmorehouse thank you for the excellent explanation! It looks like the key but I was missing is getting each prop registered initially so subsequent conditionals don't make a dependent prop fall through.

tmorehouse:
@alexsasharegan yeh, it took me a while to figure out why a change to one of the props wasn't making the _items return the latest data from the provider, as the conditions to trigger the clearing computed prop cache were if-ed out on the initial run of the _items computed prop.  Wasn't until I found that article (https://skyronic.com/blog/vuejs-internals-computed-properties) that I really understood what was going on.

alexsasharegan:
Yeah, the computed props are one of the coolest features of vue. Anything cached though, ANYTHING, always gets tricky. I'm back in some wifi, so I'm gonna check out that article!

tmorehouse:
Also, with the new `head-clicked` event, you can listen on `@head-clicked` (or `@headClicked`) on `b-table` to see if the sort order has changed, So your parent app/component could just update the `items` binding (as an array), and forgo the use of provider functions too, since sorting change trigger was the key thing missing in the original b-table.

lib/components/table.vue
tmorehouse: Scoped the table classes by using `.b-table` now

tmorehouse:
@alexsasharegan I've added in your optimization suggestion for the `recToString` function

tmorehouse:
I found one small boo boo in the ARIA sort labelling.  Will fix here in a minute.

tmorehouse:
Also note that the `field.label` is rendered by default as `{{ field.label }}` in the `HEAD_` and `FOOT_` scoped slots (where as before it was `v-html="filed.label"`.  To get HTML into the header or footer heading, you now have to use the scoped slots (just like the custom rendering for data cells)

pi0:
@tmorehouse Can't we use v-html? Some small tags may be easier to inject using props instead of scopes. For example bolding text

tmorehouse:
@pi0 Yeah, we would need to wrap the text in a `<span>` in order to get the `v-html` working:

```html
<slot :name="HEAD_'+key" ...>
  <span v-html="field.label"></span>
</slot>
```
Since I don't believe we can put a `v-html` on a `slot` tag.

tmorehouse:
@pio I choose a `div` over a `span` in the latest edit in case someone wants to use a block style element inside (`spans` aren't supposed to have block level elements inside them).

tmorehouse:
I think I might add a bit of documentation around using the events listeners for `row-clicked` and `head-clicked`, to help with questions like #468

tmorehouse:
I'm also wondering if the new `head-clicked` event should be called `sort` (or `sort-changed`)? or maybe introduce a new `sort` event that just specifies the sort direction and sort key?

tmorehouse:
I have added in a new event (`sort-changed`) which is emitted only when the sorting state has changed.

The event receives a single argument of the current table context object (same object that item provider functions receive).

alexsasharegan:
@tmorehouse, the work you're doing here is awesome! But I may leave it to you to write the test coverage for this component... :wink:

I have to say, writing test coverage has been really good for me to learn more about the mechanics of each component and of vue.

Back on topic, I appreciate the more declarative naming of `sort` or something like that vs the imperative `head-clicked`, however, that only works if that's the only use case. What other purposes require a head-click listener? A `selected` event could be used for various things. It might be fine as is.

tmorehouse:
I feel the `sort-changed`event would e more of use then the `head-clicked`, bot of which can be used to detect changes in sorting directions by parent components (which was the original intent).

I'm not really sure if the `head-clicked` is needed any more for any special parent use cases).

`head-clicked` is fired on any clicking of a `thead` or `tfoot` cell, where as the `sort-changed` is only fired if the sorting criteria changes. (clicking on a non-sortable column header of footer cell clears the current sort).

tmorehouse:
For the test suite, I'll see what I can do ;-)

alexsasharegan:
Off the top of my head, the only thing I've done is to make a table heading clickable to edit so I could let a user change field names (display names at least).

tmorehouse:
So in your case, the head-clicked` event would still be handy.  Would you need the sort information, or just the key (field) name?

alexsasharegan:
I think the head-click would suffice with the field name and it's corresponding object, while the sort information can be that context object you've been working on.

tmorehouse:
Updated the `head-clicked` event signature to return the key and field being clicked (no sorting context)

Updated the documentation to reflect the changes.

tmorehouse:
For the testing of b-table, the current demo files include other components (b-select, b-pagination).  Should the testing include these, or should I break it down into maybe a couple of tables:
- Plain basic table with no pagination, filtering or sorting (just styling),
- and a more complex table with pagination, different styles and the footer tests?

Or can i use the test suite to alter the app's data to make changes (i.e. perPage, sorting, filtering, etc?

alexsasharegan:
I've been leaning towards generating multiple examples so I can test different component features. Also, I try to only test the interop between components, but let the true component testing happen in that component's spec.

tmorehouse:
OK sounds good.  I'll create probably three demo tables to test in the examples/table files.

tmorehouse:
Well, writing the tests was an interesting experience ;-)

Currently the tests just check for the basic structure of the 3 test tables, and appropriate class names.

Next to add in tests for:
- custom formatted columns, headers, and footers
- pagination & filtering, and sorting (click event simulation)
- provider functions
- events

tmorehouse:
How do I get the tests to await for the table to update after changing the props?

I am using `setData(...)` and `nextTick()` from the helpers module, but it doesn't appear to work.

If I do the checks manually via the console they come back with the expected results.

It appears to be a timing issue I think.  Maybe using a Promise with set timeout?, and  yield/await for it to resolve?

alexsasharegan:
You can use async/await on anything that returns a Promise. Let me take a quick look and see if I can spot something.

alexsasharegan:
I'm seeing your problem now. Looking into `Vue.nextTick`. I have a workaround in mind.

alexsasharegan:
Try changing the `nextTick` function to this:

```js
export function nextTick() {
    return new Promise((resolve, reject) => {
        Vue.nextTick(resolve)
    });
}
```

tmorehouse:
So change this in the helpers file, or create my own nextTick in the table.spec.js?

alexsasharegan:
Fix the helpers file. `Vue.nextTick` doesn't return a Promise, so it cannot be awaited.

alexsasharegan:
Worked?

tmorehouse:
Nope:
```
 FAIL  __tests__/components/table.spec.js (6.38s)
  ● table › table_paginated pagination works
    expect(received).toBe(expected)
    Expected value to be (using ===):
      11
    Received:
      5
      at _callee26$ (__tests__/components/table.spec.js:485:64)
```

So basically teh same issue.  The result should be 11, but it is still getting the original value of 5

alexsasharegan:
Gonna pull it down and have a quick look. Excuse me while my node-sass takes forever to build on node 8...

alexsasharegan:
Well, nextTick resolves before the test fails, so it should be awaiting things properly at least.

tmorehouse:
Yeah, it does resolve, but before the b-table has updated it looks like. :(

tmorehouse:
Hmmmm... interesting.  If I set the per page to one less than the number of items (11), it fails, but if I set it to 2 less than the number of items (10) it works.

alexsasharegan:
I saw that! I had some super weird stuff happen to me trying to test the simple breadcrumbs component. I wonder what Jest does under the hood?

alexsasharegan:
Is `perPage` a zero indexed thing?

tmorehouse:
PerPage is a 1 based index (as is currentPage).

I know if I try to go to a page that is higher than the max number of pages, pagination component goes to page 1.

Setting pagination on a table component (not tied to a pagination component), will show an empty table (and the empty result message).

I might try "disconnecting" the pagination component in the demo.html file and see if the glitch persists.

alexsasharegan:
BTW, I see a lot of this in watcher functions: 
```js
[prop](newVal, oldVal) {
    if (oldVal !== newVal) {
        // ...
    }
},
```
I was under the impression that the watcher is *__only__* called when the value does change.

tmorehouse:
Same here, but looking at a lot of other bootstrap-vue components, I saw similar tests to prevent duplicates from happening:

```js
prop(newVal, oldVal) {
    if (newVal === oldVal) {
        return;
    }
}
````

**Comment by alexsasharegan** [comment]
Well, it certainly can't hurt. If this length thing wasn't bothering me so much, I would dive into Vue source to see if a new value is guaranteed.

**Comment by alexsasharegan** [comment]
https://github.com/vuejs/vue/issues/1014#issuecomment-119044712

Looks like it depends!

**Comment by tmorehouse** [comment]
I just did a console test of the paging inside b-table and it is working fine., so I'm thinking it might be a glitch in b-pagination (which is bound to the the same data as b-table per-page in the demo).

May have to dive into b-pagination to see what is going on.

**Comment by tmorehouse** [comment]
Duh.... it is neither component having a glitch.  It is because the select is also bound to perPage, and it has finite values of 5, 10, and 15. So it was always going to perPage == 5 when the perPage was not in the list of values.

**Comment by alexsasharegan** [comment]
Hahahaha! Sometimes you just have to laugh about it.  Welp, I suppose a test for that functionality is a good idea!

**Comment by tmorehouse** [comment]
I suppose I could add in an "All" into the select (which would be bound to items.length) ;-)

**Comment by alexsasharegan** [comment]
I'm curious now... Was nextTick actually broken? I'm confident about async/await & promises, but I always like to double check my assumptions.

**Comment by tmorehouse** [comment]
I'm not sure.  from what I googled, awate will convert a value to a resolved promise if it isn't a promise.  But nextTick is callback based (gets added to the end of the processing queue, and nextTick usually returns immediately.

**Comment by tmorehouse** [comment]
An old tweet, https://twitter.com/vuejs/status/673953024449060864, but basically says that Vue doesnt assume teh availability of promises.

**Comment by alexsasharegan** [comment]
Yeah, I think it's safest to keep it as is. On a side note, in theory, you shouldn't have to call the next tick function after using setData, since it internally awaits the next tick.

**Comment by tmorehouse** [comment]
Yeah, I threw that in as a just in case I needed two Ticks :p

**Comment by alexsasharegan** [comment]
When in doubt, double down!

**Comment by tmorehouse** [comment]
The async testing is going to be interesting, since nextTick wont help at all, as I would be testing it based on the firing of the 'table::refreshed' event.

**Comment by alexsasharegan** [comment]
Post back your challenges! I love async stuff, so I'm sure I can come up with something. But I'm signing off for now.

**Comment by tmorehouse** [comment]
I might use a sleep promise (for say 0.5 second), and have the test provider(s) sleep (for 0.25s), which should be enough time to wait and catch the `table::refreshed` event, or create a mock jest function to act as the provider.

**Comment by alexsasharegan** [comment]
I was thinking of a function that takes a VM and an event and returns a promise that resolves on the callback of vm.$once(event, resolve). Then you can initiate your Ajax and await that function

**Comment by tmorehouse** [comment]
Hmm that is an idea.  Great suggestion!

**Comment by alexsasharegan** [comment]
Here's more or less my thought:

```js
export default function awaitEventOnce(event, vm) {
    return new Promise((resolve, reject) => {
        // proxy all parameters in array
        // because promises are supposed to be resolved
        // with only one argument
        vm.$once(event, (...args) => {
            // convenience tweak to resolve single arg events
            // without the extra array access overhead
            const $return = args.length > 1 ? args : args[0]
            resolve($return)
        })
    })
}
```

alexsasharegan:
You might have to also throw in a nextTick because your event may fire, but the Vue job queue is still buffering.

tmorehouse:
I am kind of thinking now, that a user can get events about sort changing, etc, that maybe item provider functions aren't really needed when it can all be controlled outside the component (in the app) now. :P 

Users can listen for 'sort-changed` event, they can control the perPage and currentPage, and filter props, although the addition of handling the provider function isn't a lot of code though, hehe (except maybe the testing part)

tmorehouse:
I think I might throws in  test on the callback function to make sure that the value is an array, and nothing else, and if it isn't an array, issue a console.warn or error.

tmorehouse:
I'll be getting back into the swing of things WRT more tests for table this week.

tmorehouse:
@alexsasharegan Just thinking about the provider function testing.

Should we implement a setTimeout or similar to handle cases where the callback/promize/event is never fired?  Otherwise there might be (rare) situations where it hangs waiting for something to happen?

alexsasharegan:
Hmm... ajax calls should have a timeout. And for testing, usually ajax is mocked. I haven't written a mock since I had to do some java though, so I forget pretty much all that! But it definitely seems valid to try and simulate the rough conditions around fetching. I don't think we should handle errors for people, but it might be nice to `console.warn` something helpful to the developer.

Am I tracking with your question?

tmorehouse:
I was just thinking more about the testing... Maybe addinging a timeout to the `awaitEventOnce` function, just in case the test fails and never fires the event and hangs the testing.

tmorehouse:
Thinking now that for the callback and Promise methods of the provider function, I just wont use any delay (i.e. no set timeouts, or a setTimeout of 0) which should negate the need for an await function for the event to fire, and then use a simple promise/await to check that the "table::refresh" event fired.

tmorehouse:
Finally, provider function testing is complete.

tmorehouse:
@pi0 @alexsasharegan 
I was just thinking of another slight enhancement to `b-table`.

BS-4 supports variants on individual cells, but currently we only support variants on the entire row (via `_rowVariant` in the data array).  I think we could add support for cell variants either by providing an array of variants to `_rowVariant,` or adding a new modifier property of `_cellVariants` which would need to be an array of variants.

What do you guys think?  This would address issue #479

tmorehouse:
Added support for `row-hovered` event (issue #514)
Added contextual variant support for individual data cells (addresses issue #479)

Issue #392, opened 2017-05-14
<b-card> add class only for header or footer


MfeaR:
I have add class `bg-info` only for header or footer. 
I need create this block
```
<div class="card">
    <div class="card-header bg-info text-white">text</div>
     hello world
</div>
```
I tired

```
<b-card show-header>
   <div class="bg-info" slot="header" variant="info">test</div>
   hello world
 </b-card>
```

But this code transform in

```
<div class="card">
     <div class="card-header">
          <div class="bg-info text-white">text</div> 
    </div>
    hello world
</div>
```

How to do it?

tmorehouse:
Try this:

```html
<b-card show-header>
  <template slot="header">
    <div class="bg-info">test</div>
  </template>
      hello world
</b-card>
````

It might not be exactly the result you want.

losblancos113:
I have same problem as you. Have you found the solution for this issue?
@MfeaR

alexsasharegan:
Can anyone confirm they tried @tmorehouse's solution and it didn't work? @losblancos113 @MfeaR

losblancos113:
@alexsasharegan ,
yes, I tried and it did't work.

@tmorehouse 
I want to add class bg-info into div.card-header but your solution still effect .card-header childs only

tmorehouse:
We may need to add in a few extra props for adding classes to various portions of the card

alexsasharegan:
+ 1 on this enhancement. I've wanted to have a simpler way to set header and footer backgrounds. I did it with CSS before, but BS4 saves a lot of unnecessary CSS and handles inverting the text color for you.

losblancos113:
Hi guys,
I have created a new props call headerClass to <b-card> component and now it work.
You can check this out my lastest commit on my forked repository.
Code:
`            headerClass: {
                type: String,
                default: null
            },
`

`        computed: {
            headerClasses() {
                let header = {
                    'card-header' : true
                };
                if(this.headerClass != null && this.headerClass.length > 0){
                    let formattedHeaderClass = this.headerClass.replace(/\s+/g,' ').trim();
                    let classes = formattedHeaderClass.split(' ');
                    classes.forEach((clazz) => {
                        header[clazz] = true;
                    });
                }
              return header;
            },
`

alexsasharegan:
[...]

Issue #338, opened 2017-05-04
<b-card>: difficult to use header slot


alexsasharegan:
I'm using the card component to create a header with a title and a button, so I want to use the header slot rather that inline too much html in the header prop attribute.

I'm finding the setup to be a little messy. Ideally, I would just be able to use the slot to add more complex markup and be done. I don't mind passing a boolean in for the header to register though, but requiring a string makes it a little awkward. I believe Vue lets you union prop types, so perhaps you could switch this up to `[String, Boolean]`.

I'll include a fiddle to look at. I don't believe you can see the prop validation warnings in the fiddle, but while I'm developing in strict mode, they are quite glaring 😃 

[jsFiddle](https://jsfiddle.net/yu07om1d/5/)

tmorehouse:
Maybe this needs some `v-if` trickery in the `card` component on the header and slot

tmorehouse:
Or... We can test for `$slots['header']` truthiness and use that to determine if a slot has been defined by the user.

I did this trick in the `b-table` enhancements for the scoped footer slots (and if not defined, tries using the scoped header slot instead).

tmorehouse:
Actually, looking ath te `card` component template:

```js
        <!-- Card Header-->
        <component v-if="header || showHeader" :is="headerTag" class="card-header">
            <slot name="header">
                <div v-html="header"></div>
            </slot>
        </component>
```

you could just set `show-header` on the card, and then provide your `header` slot template.

tmorehouse:
Or it could be simplified (and no need for the `showHeader` prop:

```html
<component v-if="header || Boolean($slots['header'])" :is="headerTag" class="card-header">
    <slot name="header">
        <div v-html="header"></div>
    </slot>
</component>
```

And if the user has defined a `header` slot, (but left `header` prop empty/null) then it will show it.

tmorehouse:
I could add this to PR #463

tmorehouse:
https://jsfiddle.net/yu07om1d/6/

tmorehouse:
We could also do something similar to the card footer slot:

```html
<component v-if="footer || $slots['footer']"
           :is="footerTag"
           :class="['card-footer', footerVariant?('bg-'+footerVariant):'', footerClass]"
>
    <slot name="footer">
        <div v-html="footer"></div>
    </slot>
</component>
```

And not have a need for the prop `show-footer`.

alexsasharegan:
[...]

PR #463, opened 2017-05-28
[card] header and footer variants & classes

tmorehouse:
Allow contextual background variants on headers and footer, and allow user to apply arbitrary classes to headers and footer.

Addresses issue #392


tmorehouse:
I could also add in a fix for #338 (see https://github.com/bootstrap-vue/bootstrap-vue/issues/338#issuecomment-304494988)

PR #466, opened 2017-05-28
Tests/Breadcrumb

alexsasharegan:
Adding tests for breadcrumb component.

This fixes some of the logic for breadcrumb items' active state (don't believe anyone has found some little bugs there)

This also makes a change to the link component for emitting click events with a native event object and propagating the click.


lib/mixins/link.js
```diff
@@ -92,10 +92,10 @@ export default {
         linkClick(e) {
             if (!this.disabled) {
                 this.$root.$emit('clicked::link', this);
-                this.$emit('click');
+                this.$emit('click', e);
             }
 
-            if (this.disabled || (!this.isRouterLink && this._href === '#')) {
```
pi0: This was to prevent scroll to top when clicking '#' links ( dropdown items with @click handler for example)

alexsasharegan: So maybe I can catch that and prevent default, but guard the stop propagation?

pi0: You are right 👍  We may just prevent default

Issue #467, opened 2017-05-28
Changing popover background color on tooltip


LiamDotPro:
I'm currently trying to figure out how to use the popover-style object correctly and the documentation doesn't really provide what I'm supposed to be passing to change styling. From the popover component I have gathered it finds a style:

```
<template>
    <div>
        <span ref="trigger"><slot></slot></span>

        <div tabindex="-1"
             class="popover fade"
             :class="[classState ? 'show' : '', popoverAlignment]"
             ref="popover"
             @focus="$emit('focus')"
             @blur="$emit('blur')"
             :style="popoverStyle"
        >
            <div class="popover-arrow"></div>
            <h3 class="popover-title" v-if="title" v-html="title"></h3>
            <div class="popover-content">
                <div class="popover-content-wrapper">
                    <slot name="content"><span v-html="content"></span></slot>
                </div>
            </div>
        </div>

    </div>
</template>
```

So I have tried adding the `popover-style` attribute with an object containing a background property as below:

```
<template>
  <div class="col-lg-2">
    <div class="tech-block">
      <b-tooltip :content="name" popover-style="{background: #ffffff;}" placement="bottom">
        <img ref="image" class="img-fluid" :src="src">
      </b-tooltip>
    </div>
  </div>
</template>
```

I'm not really sure how I'm supposed to be changing the background of this popover really at all as the documentation isn't clear of what it is expecting and can do.

tmorehouse:
Try this:

```html
 <b-tooltip :content="name" :popover-style="{background: '#ffffff'}" placement="bottom">
    <img ref="image" class="img-fluid" :src="src">
 </b-tooltip>
```

Be sure to include `:` before the property name (i.e. `:popover-style="..."`) when passing the configuration object.  Without the  `:` (shortcut for `v-bind:`) you are passing a literal string to the component, rather than  JavaScript object.

LiamDotPro:
Updated my own copy and while the tooltip works perfectly still the background still won't change here's what I tried to be sure,

```
 <div class="tech-block">
      <b-tooltip :content="name" :popover-style="{background: '#ffffff'}" placement="bottom">
        <img ref="image" class="img-fluid" :src="src">
      </b-tooltip>
    </div>
```
Here's a gif of the popover I'm trying to change incase I've note used the correct terminology, The black background on the tooltip is what I'm trying to change which I gathered from the documentation was the popover-style.

[...]

PR #470, opened 2017-05-29
[popover] [tooltip] mixin

tmorehouse:
Moved common functionality into `popover.js` mixin.

This will help auto documentation so that only applicable props are shown in each component's docs.

Currently `b-tooltip` docs list a lot of props that are only applicable to `b-popover`, so this will help prevent some usage confusion such as issue #467 (https://github.com/bootstrap-vue/bootstrap-vue/issues/467#issuecomment-304533075)


tmorehouse:
I know tooltips and popovers will be changing when tether is removed, but figured this will help some with the current usage.

SirLamer:
I agree with your approach here, I have intended to do the same during my forthcoming re-write. So in concept I have no concerns however I have not tested yet.

tmorehouse:
Due to @alexsasharegan great updates to popover, I'm going to trash this PR and create a new one that incorporates his changes.

Issue #484, opened 2017-06-05
Tooltip initial show: no animation


alexsasharegan:
Is there some kind of prop for animation on tooltips that I'm missing? The tooltip flies in from the top-left corner on both my app (see below) and on the docs demo page (just try toggling in and out of the result view). The funny thing is that it only does that initially.

I know everybody is on hold for BS4 & popover.js, but wanted to see if there are any current solutions for getting the tooltip to behave more closely to BS4 (on initial show, the tooltip appears in place rather than being moved into place).

![b-tooltip](https://cloud.githubusercontent.com/assets/16269252/26794891/53d21bfe-49d8-11e7-945e-5eaf2ae376d1.gif)

alexsasharegan:
Looks like Tether uses an absolute position of `0,0` with transforms to place elements. Is there a way to init the tooltip so calculations can set the transform before mouseover?

alexsasharegan:
It looks like Tether doesn't wrap the element until show, and then calculates position. This could be done at the mounted hook (possible perf hit though?).

I also see a lot of underscore props added after vue creation time. Is this because Vue appears to strip references to undercored props? They are still available in `this.$data['_' + prop]` if that's the issue.

tmorehouse:
See also PR #470, which is re-organizing how popover and tooltip inherit base tether stuff.  Maybe this could be mixed in with that?

alexsasharegan:
I'll try and check that out and maybe PR your PR?

alexsasharegan:
Inception PR

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #485, opened 2017-06-05
[popover] Fix potential private props collision, and init Tether on mounted [Issue #484]

alexsasharegan:
Fix for issue #484 .

As mentioned in the above issue, assigning data props to a vue instance using private notation _(underscore prefix)_ is prohibited by Vue to avoid potential collisions with Vue mechanics. All data props assigned after instantiation are now part of the initial data object and are referenced properly with `this.$data[prop]`.

Tether is now instantiated in the mounted hook to allow transform calculations to be run. This avoids the _"fly-in"_ effect previously experienced due to the popover/tooltip being transformed from `0,0` to its final position.


tmorehouse:
@alexsasharegan 

what I can do with this is to merge your changes first, and then re-jig my tooltip/popover mixin to use your code.

alexsasharegan:
That sounds fine to me! I feel good about the changes with regards to private pros here, but I'm not sure initializing the tether element in the mounted hook fixes the weird fly-in effect.

tmorehouse:
I'm not completely sure how tether works, but what you have done makes sense, as once it is mounted it should be able to determine the location of the main element.

alexsasharegan:
Exactly. Now :fingers_crossed:

tmorehouse:
Merged :)

tmorehouse:
I've created  new PR #527 that breaks out the common parts to a mixin. Check it out when you get a chance.

PR #497, opened 2017-06-11
Fixes issue #496. Emit change event for form-radio

### LINKED ISSUE #496: How to bind to Form Radio change?
I can't figure out what's the proper way to declare a function that is triggered when user changes state of radio buttons.

This seems to work, but not really as I want: https://jsfiddle.net/ozhmj480/2/
`@input` triggers my callback on the initial set up too. I want it to be triggered only on a user action.
Is there a way to do that?

BTW, it's pretty straightforward how to do that for inputs and checkboxes. I would simply use `@change`, but it doesn't work with radio buttons.

Thanks.

gornostal:
Fixes #496

### DIFF
--- a/lib/components/form-radio.vue
+++ b/lib/components/form-radio.vue
@@ -1,20 +1,20 @@
 <template>
-    <div :class="[inputClass,this.stacked?'custom-controls-stacked':'']"
-         :id="id || null"
+    <div :id="id || null"
+         :class="[inputClass, this.stacked?'custom-controls-stacked':'']"
          role="radiogroup"
     >
         <label :class="[checkboxClass,custom?'custom-radio':null]" v-for="option in formOptions">
-            <input v-model="localValue"
+            <input :id="option.id || null"
                    :class="custom?'custom-control-input':null"
+                   ref="inputs"
                    type="radio"
+                   v-model="localValue"
                    :value="option.value"
                    :name="option.name"
-                   :id="option.id"
                    :disabled="option.disabled"
-                   @click="onClick"
-                   ref="inputs"
+                   @change="$emit('change', returnObject ? option : option.value)"
             >
-            <span class="custom-control-indicator" aria-hidden="true" v-if="custom"></span>
+            <span v-if="custom" class="custom-control-indicator" aria-hidden="true"></span>
             <span :class="custom?'custom-control-description':null" v-html="option.text"></span>
         </label>
     </div>
@@ -51,14 +51,6 @@
             returnObject: {
                 type: Boolean,
                 default: false
-            }
-        },
-        methods: {
-            onClick(e) {
-                if (this.value !== e.target.value) {
-                    // run on the next tick to have this.value updated at a time of 'change' event
-                    this.$nextTick(() => this.$emit('change', e.target.value))
-                }
             }
         }
     };
--- a/lib/components/form-radio.vue
+++ b/lib/components/form-radio.vue
@@ -11,6 +11,7 @@
                    :name="option.name"
                    :id="option.id"
                    :disabled="option.disabled"
+                   @click="onClick"
                    ref="inputs"
             >
             <span class="custom-control-indicator" aria-hidden="true" v-if="custom"></span>
@@ -50,6 +51,14 @@
             returnObject: {
                 type: Boolean,
                 default: false
+            }
+        },
+        methods: {
+            onClick(e) {
+                if (this.value !== e.target.value) {
+                    // run on the next tick to have this.value updated at a time of 'change' event
+                    this.$nextTick(() => this.$emit('change', e.target.value))
+                }
             }
         }
     };


tmorehouse:
See comments https://github.com/bootstrap-vue/bootstrap-vue/issues/496#issuecomment-307611701

gornostal:
I'm using `@click.capture` so `this.value` is updated at a time of `change` event.
As an alternative we could emit event on `vm.$nextTick`.
Maybe there is a better solution, but I don't see it now.

gornostal:
Please disregard that comment. Not ready yet

gornostal:
Done. Not sure if this is the best solution though.

alexsasharegan:
My questions: 
1. Is the nextTick call there to keep the change event from firing before the user sees the radio button's state change? This seems inconsistent with the way other inputs fire change events.
2. This works on clicks, but what about keyboard users who use the spacebar? What other ways can the value be triggered? 
3. Radios are usually used in a collection of options. I want to play with this when I get back on desktop and see how it all works. Just looking at this from the phone right now.

gornostal:
1. Yes
2. I tried selecting different radio buttons using a keyboard and this solution seems to work fine
3. I didn't try that

nextTick looks like a workaround to me, and I hope you guys can find a better solution.
Maybe because I'm new to Vuejs I don't see how to implement this improvement better.

tmorehouse:
@alexsasharegan Radio buttons will trigger a click when the keyboard is used to select them (i.e. cursor left/right/up/down), so no need to monitor keydown events :-)

Yeah, the value can possibly be an object, rather than just a single scalar value, so you might need to test this to see if the comparison works (i.e. comparing references or values).

alexsasharegan:
@tmorehouse, are you sure about this cross-browser change problem? I spun up the playground and ran the stock vue-docs example, but added a change event listener. Everything seems to be working 👌 

### Playground template
```vue
<input type="radio" id="one" :value="One" v-model="picked" @change="log">
<label for="one">One</label>
<br>
<input type="radio" id="two" :value="Two" v-model="picked" @change="log">
<label for="two">Two</label>
<br>
<span>Picked: {{ picked }}</span>
```

### Playground JS
```js

data: {
    picked: null,
    One: {one: 'one'},
    Two: {two: 'two'},
},

methods: {
  log(){
    console.log(...arguments)
  }
}
```

Also, I'm thinking we can avoid proxy emitting a lot of input events (and elsewhere) by providing examples of using `@event.native="handler"`. It cuts through the component events and listens on the html element so we're back to plain vue code again (barring a stopPropagation or something like that).

So I tried switching the `v-model` on the radio component's internal input to the raw `value` prop, and just let Vue do all the work. Works!

### Playground template
```vue
<b-form-radio type="radio" id="one" :options="options" v-model="picked" @change.native="log" />
<label for="one">Test</label>
<br>
<br>
<span>Picked: {{ picked }}</span>
```

### Playground JS
```js

data: {
    picked: null,
    options: [
      {
        text: 'One',
      	value: {one: 'one'}
      },
      {
        text:'Two',
        value: {two: 'two'}
      }
    ],
},

methods: {
  log(){
    console.log(...arguments)
  }
}

```

tmorehouse:
Hmmm, maybe most browsers have fixed this.  IE Used to fire change on both the clicked radio and the one that became unchecked, causing two change events to happen.

tmorehouse:
So if our current supported browser base **does** handle the change events correctly, then it could be as simple as add the following to the radio input(s):

```html
<input type="radio" @change="$emit('change', value)" ...>
```

alexsasharegan:
Then there's no need for localValue, right?

tmorehouse:
I think localValue handles the difference between the computed value and the value on the radio (i.e. value can be an object, where as in regular HTML the value can only be a string).

I'll have to go over the code to see what it is really doing.

alexsasharegan:
You can listen to a `change` event, but when the event fires, you can just reference your `v-model` bound prop. Naturally, the value has to be coerced to a string in the html, but I suppose it would be nicest if the event was called with the un-coerced value.

tmorehouse:
After a bit of googling, it appears that bootstrap-vue's supported browser base handles the `change` event correctly.  So I think we can revert back to `@change(...)` on each radio.  We just want to make sure we provide the "massaged" radio value (i.e. if the value is an object, not a string).

tmorehouse:
@gornostal if you don't mind, I'll tweak your PR with our latest findings.

gornostal:
I don't.
You can go ahead and do whatever you want with this PR)

tmorehouse:
I'll need to update the docs meta.json once this is merged to document the change event.  Check out the latest changes.

tmorehouse:
I might have been wrong on my assumption about value being an object, as the `returnObject` prop doesn't refer to the value being an object, but rather returning the option option (which contains a property `value`, although technically nothing would stop someone from passing an object as the `value` inside the options object/array.

alexsasharegan:
Sorry, I wasn't speaking to that part. I was looking at this section of `mixins/form-options.js`:
image

tmorehouse:
Yeah, it depends on how the `value` is updated when it is an object (i.e. passing a new object reference v.s.  changing a property within the object)

tmorehouse:
It appears that in this case, `localValue` gets updated by re-assigning the base object reference, so we should be OK.

tmorehouse:
I think we should be good to merge this in.

alexsasharegan:
Let's do it!

tmorehouse:
Merged, and I have update `form-radio` docs meta.json to reflect new event.

Issue #516, opened 2017-06-16
[list-group-item] "to" prop should accept [String, Object] as of mixins/link ?


elb:
Hello, i would expect to be able to pass an object too but i may miss something.

```
<b-list-group flush>
  <b-list-group-item tag="b-link"  v-for="(method, i) in service.methods" :key="i" :to="{name: 'method', params: {id: method.id }}">
    ...
  </b-list-group-item>
</b-list-group>
```

```
[Vue warn]: Invalid prop: type check failed for prop "to". Expected String, got Object.
```

* [components/list-group-item.vue](/bootstrap-vue/bootstrap-vue/blob/master/lib/components/list-group-item.vue#L62)
* [mixins/link.js](/bootstrap-vue/bootstrap-vue/blob/master/lib/mixins/link.js#L61)

Should i open a PR ?

Thanks for your work.

tmorehouse:
Sure, open a PR with the change.

Thanks!

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #517, opened 2017-06-16
<b-list-group-item> "to" prop should accept [String, Object]

elb:
https://github.com/bootstrap-vue/bootstrap-vue/issues/516

### DIFF
--- a/lib/components/list-group-item.vue
+++ b/lib/components/list-group-item.vue
@@ -59,7 +59,7 @@
                 default: null
             },
             to: {
-                type: String,
+                type: [String, Object],
                 default: null
             },
             href: {


alexsasharegan:
I feel like we can further abstract some of these link props to be more centralized and reuseable. Thoughts, @tmorehouse ? You recently added the `rel` attribute support. You probably went around hunting for all the components that proxy those link attrs to add that in. Wouldn't it save some Kb and sanity if we pull those into a mixin? Then we can spread those props out in the link mixin itself. Something like `baseLinkProps`.

tmorehouse:
@alexsasharegan Yeah, that might be a good option... link-extras.js mixing maybe?

alexsasharegan:
That sounds good! I feel like that could be a separate PR. This one is a slam dunk, and who knows who will get to this mixin extras idea first.

pi0:
Why not using link mixin?

tmorehouse:
@pi0 `b-list-group-item` currently doesn't use link mixin (as I think because `b-list-group-item` supports more content types than just links).  But it might be an option to move them there.

alexsasharegan:
@pi0 Link mixin contains a lot of extra props that can't be dumped blindly to other components that proxy props down to `bLink`. Currently, you would have to pluck the desired props off the link mixin by name. That's still an improvement as seen with this little bug, but not as much bang for your buck.

tmorehouse:
@alexsasharegan True, when importing the mixin, it could be selective for plucking, although the merge strategy would have to me manual

alexsasharegan:
Although, who knows, maybe all the props can be stuffed into components like this. Ultimately, I was hoping to do this:

```js
import { props as linkProps } from "../mixins/link-extras"

{
  props: {
    ...linkProps,
    // component props
  }
}
```

tmorehouse:
That might be an option for props... but computed props, and other stuff that depend on `this` from other parts of  the mixin would be more difficult to handle.

alexsasharegan:
https://github.com/alexsasharegan/bootstrap-vue/tree/issue/516

Issue #513, opened 2017-06-15
Is it possible my make an accordion with bootstrap-vue?


FragsterAt:
https://v4-alpha.getbootstrap.com/components/collapse/#accordion-example

tmorehouse:
It is possible.  You would need to have a unique visible/show state for each collapsible component, and listen/watch for changes to the values and close any other open collapse components.

tmorehouse:
@FragsterAt I have crated PR #519 which adds accordion support to `collapse`.

tmorehouse:
@FragsterAt I have crated PR #519 which adds accordion support to collapse.

PR #519, opened 2017-06-17
[collapse] Add Accordion support, ARIA, and test suite

tmorehouse:
Adds Accordion support to `b-collapse`.

Collapses can be turned into an accordion by setting the prop `accordion` on the collapse components to a common group identifier name. Uses event `accordion::toggle` emitted on $root to signal other collapses in the group to close

Adds ARIA compliance support to directive `v-b-toggle`

Addresses issue #513 

Includes updated documentation and a test suite

Working example: https://jsfiddle.net/ya15m7dv/


kylemh:
Does this implementation of an Accordion support nested items? Can I create an Accordion within `<b-card>` components? I tried doing so, but am failing miserably without documentation.

mosinve:
Is this fiddle covers your question? 
https://jsfiddle.net/cihkem/n5rgjvfj/3/

kylemh:
You're the man!

mosinve:
Second example was not saved... Now all ok.

Issue #490, opened 2017-06-07
breadcrumb error on Vue/ SSR 


hvalcourtSerdy:
Hi,

I use Vue, with Server Side Rendering and Bootstrap-vue 0.14.0.

When I reproduce the exemple here on your website https://bootstrap-vue.github.io/docs/components/breadcrumb , I get a warning from SSR with b-breadcrumb :

```[Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside <p>, or missing <tbody>. Bailing hydration and performing full client-side render.```

Maybe the problem is this one, the code look like this :

```
<section id="breadcrumb" class="row no-gutters align-items-center">
    <div class="container extra-padding">
        <div class="row">
            <div class="col">
                <ol class="breadcrumb">
                    <li class="breadcrumb-item"><a **active-class="active"** href="#">Admin</a></li>
                    <li class="breadcrumb-item"><a **active-class="active"** href="#">Manage</a></li>
                    <li class="breadcrumb-item active"><span>Library</span></li>
                </ol>
            </div>
        </div>
    </div>
</section>
```

Like you see, the attribute on <a> is **active-class="active"**.

The  **active-class="active"** is not a valid attribute, maybe the problem is there ?
`Error: Attribute active-class not allowed on element a at this point.`

Is your code is compatible with SSR ? 

Thanks for the help,

Hugo

jmharveys:
I've got the same problem @hvalcourtSerdy !

tmorehouse:
`active-class="active"` is used by Vue-router `<router-link>`

Are you using vue router?

tmorehouse:
Does the issue still appr in the current release https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/v0.16.1 ? Some changes have occurred with the b-link component since 0.14.0

tmorehouse:
@hvalcourtSerdy @jmharveys 

The `active-class` attribute should only appear on `b-link` when vue-router is installed (and you are using the `to` prop). Can you confirm if you are using vue-router?

hvalcourtSerdy:
Yes we are using vue-router, and we are on "bootstrap-vue": "^0.14.0". I will make a test on 0.16.1, and edit this post with the result

tmorehouse:
OK. we have to check with the new release that is coming out.

alexsasharegan:
[...]

Issue #473, opened 2017-05-29
v-on on nav-item not working


lukaszgoworko:
Hi

I have this Code
```
<template>
  <b-navbar class="my-navbar" toggle-breakpoint="md" toggleable fixed="top">

      <b-nav-toggle target="nav_collapse"></b-nav-toggle>

      <b-link class="navbar-brand" to="/">
        <div class="nav-logo-frame">
          <img class="logo" alt="my_brand" src="../assets/Logo.svg" type="image/svg+xml">
        </div>
      </b-link>

      <b-collapse is-nav id="nav_collapse">
        <b-nav is-nav-bar class="ml-auto">
          <b-nav-item to="contact#ScrollView_Contact" v-on:click="clickedNavItem()">Contact</b-nav-item>
          <b-nav-item to="career#ScrollView_Career" v-on:click="clickedNavItem()">Home</b-nav-item>
        </b-nav>
      </b-collapse>
    </b-navbar>
</template>

<script>
export default {
  name: 'NavigationBar',
  methods: {
    clickedNavItem() {
      console.log('toggle navigation');
    },
  },
};
</script>
```

Nothing happens. No log; no Event (Vue-Devtools in Chrome)... nothing. Do I doing somthing wrong???

tmorehouse:
This is a current issue with router-links not propagating the @click event.

use `@click.native="cickedNavItem()"`

lukaszgoworko:
thx. working

Issue #371, opened 2017-05-09
[dropdown-item] Render as button if neither HREF or TO provided, or add option to render as button?


tmorehouse:
BS V4.alpha.6 `.dropdown-items` support both using links or buttons (although there are slight styling issues between the two font-wise, which will be fixed in beta release https://github.com/twbs/bootstrap/issues/22560).
http://v4-alpha.getbootstrap.com/components/dropdowns/#menu-items

Maybe the dropdown-item component should create a `<button class="dropdown-item"></button>` when both `href` and `to` are falsy? rather than using a link tag?.

tmorehouse:
Maybe `b-dropdown-item` should use  `b-button` component?

MarZab:
Looking at bootstrap documentation, buttons are the way to go. They look identical to the normal links so I think we should cohere to that https://v4-alpha.getbootstrap.com/components/dropdowns/#menu-items

tmorehouse:
Although, there are some cases where people would like to use dropdown-items as navigation links as well.  So I think the choice of button or link should be made on the presence of `href` and/or `to`

MarZab:
Here is a fiddle with the options:
https://jsfiddle.net/7htvb7gh/3/

Looks like buttons do add a `outline` while `a role=button` do not. On the other hand button text is unselectable while the anchor tag text is.

It might be best going with a link tho, since it would be less code to maintain and it gives a more consitent result.

tmorehouse:
True.  Although dropdown menu text shouldn't be "selectable".  There are a few tweaks to the buttons in V4 Bootstrap coming soon (see https://github.com/twbs/bootstrap/issues/22560)

Right now the link mixin defaults href to `#` (or if the link is disabled).  So maybe this should be altered to allow href to be `null` or `''` (empty string)

tmorehouse:
I can do a mock-up later today to see what can be done.

tmorehouse:
@MarZab, take  look at how b-btn works: https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/button.vue

Based on the lack presence of either `to` or `href` a button is rendered.  I think this kind of model might work best for dropdown-item.

MarZab:
That makes sense since you did ask for a button to begin with.

I would argue we stick with the link as default for the dropdown for the same reason. It would look and behave exactly as the router and link variants.

tmorehouse:
[...]

PR #520, opened 2017-06-17
[🏗🛠🔗] More granular link mixin

alexsasharegan:
Per the conversation on issue #516 and it's PR #517, this proposes a more granular link mixin. This allows components that consume part of the internal link component API to automatically stay up to date with proxying all the link props down to the b-Link component itself.


tmorehouse:
@alexsasharegan I'll take a look at your work once I get back on Sunday :)

alexsasharegan:
This one has me stumped. The button is rendered with the disabled attribute perfectly. The component has `disabled === true`. The element does **not** contain the disabled attribute, thus the test fails. It does not emit a click event when it's disabled, which is correct, so behavior is still a-okay.

image
image
image
image

alexsasharegan:
The test doesn't _have to_ check the disabled property, but it bugs the hell out of me that it worked before and not now.

tmorehouse:
Odd.  Technically `<a>` elements do not support the disabled property (they wont act disabled, hence the use/need of the `@click.stop.prevent` (or by passing `$event` to the handler and using `e.preventDafault()` and `e.stopPropagation()` when the a has the disabled class/state)

alexsasharegan:
Ah, I think I got it then. Button's new link defaults are forcing it into an anchor tag!

tmorehouse:
You will probably have to override the `href` prop in `b-button` to default it to `null` or an empty string `''`

alexsasharegan:
I'll definitely have to check and see what other components pivot off of the default values like the button component.

alexsasharegan:
Looks like we don't have an object spread plugin. Probably passed on my end since I'm running Node 8.

tmorehouse:
Ah, yeah... not all ES6 features are available.  Babel can only do so much with transpiling.

You might need to use `Object.assign(...)` to merge your imported props, or create, outside of the `export default` to create the object(s) to create a local `mixin` inside the component vue file.

alexsasharegan:
Documentation for options config in Bublé is lacking... Any ideas how to get this working?
image

tmorehouse:
@alexsasharegan no clue offhand.... maybe @pi0 might know

lib/components/button.vue
```diff
@@ -15,10 +15,7 @@ const linkPropsNoDefault = {
     href: { type: linkProps.href.type },
     to: { type: linkProps.to.type }
 };
-const noConflictLinkProps = {
-    ...omitLinkProps('disabled', 'href', 'to'),
-    ...linkPropsNoDefault
-};
+const noConflictLinkProps = Object.assign(omitLinkProps('disabled', 'href', 'to'), linkPropsNoDefault);
```
tmorehouse: Would you want this instead:
```js
const noConflictLinkProps = Object.assign({} omitLinkProps('disabled', 'href', 'to'), linkPropsNoDefault);
```
As object assign tries to modify the first object passed to it.

tmorehouse: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

alexsasharegan: I think it's effectively the same. The function uses a fresh object:

```js
export function omitLinkProps(...propsToOmit) {
    return Object.keys(props).reduce((memo, prop) => {
        if (!propsToOmit.includes(prop)) {
            memo[prop] = props[prop];
        }

        return memo;
    }, {});
}
```

alexsasharegan:
I have at least one more component to fix default props on before I remove the WIP tag.

tmorehouse:
So it will become a WAP  (work after progress), hehe.

Excellent work!

tmorehouse:
Just thinking about issue #490 

Iwonder if we should make sure that `active-class="active"` is ever only added to links that have a `to` prop (i.e. `router-link`?

alexsasharegan:
I'll take a look at that. Should be a quick fix on the link template.

alexsasharegan:
> I wonder if we should make sure that active-class="active" is ever only added to links that have a to prop (i.e. router-link is the only component that uses this property)?
- This is only ever added to a `<router-link>`. Is that sufficient?

tmorehouse:
Yeah, that is correct.  It is used by the router-link to handle styling if the link becomes active on the local browser router instance.

alexsasharegan:
I'm not familiar with SSR yet, but everything looks good. I plugged in your breadcrumb ARIA fixes if we want to use this. I noticed the `aria-current` was missing the colon for the shorthand binding in the template. If we don't merge this, we'll want to fix that.

tmorehouse:
Oops, I will make  quick fix for that in the master.  Can you add the colon to your PR version of breadcrumb?

tmorehouse:
I have added the `:` binding for aria-current in breadcrumb.

alexsasharegan:
Yeah, I added it as well.

alexsasharegan:
So if #490 is running into issues using vue-router, do we have any ideas on what we can tweak? When I use active-class, it doesn't render in the html. Could this be a bug outside the BV lib?

tmorehouse:
It is possible.

alexsasharegan:
Looks like this PR fixes that issue actually: https://github.com/bootstrap-vue/bootstrap-vue/issues/490#issuecomment-309518728

lib/components/button.vue
```diff
@@ -1,87 +1,96 @@
 <template>
-    <button :class="classList"
+    <button v-bind="conditionalLinkProps"
```
tmorehouse: This is a neat trick.  I never realized one could bind multiple props/values in a single v-bind (until I just read the docs :wink: )

lib/components/button.vue
```diff
@@ -1,87 +1,96 @@
...
-            btnDisabled() {
-                return this.disabled ? 'disabled' : '';
-            }
+// Grab a fresh object of link props (omitLinkProps does this)
+// less the 'href' and 'to' props
+// that we will reconstruct without any defaults
+// so our computed 'componentType' functions properly
+const linkProps = Object.assign(omitLinkProps('href', 'to'), {
+    href: { type: originalLinkProps.href.type },
+    to: { type: originalLinkProps.to.type }
```
tmorehouse: And I like how you pulled in the original prop types, but not the defaults.

lib/components/button.vue
```diff
@@ -1,87 +1,96 @@
...
-                    this.$emit('click', e);
-                }
+
+        conditionalLinkProps() {
+            return this.componentType === 'button' ? {} : this.linkProps;
+        },
+    },
+
+    // merge our prepared link props with button props
+    props: Object.assign(linkProps, {
```
tmorehouse: Another nice trick for merging

lib/components/list-group-item.vue
```diff
@@ -1,79 +1,69 @@
...
-            myTag() {
-                if (this.tag) {
-                    return this.tag;
-                }
-                return (this.to || this.href) ? 'b-link' : 'div';
+export default {
+    components: { bLink },
+
+    computed: {
+        linkProps: computed.linkProps,
```
tmorehouse: I wonder how vue handles caching of this type of computed prop?

I'll take a peak at the source for the `computed.linkProps` to see about reactivity registration.

tmorehouse: Or maybe `computed: {...}` could be `computed: Object.assign({}, computed.linkprops, { ...})`? Although then you woul lose the nice compact `v-bind="linkProps"`

alexsasharegan:
All right, I think I've double and triple checked things here.

lib/components/list-group-item.vue
alexsasharegan: I can mock up a test in the playground and inspect the runtime values to see how smart vue is about binding the computed getter through a mixin. It's a tad on the fancy side, but that's how I find the boundaries!

lib/components/button.vue
alexsasharegan: Gotta keep it dynamic!

lib/components/button.vue
alexsasharegan: I want object spread!

tmorehouse:
From my skim through tonight, all looks pretty good.

lib/components/button.vue
alexsasharegan: I found this looking for proxying events dynamically. That eventually lead me to that `@event.native`, but this was an awesome find along the way!

alexsasharegan:
I'll probably get a chance to run that runtime test tomorrow morning.

tmorehouse:
I think the work you are doing here will potentially address #371

alexsasharegan:
The computed properties work awesome 👍 
I did have to fix a little something on the list-group-item, since it was defaulting to an anchor tag and not utilizing the b-link component.

lib/components/list-group-item.vue
alexsasharegan: Works like a charm from what I can tell!

tmorehouse:
Excellent work!

lib/components/dropdown-item.vue
```diff
@@ -1,31 +1,21 @@
 <template>
     <b-link class="dropdown-item"
             role="menuitem"
```
tmorehouse: In BS V4.alpha.6, dropdown items can also be a button: http://v4-alpha.getbootstrap.com/components/dropdowns/#menu-items

I winder if this could be conditional, and if no `to` or `href` set, then a `b-btn` is used.

```html
<b-link v-if="href || to" ...>
   <slot></slot>
</b-link>
<b-btn v-else class="dropdown-item" @click="$emit('click', $event)">
  </slot>
</b-btn>
```

alexsasharegan: That's a tough one to decide. I feel like a separate component makes more sense, but it also seems like unnecessary weight.

alexsasharegan: If they were bundled into the same component, which one would be the smartest default? I guess button, like you said, because you can detect link props to make it a b-link.

alexsasharegan: Looks like button props might be worth abstracting as well in that case.

alexsasharegan: ...and a button can be a link!

tmorehouse: Although most button props are not needed in dropdowns (i.e. contextual styles, size, etc)

Maybe use a regular 'button` element and no need to import b-button props?

tmorehouse: Although, disabled props are needed for button in  dropdown menu.

tmorehouse: I'll make a mockup in the playground or jsFiddle and try placing a regular button element within a dropdown menu (and adding the appropriate `dropdown-item` class and see how it works

alexsasharegan: Hmm... I this is def worth figuring out, but maybe a different PR.

alexsasharegan: I'm pretty confident this will solve issues, not create them, but I'd love to give it some time in the wild to confirm its stability.

tmorehouse: OK, we can separate out using a button in dropdown in a future PR.

I think the current workaround of manually adding a button in the markup will suffice for now.  A separate component might be easiest to implement.

lib/components/link.vue
```diff
@@ -13,8 +13,7 @@
        :event="event"
        :tag="tag"
        :class="linkClassObject"
-       @click="linkClick"
-    >
+       @click="linkClick">
```
tmorehouse: I wonder if this should be @click.native="linkClick($event)"`, as router link doesn't propagate a click event upwards.

See issue #534, and another few issues that I can't recall offhand)

tmorehouse: Related issues: #473 #362 https://github.com/bootstrap-vue/bootstrap-vue/issues/284#issuecomment-297158880

alexsasharegan: Hmm... good catch!

tmorehouse: This would probably fix quite a few things.  and might allow us to clean out some code in related components (i.e. nav-item`, dropdown-item`, etc) that use `@click.native` to get around this.

alexsasharegan: I'm having a hard time testing the `click.native` in the playground. The playground doesn't get the $router. But I'm willing to go on faith here 😉

lib/components/dropdown-item.vue
tmorehouse: I made a quick fiddle which places button inside a dropdown https://jsfiddle.net/q524gxh7/

So it would be an easy component to make.  It just needs to emit a click event upwards, and accept the disabled prop.

lib/components/nav-item.vue
```diff
@@ -9,12 +9,12 @@
...
 
 <script>
-import { props, computed } from '../mixins/link';
+import { props as linkProps, computed } from '../mixins/link';
 import bLink from './link.vue';
 
 export default {
     components: { bLink },
-    props,
+    props: linkProps,
```
tmorehouse: :+1:

lib/components/link.vue
```diff
@@ -13,7 +13,7 @@
        :event="event"
        :tag="tag"
        :class="linkClassObject"
-       @click="linkClick">
+       @click.native="linkClick">
```
tmorehouse: :+1:

lib/components/dropdown-item.vue
alexsasharegan: This button doesn't include a role attribute. Should it have something for that? I defer to your accessibility expertise!

tmorehouse: I;'m leaning more towards a new component `b-dropdown-item-btn` now, since removing the default `href` of `#` from `b-dropdown-item` might be breaking for some users (if they are expecting `a` links (although I don't think functionality would be affected, they would just get a button instead of a link)

alexsasharegan:
```vue
<template>
    <button class="dropdown-item"
            type="button"
            :disabled="disabled"
            @click="$emit('click', $event)">
        <slot></slot>
    </button>
</template>

<script>
export default {
    props: {
        disabled: {
            type: Boolean,
            default: false
        }
    },
};
</script>
```

tmorehouse: Yeah, basically.  And buttons respect the disabled prop, so no need to do any special handling.

alexsasharegan: And I was just calling it `dropdown-button` (with the shorthand option). Seems in keeping with the others right?

alexsasharegan: When using the button, the dropdown doesn't close. I can't seem to find the logic around that. I thought the clickOutListener might be doing it, but it explicitly avoids running when triggered from an element inside the dropdown.

tmorehouse: Yeh, I was just thinking that some people might get confused and think it is the actual dropdown toggle button though.

tmorehouse: So `dropdown-item-btn/button` might be more _clear_ top some users

alexsasharegan: Valid point

tmorehouse: And it should have a role of` menuitem`:
```html
<template>
    <button class="dropdown-item"
            type="button"
            role="menuitem"
            :disabled="disabled"
            @click="$emit('click', $event)">
        <slot></slot>
    </button>
</template>
```

tmorehouse: Although if we update the documentation for dropdown, and explain the sub-components  bit more, it might clarify their usage and we could go with `dropdown-button`

tmorehouse:
You can also add the reference to b`dropdown-button` in the dropdown docs `meta.json` file.

alexsasharegan:
Ooh, that was cool! I haven't been in there before.

lib/components/dropdown-item-button.vue
```diff
@@ -0,0 +1,26 @@
...
+export default {
+    props: {
+        disabled: {
+            type: Boolean,
+            default: false
+        }
+    },
+    methods: {
+        onClick(e) {
+            this.$parent.$el.click(e);
```
tmorehouse: Is this.$parent.$el.click(e)` really needed?   The regular `this.$emit('click',e) should propagate up.

alexsasharegan: It don't work! I commented above about that. The clickout is ignoring anything inside the dropdown's element tree. Not sure why this doesn't work, but the others do...

tmorehouse: Hmmm.  b-link calls it's click(e) event, which basically emits up via this $emit.

tmorehouse: What about copying in the b-link click event (which is in the mixin)?

alexsasharegan: Lines 31 - 32 are where the items live. There's no click listener there:
```vue
<template>
        <div :class="['dropdown-menu',{'dropdown-menu-right': right}]"
             ref="menu"
             role="menu"
             :aria-labelledby="id ? (id + (split ? '__BV_toggle_' : '__BV_button_')) : null"
             @keyup.esc="onEsc"
             @keydown.tab="onTab"
             @keydown.up="focusNext($event,true)"
             @keydown.down="focusNext($event,false)"
             @mouseover="focusHovered($event)"
        >
            <slot></slot>
        </div>
</template>
```

alexsasharegan: I can add a click listener on the default slot wrapper that calls the clickout listener.

tmorehouse: `b-link` does the following:
```js
    linkClick(e) {
        if (!this.disabled) {
            this.$root.$emit("clicked::link", this);
            this.$emit("click", e);
        } else {
            e.stopPropagation();
        }

        if (!this.isRouterLink && this._href === "#") {
            // stop scroll-to-top behavior
            e.preventDefault();
        }
```
which is basically a `this.$emit(...)`

tmorehouse: dropdown.vue is the one that listens for the @click event from child components.

alexsasharegan: Yup, stared at that for a while. But follow the trail, where does that event land? `dropdown.vue` _doesn't_ listen on it's children. It just uses the clickout mixin

tmorehouse: Although maybe `this.$parent.$el.emit(...)` might be a good temporary workaround.

alexsasharegan: Alright, I think I might have it figured out—`this.$root.$on('clicked::link', listener);`

alexsasharegan: It's listening to $root!

lib/components/list-group-item.vue
alexsasharegan: By the way, I think we might have been overthinking this. When assigning `computed.linkProps`, you're just assigning the function, not the object of computed functions. Vue does all the work to reuse that single function and call it in the correct context of the component instance.

lib/components/button.vue
tmorehouse: Once all the major browser switch to ES6, we can always revisit :smiley:

tmorehouse:
How does the auto-documentation look for the link related components?  do they now show only the imported props?

alexsasharegan:
image
image

tmorehouse:
Excellent! :grinning:

tmorehouse:
Here is a question... does dropdown-item (and others) actually respect the `tag` prop?  I tried on the current release on dropdown, setting `tag="button"` and it still renders as an `a`.  So maybe tag isn't really required on some of these components?

alexsasharegan:
My understanding is this: if you use the router and the props required to trigger b-link as a router-link, then the `tag` prop is passed to router-link.

alexsasharegan:
https://router.vuejs.org/en/api/router-link.html
image

alexsasharegan:
It's arguably a pretty edge-case feature.

tmorehouse:
Ah, OK.  that makes sense.  Maybe we should document which props are router-link only props somewhere...

tmorehouse:
I wonder if these props would be better suited to having `router-` prepended to them?  this would be a breaking change though.

alexsasharegan:
That sounds nice. Self-documenting for users and us alike!

tmorehouse:
We could have both props (prepended and non-prepended), and favour the value of hte pre-pended prop (i.e. `this.routerTag || this.tag`

And put a deprecation note in the docs, and remove the non pre-pended ones once release 1.0.0 rolls around

alexsasharegan:
That can get a little awkward depending on how you set up defaults. How about the default anchor tag moves to `routerLink`, the `tag` default is `null`, and then we use a computed prop like so:

```js
componentTag(){
    if (this.tag) {
        console.warn('<b-link> "tag" property is deprecated, please use "routerTag" property instead.');

        return this.tag
    }

    return this.routerTag
}
```

tmorehouse:
That makes sense... and when we remove the old props we can just re-jig the calls

lib/components/table.vue
```diff
@@ -132,7 +134,7 @@
                 default: () => {
                     if (this.itemsProvider) {
                         // Deprecate itemsProvider
-                        console.warn('b-table: prop items-provider has been deprecated. Pass a function to items instead');
+                        warn('b-table: prop items-provider has been deprecated. Pass a function to items instead');
```
tmorehouse: :+1:

tmorehouse:
And what is nice about this wrapper computed prop is that it will only generate the warn when the deprecated property changes

lib/components/table.vue
alexsasharegan: I just tried to copy Vue's helper messages. They sort of namespace themselves `[Vue warn]:`. We could even style them all if we wanted to be clever.

alexsasharegan:
Just curious, where are we at on this? Is this the last PR before pushing 0.17.0?

tmorehouse:
I think this might be the last PR :-)

PR #526, opened 2017-06-19
[breadcrumb] ARIA changes

tmorehouse:
Removed role="navigation" per https://github.com/bootstrap-vue/bootstrap-vue/issues/333#issuecomment-309303101

Added `aria-current` attribute on active item. Allow user to specify `aria-current` value via prop (defaults to `"location"`)

Updated unit tests to reflect changes


Decrepidos:
Great solution 😁

Issue #523, opened 2017-06-17
Pagination bad UX: n+1 page should always be visible


tristanbes:
Hello,

Some users navigate through the pagination by clicking on the number of the page instead of using the `>` and `<`. Currently, it feels weird because page number 2 is never shown on the pagination. It force you to position your mouse to `>` and then to go back with your mouse on item `3`

<img width="237" alt="metadata_title_account_list_cv" src="https://user-images.githubusercontent.com/346010/27254668-724a9afc-538d-11e7-8285-a47ae8fcbe71.png">

And as soon as i'm on page 2, the page 3 button appears, which is the right & by default behaviour we should have (even on page 1)

<img width="310" alt="metadata_title_account_list_cv" src="https://user-images.githubusercontent.com/346010/27254671-9bcb586c-538d-11e7-887c-6a0d084be0b7.png">

Can you consider always showing `n+1` page as a button please ? (or provide an option to allow us to do this if you don't want to change your default behaviour).

Thank you :-)

tmorehouse:
Yeah, pagination does have some strangeness in how it renders the buttons.  We're definitely getting ready to work on this issue.  Improvements might be available in release 0.17.0

tmorehouse:
There are already a few enhancements available in the current master, but most do not apply to the issue you bring up.

tmorehouse:
PR #528 should address some of the UX experience.

tmorehouse:
PR #528 now addresses N-1 & N+1 buttons always being present (where applicable)

Issue #508, opened 2017-06-15
Pagination - Ellipsis while no pages in between


b-zee:
In the following scenario, ellipsis are shown between the page links of 1 and 2:
![image](https://user-images.githubusercontent.com/6576760/27177040-585a2456-51c5-11e7-9ef1-20a5ae59bbd1.png)

There should be no ellipsis, as it makes one think there is a page in between.

tmorehouse:
Yeah, there is a lot of weirdness in how the pagination buttons are rendered.  We're hoping to work on this before release 0.17.0 comes out.

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

Issue #506, opened 2017-06-15
Pagination - first and last page do not show next and previous page links


b-zee:
When either on the first or last page when paginating, the next or previous page link is not visible.

When on page 1, page link to page 2 is not there, for example:
![image](https://user-images.githubusercontent.com/6576760/27176753-24ebe1a0-51c4-11e7-82ea-eb125405653e.png)

Same case for last page:
![image](https://user-images.githubusercontent.com/6576760/27176790-448c482e-51c4-11e7-930f-1d9ed3e45ff2.png)

This seems confusing, as you'd expect the behaviour to be similar through all pages. Like in this scenario:
![image](https://user-images.githubusercontent.com/6576760/27177192-c9963b32-51c5-11e7-9d50-592e0c055f6a.png)

tmorehouse:
Depending on you screen size, some buttons will be "hidden". (i.e. on small mobile screens)

You can also set a limit (`limit` prop on the pagination component) on the maximum number of buttons shown (i.e. set it to 12 in the example above)

b-zee:
The examples shown are on a `lg` viewport.

The `limit` prop is static. I would say a `limit` above 20 is not ever practical either. So with a dynamic amount of pages, or with an amount of pages above 20, the pagination does not work correctly.

Also, with setting a limit, the pagination will break too, e.g. with a limit of 4:
![image](https://user-images.githubusercontent.com/6576760/27186958-36d78a2e-51ea-11e7-86f1-178cb6b262c8.png)

As you can see the first page (1) is not shown (again on `lg`).

tmorehouse:
the `<<` links to the first page in the above.

tmorehouse:
There is a lot of weirdness in how the pagination buttons are rendered.  We're hoping to work on this before release 0.17.0 comes out.

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

Issue #483, opened 2017-06-05
[enhancement] Add features to pagination


stefan-angelov:
I use your pagination but it missing some features. Would you consider to add different HTML output, to be more flexible like https://github.com/sant123/vuejs-uib-pagination ?

alexsasharegan:
I believe the goal of Bootstrap-Vue is to adhere closely to the component library defined by Bootstrap 4. Functionality for common use-cases are often supported, but non-standard bootstrap components are not something intended to be part of bootstrap-vue core. 

Extra/enhanced components is an ongoing topic opened up at #417 . I think your suggestion might fit well into that conversation.

tmorehouse:
While we strive to no bloat most components with overwhelming number of options (there are a few exceptions), we have added some enhancements to pagination with PR #528 which should be available in the next release, coming soon.

tmorehouse:
@stefan-angelov check out the fiddle for the new features coming in the next release: https://jsfiddle.net/5gfjkdfy/

And the update docs at https://github.com/bootstrap-vue/bootstrap-vue/blob/master/docs/components/pagination/README.md

Will some of these changes help you?

stefan-angelov:
In fact, these features are exactly what i need. Thank you

tmorehouse:
Excellent.   This will be available soon.  We don't have an exact date yet for the release, but it i not far off into the future)

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #528, opened 2017-06-20
[pagination] UX enhancements

tmorehouse:
UX Enhancements to `b-pagination`.  Provides a more consistent and expected user experience

- `limit` now refers to the number of buttons (including ellipsis, but not prev/next). Ensures that a consistent number of buttons are always displayed
- Props for ARIA roles
- Props for `ellipsis`, `firstText`, `prevText`, `nextText`, & `lastText` button text
- Props to hide first/last buttons, and ellipsis indicators
- Minor updates

Complete working prototype: ~https://jsfiddle.net/4xajb5ez/~ ~https://jsfiddle.net/et8y5xx3/~ https://jsfiddle.net/5gfjkdfy/

~It is still not perfect on smaller (`xs`) screens for larger `limit` values.~

Addresses issue #523, #508, & #506 and some enhancements for #483


alexsasharegan:
I'll give this a look-see over some coffee tomorrow morning! 👀 ☕️

tmorehouse:
I am contemplating allowing the user to specify the screen breakpoint size (i.e. `xs`, `sm`, `md`, etc) for hiding buttons, or possible defaulting down to the limit 3 size for small screens, as it is still not overly intuitive on small screens.

lib/components/pagination.vue
```diff
@@ -81,219 +87,286 @@
...
-                return this.size ? `pagination-${this.size}` : '';
-            },
-            pageLinks() {
-                if (this.currentPage > this.numberOfPages) {
-                    this.currentPage = 1;
-                }
+        btnSize() {
+            return this.size ? `pagination-${this.size}` : '';
+        },
+        pageLinks() {
```
alexsasharegan: Spent a while grocking that this computed prop is used as a range. Maybe we can add a comment, or just rename that a little more intuitively? `pageLinkRange` perhaps

tmorehouse: Yeah.   that might be a good idea.  It confused me as well.

tmorehouse: I also thought of making it return an actual array of numbers too (and could then ditch the `offset` data property, but that would put a bit more computation/processing in the computed prop

lib/components/pagination.vue
```diff
@@ -81,219 +87,286 @@
...
-                // Defaults
-                this.diff = 1;
-                this.showPrev = false;
-                this.showNext = false;
+            // Defaults
+            this.offset = 1;
+            this.showFirstPage = false;
+            this.showFirstDots = false;
+            this.showLastPage = false;
+            this.showLastDots = false;
```
alexsasharegan: For posterity, maybe each of these special cases can get a little explanation.

lib/components/pagination.vue
```diff
@@ -81,219 +87,286 @@
...
 
-                // If less pages than limit just show this pages
-                if (this.numberOfPages <= this.limit) {
-                    return this.numberOfPages;
-                }
+            // Special case
```
alexsasharegan:
```js
/**
 * Special Case:
 * - Less pages available than the limit of displayed pages
 */

File: lib/components/pagination.vue (hunk 2 above)
Comment (tmorehouse) [inline lib/components/pagination.vue]: I do have comments in the template for their use, but yeah, adding a component would be good for future contributors

File: lib/components/pagination.vue [hunk 4]
```diff
@@ -81,219 +87,286 @@
...
+                return this.numberOfPages;
+            }
 
-                // If at the beginning of the list
-                if (this.currentPage <= this.limit - 2) {
-                    this.diff = 1;
-                    this.showNext = true;
-                    return this.limit - 2;
-                }
+            // Special case
```
alexsasharegan: Still trying to figure this one out. Also curious if it should take the number of pages into account.

alexsasharegan:
I feel like I need a manifest of intent... Under what conditions are the ellipses supposed to be displayed is probably the most difficult to understand part.

alexsasharegan:
Should the ellipses serve the same functionality as the left/right arrows? They don't feel like a smooth UX while I'm playing with the docs sample and jumping around. I want them to do something when I click on them.

alexsasharegan:
So limit is currently the number of page links shown, **_plus_** the bookend pages **_only if_** a bookend page is not contiguous to the limit pages relative to your current position? Why does pagination always have to be such a headache!

tmorehouse:
`limit` sets the total of the page indicators (including ellipsis, nut not including the `prev` or `next` buttons.

Yeah pagination is definitely a headache to catch all the weird cases.

alexsasharegan:
But it doesn't _currently_ (not sure if you've changed this behavior) include the bookend page indicators if they are not contiguous to the current page section. That's a little weird to me.
image

image
image

tmorehouse:
Make the result quadrant wider than `xs` and you will see regular pagination.  THis new code still has the `hidden-xs-down` class on all the non current page links... which is not ideal. see my comment above https://github.com/bootstrap-vue/bootstrap-vue/pull/528#issuecomment-309655140  for some thoughts

tmorehouse:
I might also make a few tweaks to handle odd/even limits (as odd limits appear to work best)

alexsasharegan:
Interesting. I feel like it would be ideal to write out the desired behavior starting with full-screen devices (no special cases) down to mobile screens.

alexsasharegan:
Just for curiosity's sake, here's how amazon does it:
image
image
image
image
image

Obnoxiously, their last page is not even clickable.

tmorehouse:
Yeah, I had to doodle out a bunch of scenarios on paper to get my head around it, and then created an excel spreadsheet that generated all the outcomes based on input values.

And yep, I think the n+1 and n-1 buttons should always appear around the current page selection (if applicable).  Trying to determine the best way top do this.  I used the original pagination as the basis for this PR, but maybe I should have blown it away and re-wrote it from scratch :p

alexsasharegan:
It might be nice to create an object instead of that pageLinks range. That could compute the meat (current page with n + 1/n - 1 type content), the start bookend values (or show disabled when the meat encompasses the start), the end bookend values (likewise), and props for collapsed content left and right.

tmorehouse:
Yeah, was just working on changing the pageLinks to return an array of intermediate pages

alexsasharegan:
Something like this I suppose:

```js
{
  beginning: {
    show: true,
    value: 1
  },
  collapsedLeft: true,
  middle: [3, 4, 5],
  collapsedRight: true,
  end: {
    show: true,
    value: 10
  },
}
```

tmorehouse:
What if I were to get rid of the bookend page buttons (i.e. page 1 and page N), and just add in a `<<` and `>>' before (and after respectively) the `<` and `>`?  I think that would simplify the computations greatly

alexsasharegan:
I like that idea. So one set of navigation arrows pushes back/forward one page, and the other moves to beginning/end?

tmorehouse:
Would we even need the ellipsis with no page 1 and page N bookends? They area good visual clue that there are more pages though.

alexsasharegan:
As much as I want to 86 them, I think they are a nice visual cue.

tmorehouse:
@alexsasharegan 

Check out this latest version: https://jsfiddle.net/et8y5xx3/

It improves the display on smaller screens, and includes always showing N-1 & N+1 buttons (for limits above 3). For xs screen at most 3 buttons are ever shown to keep it compact. Ellipsis will only ever be shown when limit is greater than 4.

alexsasharegan:
That works really nicely!

tmorehouse:
:beers: for sure! ;-)

tmorehouse:
And on XS screens, 3 page buttons plus the 4 next/prev buttons should keep it below 360px wide in most cases for size `md` (default)

tmorehouse:
I wonder if we should add in an option to hide the goto first/last buttons? or leave them in, as they are handy.

alexsasharegan:
They could be a negative type Boolean prop:
```js
{
  props: {
    hideGotoEndBtns: Boolean
  }
}
```

lib/components/pagination.vue
```diff
@@ -10,290 +10,404 @@
...
+              // We are near the end of the list
+              this.showFirstDots = true;
+              numLinks = numLinks - 1;
+              startNum = this.numberOfPages - numLinks + 1;
+            } else {
+              // We are somewhere in the middle of the page list
+              if (this.limit > ELLIPSIS_THRESHOLD) {
+                this.showFirstDots = true;
+                this.showLastDots = true;
+                numLinks = numLinks - 2;
```
alexsasharegan: This might be good to keep `numLinks = this.limit - 2` in case some things change somewhere and mutate the `numLinks` variable.

lib/components/pagination.vue
```diff
@@ -10,290 +10,404 @@
...
-        data() {
-            return {
-                diff: 1,
-                showPrev: false,
-                showNext: false,
-                currentPage: this.value
-            };
+// Make an aray of N to N+X
+function makePageArray(startNum, numPages) {
+    return Array.apply(null, {length: numPages}).map(function(value, index){
```
alexsasharegan: I think this is a learning opportunity for me. Let me talk it out to see if I get it:

You call the true Array.prototype.map method on an object (normally not kosher), but by giving it a length, you're able to iterate the desired length and return the transformed array.

lib/components/pagination.vue
tmorehouse: `numLinks`, currently, is only mutated at most one time, as the ifs are all else ifs.  But yeah, it might be safer to set that as `numLinks = this.limit -2`

lib/components/pagination.vue
alexsasharegan: This is sweet.

tmorehouse: Yep.  rather than looping and pushing onto an initially empty array.  ES6 has `Array.fill(...)`, which makes an array of N length where each element is the same value... which would be nicer, but might require transpiling.

I _aquired_ that snippit from http://www.jstips.co/en/javascript/create-range-0...n-easily-using-one-line/

alexsasharegan: I suppose all your really doing is shorthand for the constructor variety:

```js
const array = new Array(length)
array.map( fn )
// vs.
Array.apply( null, { length } ).map( fn )
```

tmorehouse: What happens is that one is making a null filled array based on the length property, and then running map across it to generate the filled in new array.

alexsasharegan: Nice, I get it. And I like it.

tmorehouse: I could have easily used a while or for loop and pushed the objects onto the array, which might be more clear, but probably not as fast as the native code

alexsasharegan: Looks like `.fill` is an instance method. You might have to do something more like this:
```js
new Array(10).fill(null).map()
```
image

alexsasharegan: That's a nice site for tips!

alexsasharegan: While loops, IMO, are either not as declarative as higher order iterators, or they're valuable for some trickery for generators or recursion.

lib/components/pagination.vue
```diff
@@ -340,6 +343,10 @@ export default {
             type: String,
             default: 'md'
         },
+        hideGotoEndButtons: {
+            type: Boolean,
+            default: flase
```
alexsasharegan: We need a Troy-O emoji 😜

alexsasharegan: Maybe :man_facepalming: ?

tmorehouse:
I have added in `hide-goto-end-buttons` support.

lib/components/pagination.vue
tmorehouse: Hehehe

tmorehouse: Too many fingers on keyboard error :laughing:

alexsasharegan: It's a sign to switch out of GH/Vim and give vscode a shot!

alexsasharegan: 🖖 ⌨️ 🖖

tmorehouse:
I think this is ready to merge.

I'll try and make a test suite for it after the next release.

alexsasharegan:
I'm in a big sprint (more like marathon) to release a new site (using BV of course), but if I can get some time, I may be able to contribute to a test suite. Maybe I'll tackle a different area though since this is more fresh in your mind. I'll ping you if I'm motivated to work towards this though.

tmorehouse:
Yeah, I'm eager to get one of my older sites converted over to BV as well.  I wont be using popovers or tooltips, so I can avoid most of the tether/popper.js stuff, although the new BS 4.beta will use popper for Dropdown positioning, and I do use dropdowns (although I can convert them to a toolbar if necessary)

alexsasharegan:
I do need the Popover to show an address on an order card, but I've been canning all the tooltips for the ridiculous things they can do to the dom.
image

tmorehouse:
Revised jsFiddle https://jsfiddle.net/y7yk5b00/

Issue #529, opened 2017-06-20
Table column max/min width


CrixuAMG:
I would like to see a max/min width option for the table columns, as this would put more table customisation on the table component instead of throughout the project.

tmorehouse:
See issue #492 for a solution using classes in the field definition object.

tmorehouse:
@CrixuAMG I have added a new quick PR that allows you to add CSS styles (in javascript object format) to the table TH cells on a per field basis.

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

Issue #492, opened 2017-06-08
Table column max width


Zyles:
How do you specify a max width for a specific table column?

tmorehouse:
You can specify a "class" on your field definition, and have that class define the width:

```js
fields: {
  foo: {
    'class': 'my-class',
    sortable: true
  },
  bar: {
    sortable: true
  },
}
```

and in your CSS:
```css
.my-class { max-width: 100px; }
```

You can also pass an array of class names to the `class` property.

Zyles:
Thank you.

CrixuAMG:
It would be much better if you could provide a width and set that for the fields.

tmorehouse:
For specifying a width,  then one would need to determine which units to use (%, px, rem, em, vw, etc), which would also impede responsive settings, so classes are the easiest way for now.  It might be something we revisit in the future.

You could also use custom slots and put the CSS on a wrapper div around the data to force a specific width of a column.

tmorehouse:
@Zyles  I have added a new quick PR that allows you to add CSS styles (in javascript object format) to the table TH cells on a per field basis.

damaimrs:
i've added a class inside of my <th> properties, but i cant set the width of it by css :(

this is my source code:

fields: {
        'no': {
          class: 'no'
        },
        'airwaybill_number': {
          class: 'awb_number',
          sortable: true
        },
        'message': {
          class: 'message',
          sortable: true
        }

.no {
  width : 100px !important;
}

jrudnik:
@damaimrs this approach was not working for me as well. However, now there is a neet solution available:
`fields:{
'no':{thStyle: {width: '100px !important'}}
}`
Please refer to https://github.com/bootstrap-vue/bootstrap-vue/pull/531

sevsev9:
I know this is old but it might still help someone.

This approach won't work in a scoped style because the data tag inside the HTML tag won't be written by the Vue framework due to no association.

For this to work you'd have to unscope the style so the style class takes effect on not only specific HTML tags.

So instead of:
```
<style scoped>
    .foo {
          overflow: hidden;
    }
    ...
</style>
```

remove the scoped from the style tag:
```
<style>
    .foo {
          overflow: hidden;
    }
    ...
</style>
```

Or just simply use the solution specified in [#531](https://github.com/bootstrap-vue/bootstrap-vue/pull/531).

tmorehouse:
[...]

PR #531, opened 2017-06-21
[table] Styles on table TH cells, classes on TH and TD cells

tmorehouse:
Adds a new property to the fields object `thStyle` to allow users to apply inline CSS styles ton the header (and footer) `th` elements.

Provides a generic solution for issues #529 and #492


tmorehouse:
I was originally thinking of `min-width`, `max-width` and `width` CSS properties, but this option might be more flexible (i.e. a style object)

tmorehouse:
Also added two new field properties for applying a class only to the `TH` or `TD` cells.  Updated documentation to reflect additions.

This addition will allow user to apply separate classes to the field header and field data cells (i.e text alignment classes, etc) .

Issue #493, opened 2017-06-09
Problem on reset form file


meyt:
When setting `null` value to model of `b-form-file`, filename not changed, 
there is example of my behavior for reset a form file value.
https://jsfiddle.net/3dw5djb1/6/

tmorehouse:
This appears to be how the v-model is internally bound,  It is currently bound as a one way. File inputs are a bit different than other inputs, as they only allow you to programmatically set the value to an empty string, which makes the two-way v-model not work correctly in some cases.

I might be able to modify the behaviour to handle two way binding to allow you to set it to null or an empty string, or introduce a `reset()`method that could be called.

tmorehouse:
Because of the limitation of file inputs in Vue, the PR adds a reset() method to the component that you can call to clear the file from the input.  You will need a reference to the file component:

```html
<div id="app">
  <b-form-file
    ref="fileinput"
    v-model="file"
    choose-label="SELECT"
    placeholder="NO FILE"
  ></b-form-file>
  <br> Selected file: {{file && file.name}}
  <br>
  <b-btn @click="del()">DELETE</b-btn>
```
```js
window.app = new Vue({
  el: '#app',
  data: {
    file: null
  },
  methods:{
    del: function(){
    	this.$refs.fileinput.reset();
    }
  }
});
```

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Please refer to the update docs for details.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #535, opened 2017-06-22
[form-file] reset() method for clearing file value

tmorehouse:
Allow user to reset() `b-form-file` to clear the file input without resetting the entire form

Workaround for issue #493 

Users can call the component's `reset()` method to clear the input, by making a reference to the component:

i.e. `this.$refs.fileinput.reset()`

Includes workaround for IE 10 (file inputs are read-only on IE 8 to IE 10) https://stackoverflow.com/a/35323290/2744776

### DIFF
--- a/docs/components/form-file/README.md
+++ b/docs/components/form-file/README.md
@@ -39,4 +39,31 @@ Also it is advised to use [:lang()](https://developer.mozilla.org/en-US/docs/Web
 }
 ```
 
+### Clearing the file selection
+Because of limitations in the value binding with `<input type="file">` elements, `v-model` for `b-form-file`is
+unidirectional, and cannot be used to set or clear the file(s) selection.  To get around this 
+limitation `b-form-file` provides a `reset()` method that can be called to clear the file input.
+
+To take advantage of the `reset()` method, you will need to obtain a reference to the `b-form-file` component:
+
+```html
+<div id="#app">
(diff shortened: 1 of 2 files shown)


tmorehouse:
`v-model` may not be the best way for handling a reset of the file input. 

I thought about using a validator to ensure the value is null for reset, but value will change when a new file selection is made.

The other option was to create a `reset()` method that could be called by the parent (i.e. `this.$refs.fileinput.reset()`)  as `file` inputs are tricky (and do not support v-model directly).

The only thing here I would be worried about is if the state value bound to v-model in the app changes (to something other than 'null (or the existing file list), say for forms that are swapped in and out of the document on a `multi-page` form.

tmorehouse:
@pi0 @alexsasharegan what are your thoughts on this?

alexsasharegan:
I tried playing with the inputs FileList object, but that's native, read-only kinda stuff. This is probably the most realistic way. I've also seen some people use `$nextTick` and some `v-if` hacks to basically blow away the whole input and re-render it, but we'll avoid that level of hackery 😜

tmorehouse:
Yeah, I wanted to avoid blowing away the input completely, or wrapping the input in another form element, as that could potentially interfere with other stuff.

tmorehouse:
The trick with setting the input type to a null string and back to 'file' works on IE 10 (and other modern browsers don't complain). is pretty clean without having to trash the input or meddle with the DOM.

Issue #537, opened 2017-06-22
[modal] Close on Esc key doesn't work with tabs inside


stefan-angelov:
I use modal with tabs and when i hide the tabs' heads (display: none) and press Esc key the modal doesn't close. JS Fiddle: [https://jsfiddle.net/94rt6vhL/2/](url)

tmorehouse:
Modal focuses the first input/button/link that it finds in the modal content.  In this example you have hidden the tabs tab buttons.  Revealing the tab buttons makes the close on ec work.

See: https://jsfiddle.net/94rt6vhL/3/

Hiding the tab buttons break tab functionality.

tmorehouse:
I have created  PR #538  that may help with this issue.

stefan-angelov:
Thank you for the fast response!!!!

tmorehouse:
v0.17.0 has now been [released](https://github.com/bootstrap-vue/bootstrap-vue/releases/tag/0.17.0) and this issue _should_ be addresed.

Try out the latest, and if you run into issues/bugs, please [create an issue](https://github.com/bootstrap-vue/bootstrap-vue/issues)

PR #538, opened 2017-06-22
[modal] better element visibility checking

tmorehouse:
Provides more robust element visibility checks.

May address issue #537 (where user has hidden the modal content tabs button via style sheet)

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -101,7 +101,17 @@
             return null;
         }
         let els = Array.prototype.slice.call(root.querySelectorAll(selector));
-        return els.find(el => isVisible(el));
+        
+        /* return els.find(el => isVisible(el)); */
+        let el = null;
+        let i = 0;
+        while (!el || i < els.length) {
+            if (isVisible(els[i]) {
+                els = els[i];
+            }
+            i++;
+        }
+        return el;
     }
 
     export default {
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -101,7 +101,8 @@
             return null;
         }
         let els = Array.prototype.slice.call(root.querySelectorAll(selector));
-        
+
+        /* IE 10 & 11 do not support native array.find() */
         /* return els.find(el => isVisible(el)); */
         let el = null;
         let i = 0;
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -105,9 +105,9 @@
         /* return els.find(el => isVisible(el)); */
         let el = null;
         let i = 0;
-        while (!el || i < els.length) {
-            if (isVisible(els[i]) {
-                els = els[i];
+        while (!el && i < els.length) {
+            if (isVisible(els[i])) {
+                el = els[i];
             }
             i++;
         }
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -95,7 +95,7 @@
         return el && (el.offsetWidth > 0 || el.offsetHeight > 0);
     }
     
-    // Find hte first visible element
+    // Find the first visible element contained in a given root element
     function findFirstVisible(root, selector) {
         if (!root || !root.querySelectorAll || !selector) {
             return null;
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -82,13 +82,27 @@
     import bBtn from './button.vue';
 
     const FOCUS_SELECTOR = [
-        'button:not([disabled]):not([style*="display: none"]):not([style*="display:none"])',
-        'input:not([disabled]):not([style*="display: none"]):not([style*="display:none"])',
-        'select:not([disabled]):not([style*="display: none"]):not([style*="display:none"])',
-        'textarea:not([disabled]):not([style*="display: none"]):not([style*="display:none"])',
-        'a:not([disabled]):not(.disabled):not([style*="display: none"]):not([style*="display:none"])',
-        '[tabindex]:not([disabled]):not(.disabled):not([style*="display: none"]):not([style*="display:none"])'
+        'button:not([disabled])',
+        'input:not([disabled])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        'a:not([disabled]):not(.disabled)',
+        '[tabindex]:not([disabled]):not(.disabled)'
     ].join(',');
+
+    // Determine if an HTML element is visible - Faster than CSS check
+    function isVisible(el) {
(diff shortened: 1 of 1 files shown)


tmorehouse:
There are a few components that use the element visibility check, and querySelectors/filters

It might be handy to move these utility functions and common calls to a util/dom.js helper

lib/components/modal.vue
```diff
@@ -82,14 +82,28 @@
...
+    function isVisible(el) {
+        return el && (el.offsetWidth > 0 || el.offsetHeight > 0);
+    }
+    
+    // Find the first visible element contained in a given root element
+    function findFirstVisible(root, selector) {
+        if (!root || !root.querySelectorAll || !selector) {
+            return null;
+        }
+        let els = Array.prototype.slice.call(root.querySelectorAll(selector));
```
alexsasharegan: You might as well use the es2015 `Array.from(root.querySelectorAll(selector));`

tmorehouse: Is that available in IE 10?

alexsasharegan: It should be with Bublé. Otherwise things like `Array.prototype.find` would also fail.

tmorehouse: Nope... IE is a pain in the butt again:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from
image

alexsasharegan: I'm way more familiar with Babel though, so I'd love @pi0 to weigh in on the settings when he has time.

alexsasharegan: Yeah, same with find 
image

tmorehouse: Once BS 4 drops IE 10 & 11 support we could switch to `Array.from(...)`

alexsasharegan: You can always try it, run the build script, then do a find in file for it in the dist files.

tmorehouse: Crap... I think I use `array.find` in a couple of spots. Grrr.

alexsasharegan: Welp, just tried my test, and `Array.from` does **NOT** transpile. Interesting.

alexsasharegan: So yeah, surprises all around!

tmorehouse: How about Array.find?

alexsasharegan: Remains un-transpiled as well. I'm guessing all the es2015-ish array methods might suspect.

alexsasharegan: Things like:
- Array.prototype.some
- Array.prototype.every
- Array.prototype.find

alexsasharegan: Probably also Array.isArray

tmorehouse: Array.filter is supported :)  I can use the filter to return a list of elements passing the test, and then return just the first element... although that might be  waste of cycles

tmorehouse: Or just resort to a while loop.

alexsasharegan: I mean, we could always polyfill:
```js
// polyfill.js
// https://tc39.github.io/ecma262/#sec-array.prototype.find
if (!Array.prototype.find) {
  Object.defineProperty(Array.prototype, 'find', {
    value: function(predicate) {
     // 1. Let O be ? ToObject(this value).
      if (this == null) {
        throw new TypeError('"this" is null or not defined');
      }

      var o = Object(this);

      // 2. Let len be ? ToLength(? Get(O, "length")).
      var len = o.length >>> 0;

      // 3. If IsCallable(predicate) is false, throw a TypeError exception.
      if (typeof predicate !== 'function') {
        throw new TypeError('predicate must be a function');
      }

      // 4. If thisArg was supplied, let T be thisArg; else let T be undefined.
      var thisArg = arguments[1];

      // 5. Let k be 0.
      var k = 0;

      // 6. Repeat, while k < len
      while (k < len) {
        // a. Let Pk be ! ToString(k).
        // b. Let kValue be ? Get(O, Pk).
        // c. Let testResult be ToBoolean(? Call(predicate, T, « kValue, k, O »)).
        // d. If testResult is true, return kValue.
        var kValue = o[k];
        if (predicate.call(thisArg, kValue, k, o)) {
          return kValue;
        }
        // e. Increase k by 1.
        k++;
      }

      // 7. Return undefined.
      return undefined;
    }
  });
}
```

tmorehouse: Yeah, a polyfil could be handy.  I had to polyfill element.closest for ScrollSpy.

lib/components/modal.vue
```diff
@@ -101,7 +101,17 @@
...
         }
         let els = Array.prototype.slice.call(root.querySelectorAll(selector));
-        return els.find(el => isVisible(el));
+        
+        /* return els.find(el => isVisible(el)); */
+        let el = null;
+        let i = 0;
+        while (!el || i < els.length) {
+            if (isVisible(els[i]) {
+                els = els[i];
```
alexsasharegan: Wouldn't a more apples to apples behavior be to return early here?

tmorehouse: Oops els = els should be el = els...

alexsasharegan: Just kidding, the `!el` check saves extra 🍌  loops.

tmorehouse: Fixed now

lib/components/modal.vue
tmorehouse: Although i could check for the native function first, and then fallback to the simple while loop

alexsasharegan: Either way, a few helpful comments for why not using only built-ins is certainly a smart idea. The while loop is probably less code than the feature detection.

tmorehouse:
I could make up a common dom-utils.js helper and start using that in the various components that need to find elements in the doc, or test visibility, etc.  It could wait until after 0.17.0 is released though, so that we don't postpone the next release too much.

alexsasharegan:
Post release 👍

lib/components/modal.vue
tmorehouse: Yeah, for this particular case the loop is pretty small

tmorehouse:
Post release it is for dom-utils ;-)

lib/components/modal.vue
tmorehouse: I found this cool ES6 compatibility table (browser, transpilers, server environments, etc):

https://kangax.github.io/compat-table/es6/

Issue #545, opened 2017-06-24
[form-radio/checkbox/select] contextual states


tmorehouse:
It appears that these components do not reflect the `state` prop setting (success, danger, warning)

tmorehouse:
It appears that `form-select` contextual state may be an issue with Bootstrap CSS

tmorehouse:
PR #547 fixes form-radio

tmorehouse:
This issue is mainly due to the lack of contextual state support directly on form controls.

PR #544, opened 2017-06-24
[docs] form control & button documentation update

tmorehouse:
Updates to the form controls documentation.

Feel free to add additional changes to this branch

PR #547, opened 2017-06-25
[form-radio] contextual state fix

tmorehouse:
Apply contextual state to form-radio

Fix for #545 (form-radio only)


tmorehouse:
This PR should probably be merged before the `docs` PR #544, as we can add additional documentation of the contextual state to the form-radio docs.

PR #556, opened 2017-06-26
[navbar] sticky property should be a boolean, not a string

pimlie:
The sticky property could/should be just a boolean, now you have to type `sticky="blalal"` but just `sticky` should be sufficient. Looking at the bootstrap docs there is no other implementation then `sticky-top` at the moment.


tmorehouse:
This looks like a solid idea.

Issue #417, opened 2017-05-18
[proposal] Extra components


pi0:
Users require more and more enhanced components based on basic BS4 and we cannot add them to basic component set as it makes package bigger and bigger. Also we may keep release cycle / tests easier by keeping all components in a single repo. I'll start providing directory structure and necessary extra build steps for that. 
Please feel free to share your ideas and component requirements here. /CC @bootstrap-vue/dev

alexsasharegan:
So I know there have been a lot of proposals for additional components (myself included), but I vote for this lib to keep a fairly tight focus on implementing just the BS4 catalog. I haven't maintained a big open source lib before, but everyone always says the hardest part is keeping it focused and avoiding unnecessary bloat.

Is that what you mean by "enhanced components"? Most of the issues I've seen with the current component set seem more like bug fixes so components play nice with each other and the outside world.

Aren't all the components already in a single repo?

If you were able to scaffold testing, I would help PR tests to cover the current functionality. I'm not super experienced with setting up a testing environment, but writing them is not a problem.

pi0:
@alexsasharegan About tests there is an open Issue #161 . (Spoiler: We even already had an sponser offer from BrowserStack to provide us free Infrastructure) I can implement tests workflow for both e2e and unit test but that needs lots of time/contribution effort to write all tests. If you guys have enough time i can do this.

About keeping single repo but separate builds i think this can help keep track of core changes easier and tests can run for each **core component** change against all **official extra components**. I'm always open to and rely community so this won't make any development bottleneck and would provide more quality and faster iterations.

Also about **enhanced** components term, i talk about components composed of core ones such as date picker & tables (with built-in pagination etc) that can't be just added to core base.

PierBover:
If all the components were included in the same repo, wouldn't [tree shaking](https://webpack.js.org/guides/tree-shaking/) simply not bundle the components that aren't used in a project?

Anyway, whatever option you decide @pi0 is good:

* including the components in the `bootstrap-vue` repo
* including the components in their own repo in the `bootstrap-vue` organization
* keeping a list of compatible components in other repos

pi0:
[...]

PR #558, opened 2017-06-26
feat(tabs): Added support for vertical tabs

pimlie:
With this commit we support vertical tabs by setting `<b-tabs position="left"` or `<b-tabs position="right"`. The boolean property `bottom` has been removed as that would be redundant.

The only issue I still have is using the `card` property with postition left or right. It seems the `card-block` class has preference over the `col` class causing the tab-content div to have a width of 100% and being displayed underneath the list of tabs. Not sure what the best solution to this is, any recommendations would be appreciated


tmorehouse:
This would be a breaking change in usage, so bottom should be left in place (for now) with a deprecation warning, and a test to see if bottom is in place.

a computed prop could be used to generate the test.

pimlie:
What is the preferred method of generating a warning? Would `console.warn()` be ok?

tmorehouse:
There is a new helper file (`/utils/warn;`) which contains a `warn` method.:

```js
import warn from '../utils/warn';

export default {
    computed: {
        computedPosition() {
            if (this.bottom) {
                warn('b-tabs: prop bottom has been deprecated. Use position instead');
                return `bottom`;
            }
            return this.position;
        }
    }
}

File: lib/components/tabs.vue [hunk 1]
```diff
@@ -132,6 +136,14 @@
             },
             navStyle() {
                 return this.pills ? 'pills' : 'tabs';
+            },
+            computedPosition() {
+                if (this.bottom) {
+                    console.warn('Using the bottom property is deprecated and will be removed in a future version, use position="bottom" instead')
```
tmorehouse: use the `/utils/warn` helper method here,  See comment in the main pull request thread.

pimlie: Sorry, I was a bit too enthusiastic. Have implemented your recommendations.

tmorehouse:
Much better.

I will need to check keyboard navigation to see if any adjustments are needed, as well as documentation.

This may not be implemented in v0.17.0 which is being drafted right now, but may be in release v0.17.1

pimlie:
While implementing these changes in my project I noticed that using `col-auto` to set the width of the tabs column does not always have the results you would like. I have made a new commit for a `col-size` property with which you can set the column size explictly. E.g. `<b-tabs position="left" col-size="4" pills>`

The `col-size` property is just a string which is added to `col-` when set. This does mean that you could get unexpected results by setting e.g. `col-size="4 bg-warning"` as the bg-warning is also just added as class. The reason I used a string was so you could also set `col-size="md-4"`.

If you have a better implementation for this let me know. Maybe instead of `col-size` a `vertical-column-class` property with default value 'col-auto' would be better?

alexsasharegan:
@pimlie, does this also support usage of plain `.col`?
image

pimlie:
@alexsasharegan, nope it doesnt. Allthough you could set `col-size="6"` to have the same result, probably it would be better to have full control with a vertical column class property?

pimlie:
replace col-size prop with 'vertical-column-class', now the class of the column for the tab-list is `col` by default. If you want to change the width you can set `<tabs vertical-column-class="col-auto"/>` or `<tabs vertical-column-class="col-md-6 col-lg-4"/>`

pimlie:
I have added the scss to have the vertical tabs play nicely with cards. This is probably not what you want for the main components, maybe we should keep the tabs component clean and have an extra component with support for positions bottom, left and right?

tmorehouse:
We usually don't want to fiddle around too much with the the CSS just in case things change on bootstrap's CSS

pimlie:
As in never or as long as v4 is still alpha?

tmorehouse:
Once BS V4 becomes beta, things should stabilize.

pimlie:
Well, in the mean time I dont mind supporting this. And if not at least I have seen now it will work with the current implementation :)

fyi, I wont touch this anymore unless you ask me to!

tmorehouse:
@pi0 and @alexsasharegan could you guys take a look at the proposed changes?

pi0:
Sass customizations are a lot! Suggest at lease waiting for BS4.rc before merge into upstream.

alexsasharegan:
I'm traveling today, but I'll see if I can pull down the branch and play with it on the plane!

pi0:
BTW Seems a nice and useful enhanenent. So thanks @pimlie for your awesome contribution for this improvement <3

pimlie:
@tmorehouse Is there anything I can do to get this committed? If you want I can have a look at v4-beta compatibility but I hope you understand I will only spent more time on this if you are still really considering accepting this PR ;)

tmorehouse:
There are some new features in BS V4.beta, and some new components in BootstrapVue v1.0.0.beta that you might be able to refactor some of this for.  Specifically the new `<b-col>` component (waiting final tweaks & merge on the 1.x branch of BootstrapVue)

Maybe if you could refactor against the `1.x` branch?

tmorehouse:
Also, many of ht non-state maintaining components have been converted to "functional components" (which have no `this`).

Morgul:
Just throwing a comment on here that I have a use case for vertical tabs, and I'm in the middle of converting from Angular 1.x (don't ask) to Vue. I'd love to see this make it in.

tmorehouse:
We might need to revisit this now that we have new functional components for rows/columns (note that rows need to be placed in a container element (b-container, probably with the fluid prop set)

May need to nest the components, or do  few v-it/v-else's to place the needed wrappers.

tmorehouse:
Also, the new BS V4.beta line is now in the `dev` branch.

tmorehouse:
@pimlie Note this is being applied to the incorrect branch.  The PR must be applied against the `dev` branch, not `master`

pimlie:
@tmorehouse Yeah sorry, I think I managed to fix the git issues. If not I can create a new PR, but I was actually trying to show you my current idea of the implementation. 

Because I am looking into this and I think we can use just 2 columns. For position top/bottom we can just set both columns to a width of 12, and for position right/bottom we can use `order-12` to move the tab-header column to the back. See my latest commit (currently has hardcoded col properties for position bottom): https://github.com/pimlie/bootstrap-vue/blob/5873fad29242d69ba98840d9541cf2b9f2eb89c4/lib/components/tabs.vue

Although it looks really clean this way, this solution has some drawbacks. First I am not fully sure what the best implementation is to let the user set all possible `b-col` properties for both the tab-header column as the tab-content column. It are quite a lot of properties to include and prefix.

Secondly this means that there can be overlap between properties that are (internally) set by the position property and properties which are set by the user. Eg if you set the position to bottom then both columns need to have a width of 12, but if the user then overrides the width of boths columns  (as we want for position left/right) the columns can be rendered next to each other. 

What are your thoughts about this?

tmorehouse:
The only issue with using order/etc to move the position of the tab controls, is we need to make sure they remain in semantic order for assistive technologies (actually having the tab controls always physically be at the top of the markup is actually good for screen reader users).

One also has to think how this will appear on small screens (should the tabs always move to the top/bottom instead of left/right when on sm screens)

Providing a few options, such as `vertical-breakpoint`, `vertical-tab-cols`, when `position` is set to `left` or `right`, would probably be enough. And when screen width is below `vertical-brekpoint` the tabs always fallback to top position

pimlie:
@tmorehouse I have added the breakpoint properties and added documentation / examples for the position property. Let me know what you think.

fyi, I needed to install `vue-style-loader`, `node-sass` and `sass-loader` npm packages to be able to run the docs.

pi0:
@pimlie Project structure has been changed since this PR was open. That would be awesome if you can retarget changes towards it. 

Thanks again for your help.

pimlie:
@pi0 Sure, I promised I would support this feature so I will.

But erhm, I dont want to sound like a dick as you put much more time in this project as me, but can you give me some confirmation my efforts won't be in vain again?
E.g there have been discussions in the past about
- not wanting to stray away too far from bootstrap functionality it self
- not wanting to include too much custom css (specifically this thread: https://github.com/bootstrap-vue/bootstrap-vue/issues/417#issuecomment-333344622)
- or support components like that through extra-components

As this PR checks (possibly) all those three issues, has a conscious decision been made by the bootstrap-vue team how you want to support the features from this PR? E.g. I'm quite sure the implementation for an extra-component is different then 'just' changing the main tabs component so I'd prefer to know in advance ;)

Thanks! :kissing_cat:

pi0:
@pimlie You are already doing an awesome job and I should say apologizes for any difficulties on this. Actually, the reason behind why not preferring to go too far from Bootstrap itself is if we provide more and more custom CSS rules, it may make both maintenances and theme support harder. What I suggest is extending `b-tabs` in a new component inside extra components.

I've sent you an invitation to our slack group. So can talk farther on this :)

tmorehouse:
I'm wondering if we maybe add an option to pass additional classes to the `tab-container` and the `nav.nav-tabs` container?

Would that be enough to target the components with custom CSS to change the layout to vertical tabs?

tmorehouse:
And since you could pass any class using these proposed props, you can use the responsive class variations to make the tabs switch from vertical to horizontal based on viewport size.

tmorehouse:
So basically we add the following three new props:
- `content-class` (gets added to the `tab-container` `<div>`)
- `nav-wrapper-class` (gets added to the `tab-container` `<div>`)
- `nav-class` (gets added to the `.nav` `<ul>`)

All three could accept a string, array or object of classes

tmorehouse:
Here is a screen shot from a mocked up example:

image

tmorehouse:
Playing where the classes are added, and which classes, I was able to get hte following using the three new proposed props:

image

Using classes `col-auto` (fit to content) and `col` (use remaining available space), the nav section will grow to fit the width of the tab buttons.

Manually ading `card-body` to the `content-class` prop provides the padding on the content, and adding `card-header border-0` to the `nav-class` prop dds the default nav header background and removes the extra border added to the bottom

tmorehouse:
And using responsive classes (for col-{breakpoint} and col-{breakpoint}-auto, and flex-column) one can get the tabs automatically back on top (or bottom) for smaller screens

pimlie:
Just tried it in my browers dev console as well and the html looks like it just might work. Not that I expected any other outcome as it was coming from you ;)

I was thinking that may be it would an interesting option if we could extend `b-tabs` and inject e.g. a `b-col` component where now a plain `<div>` is used as nav-tab container, that way we could benefit from all existing grid features already. But as the `vertical-breakpoint` and `vertical-column-size` properties will be passed as string/number anyway, its not that difficult to just use the bootstrap grid classes directly. So I guess there wont be much benefits for doing that?

Thinking further because of the above, with the move to functional components why wouldn't we implement the functional components as real javascript classes? That way you can really separate all logic in different methods which you can easily override by extending the component class. Are their major drawbacks for doing that? As you would be adding another abstraction layer between vue and bootstrap I'd guess it could be a bit slower?
Currently it seems just so wasteful that almost none of the component specific logic is accessible for derived components to re-use.

So the export would look a bit more like:
```js
// class definition in a separate file
class bTabs extends idMixin(baseClass) {
  static getProperties () {
    return {
      // props
    }
  }
  static renderNavs (t) {
    return h(
      'ul',
      {
        class: [ 'nav', `nav-${t.navStyle}`, { [`card-header-${t.navStyle}`]: t.card, small: t.small } ],
        attrs: { role: 'tablist', tabindex: '0' },
        on: { keydown: t.onKeynav }
      },
      [ buttons, t.$slots.tabs ]
    )
  }
  static render (t) {
    return h(
      t.tag,
      {
        class: [ 'tabs' ],
        attrs: { id: t.safeId() }
      },
      [
        t.bottom ? content : h(false),
        h('div', { class: { 'card-header': t.card } }, [ this.constructor.renderNavs() ]),
        t.bottom ? h(false) : content
      ]
    )
  }
}

// current tabs.js could probably be a default export
// thats the same for all components?
import { bTabs } from '../componentClasses'
export default {
  render (h) {
    return bTabs.render(this)
  },
  props: bTabs.getProps()
}
```
But maybe thats on the roadmap already? Anyway, sorry for going off-topic

tmorehouse:
Hmmm that might be an option.  although you cant use method `h` directly in the methods, you would have to use `this.$createElement(...)` (which is the same function that is passed to the render function)

tmorehouse:
With the `extends` option, you can import an existing component and extend it, replacing/inheriting the imported component. Extends is similar to mixins although the "priorities" or merging is different.

https://vuejs.org/v2/api/#extends

tmorehouse:
As a note, b-tabs is still a full fledged component, not a functional one.  It just has the template replaced with a render function (rather than the build process compiling the template), mainly to improve ES module imports when importing individual components.

tmorehouse:
With the use render functions, we do have a  lot more control over how the final component is built and rendered into HTML (much easier conditional logic, and the ability to add wrapper element conditionally), with out adding a lot of bloat to the templates.

pimlie:
But as a Vue component is not a real javascript class, I guess you dont have access to parent methods? 

Eg you wont be able to do this:
```js
const bTabsExtra = {
  extends: bTabs,
  render () {
    const el = super.render()
    // only change some stuff
   return el
  }
}
```
So you would need to re-implement the whole render method and dont have access to that logic, not?

tmorehouse:
Just looking at the available responsive bootstrap V4 utility classes (http://getbootstrap.com/docs/4.0/utilities/borders/), it appears that the border-* classes are not responsive (it is needed to remove the extra border at the bottom of the nav container when vertical)

Although the other classes (col-*, flex-* are responsive.

tmorehouse:
And now, with render functions, we could easily make it so that the tabs itself wraps itself in a `card` component, without the user needing to wrap the `<b-tabs>`... although this would be a breaking change

tmorehouse:
Playing with the mock-up a bit more, using the following classes for vertical:
image

Will render the following when the screen is below "md" breakpoint:
image

Although it doesn't work well without the `pills` option.

pimlie:
I see the benefits of using the render functions indeed (sry I mixed that up with functional which is only about state), but I think there could be benefits of taking it a step further by adding another abstraction layer with component classes. Especially for users who specifically want to stray away from the default bootstrap behaviour ;)

On topic, looks like you already have it sorted out :tada: I am almost wondering if we really need a separate component for this? We could just make it a documentation issue / example?

tmorehouse:
The three new props would adds extra flexibility (i.e. if people want to use background-color classes, etc, as well).

Right now I'm just toying around to see how complex it would be to add a vertical option.  There are a few issues with responsive switching between vertical/horizontal based on breakpoint.  Bit if it were a non-responsive solution, it might not be that difficult to add a vertical option now (some conditional classes), but still leave in the three new props for added functionality

tmorehouse:
Pending PR #1362 ads support for vertical tabs.

young-altair:
Can **b-tab** work responsively like in the JQuery version on the following page?
[https://www.jqueryscript.net/demo/jQuery-Plugin-For-Responsive-Bootstrap-Tabs-responsive-tab/](https://www.jqueryscript.net/demo/jQuery-Plugin-For-Responsive-Bootstrap-Tabs-responsive-tab/)
Really need this. Thanks

PR #563, opened 2017-06-27
Fix some indent & consistency mistakes

sgraewe:


yordis:
@alexsasharegan pleease use 2 spaces or just switch to tabs 😭 

Also add `.editorconfig` it will help a lot of people by default

tmorehouse:
The currnt XO config specified 4 sapces for indents (no tabs)

yordis:
@thomasfowler I think has been 5 years since the last time I had to use `4` spaces 😄

pi0:
@sgraewe @yordis thanks for this PR and interest. It has some conflicts also we are internally talking for a better linter config. Will apply that soon. Thanks again :)

yordis:
@pi0 just please don't use 4 spaces 🙏 or just use `tab` so you could configure it the way you want but data prove that people prefer spaces on open source and `2` spaces for indentation, unless you just want to see the `php` community

pi0:
We are probably going with standard.js or vue eslint which both advice 2 spaces too so don't worry :) BTW That's just convention! At our company we have serval projects with different conventions, try to adopt with standard and using IDE Eslint extionstions. They do all this things automatically.

PR #566, opened 2017-06-27
Remove explicit Tether dependency

peterblazejewicz:
This is trivial PR:
Tether.js is always installed as part of BS 4 dependency:
https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.6/package.json\#L49
This RP removes some noise from configuration.
Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
   },
   "dependencies": {
     "bootstrap": "^4.0.0-alpha.6",
-    "tether": "latest",
     "vue": "2.*.*"
   },
   "devDependencies": {


tmorehouse:
Bootstrap-vue doesn't require bootstraps Java script code, so this would cause popovers and tooltips to stop working.

peterblazejewicz:
> Bootstrap-vue doesn't require bootstraps Java script code, so this would cause popovers and tooltips to stop working.

Hm, this was not about client code, but tooling configuration. The Tether package is always installed alongside Bootstrap package

pi0:
I think that would be better kept until we fully remove tether dependency. Seems bootstrap project is going to plan same thing, so keeping that won't make any breaking changes. Anyway feel free marge or merge not :))

pi0:
Will apply better patch when migrated. Thanks for PR :)

PR #567, opened 2017-06-27
Remove implicit Yarn dependency

peterblazejewicz:
This is trivial PR
The Yarn is not explicit dependency to the library.
Users are free to user Yarn or NPM, but adding Yarn into script made
it dependency. This fixes this issue while still offering both solutions to the users.

Thanks!


pi0:
OK, NPM Lover :P

corbinu:
Looks like this has regressed

Issue #569, opened 2017-06-28
collapse::toggle::state triggered too many times


yordis:
Everytime I click on the toggle button I see that this emitting happens too many times.

### Versions
- bootstrap-vue: "^0.16.1"
- vue: "^2.3.3"

- [x] Clear the Filter Events
- [x] Other collapser: No

<img width="1539" alt="screen shot 2017-06-28 at 5 39 35 am" src="https://user-images.githubusercontent.com/4237280/27630694-3544ba72-5bc4-11e7-9f22-1982965ebf7c.png">

tmorehouse:
Which version of Bootstrap-Vue are you using, and would you be able to provide a simple fiddle showing the issue?

alexsasharegan:
Also, be sure to clear your dev tools before testing. I know I've seen my events page be full of old data before.

tmorehouse:
Also, do you have any other collapsible elements on your page? Since the state message is emitted to $root, all collapsible elements will receive  copy and filter the event out if it doesn't match the element's ID.

yordis:
Working on it!

yordis:
@tmorehouse @alexsasharegan I found the issue. Everytime I change the page with `vue-route` it will add an extra call.

yordis:
Yeah unless Vue clean the subscription this will never be cleaned up https://github.com/bootstrap-vue/bootstrap-vue/blob/99afe053199ec9d67c06615a4fcef2ae252dfdd6/lib/components/nav-toggle.vue#L52

tmorehouse:
Hmmm, I wonder if there are a fw other components that might have this issue (such as modal).

alexsasharegan:
This is really curious. I'm also using Vue router, and I don't experience this event duplication. Can you maybe dump in some code from your navbar and it's parent component?

tmorehouse:
Also curious if you are using `<keep-alive>` or not.

yordis:
@alexsasharegan my App.vue is doing a full replacement. Also I removed every single element and left off with the header from the basic config
```
<template>
  <router-view></router-view>
</template>
```
I am not using any `keep-alive` what is that?

tmorehouse:
Keep alive keeps a component in memory on the browser, so that when it is flipped out, and then latter returned, it doesn't have to re-render the component(s). https://vuejs.org/v2/api/#keep-alive

alexsasharegan:
[...]

PR #574, opened 2017-06-29
[collapse][nav-toggle] Ensure $root event listeners are removed on unmount

tmorehouse:
Ensure that our state listeners get removed before unmounted.

Possible fix for issue #569 

I'm not positive about the move from `created` to `mounted` for initializing the event listeners though.  Thoughts?


tmorehouse:
Also, wondering if "modal" also has this issue or not?

tmorehouse:
@pi0 and @alexsasharegan 

Could you take a look at this PR? Is enabling the listeners on mounted instead of created OK?

alexsasharegan:
So listening to root events could easily be done on either created or mounted since it's not a dom event. If the event listener makes a dom related change, I saw use the mounted hook instead. Why listen for an event when you haven't even rendered yet?

tmorehouse:
It might be possible that state is issued by another element/component before the component is mounted. that would be my only concern.

alexsasharegan:
Ultimately, I have a gut feeling this isn't a bug (or at least not the bug we think it is), but something else about how the issue creator has the vue app setup.

tmorehouse:
And I was under the understanding that when a component is "taken down", that any component event listeners are also removed.  But I could be wrong

alexsasharegan:
You're not wrong:
https://github.com/vuejs/vue/blob/07a37264c3b56952435331983adcf4081390b400/src/core/instance/lifecycle.js#L130

alexsasharegan:
Let me pull that from master branch: https://github.com/vuejs/vue/blob/master/src/core/instance/lifecycle.js#L127

tmorehouse:
so does `vm.$off()` remove only `mv.$on(...)` events, or does it include `vm.$root.$on(...)` events as well?

alexsasharegan:
I think you just figured it out!

alexsasharegan:
https://github.com/vuejs/vue/blob/master/src/core/instance/events.js#L69-L103

alexsasharegan:
If this is in fact the problem (and it sounds rather promising), I think a mixin might even be warranted. This could allow for a special root event creator method that would manage root events and unbind them so nobody else makes this mistake.

tmorehouse:
so a collapse state listener mixin...

alexsasharegan:
Potentially like this:
```js
export default {
    data() {
        return {
            bv_root_listeners: []
        }
    },

    methods: {
        listenOnRoot(event, callback) {
            this.bv_root_listeners.push({ event, callback })
            this.$root.$on(event, callback)
        }
    },

    destroyed() {
        let listener = undefined
        while ((listener = this.bv_root_listeners.shift())) {
            this.$root.$off(listener.event, listener.callback)
        }
    }
}
```

alexsasharegan:
Or maybe tweak the destroyed loop:
```js
destroyed() {
    while (this.bv_root_listeners.length) {
        let listener = this.bv_root_listeners.shift()
        this.$root.$off(listener.event, listener.callback)
    }
}
```

alexsasharegan:
This is just a different target from something I did previously for binding events outside the component's dom:
```js
export default {
    data() {
        return {
            vmListeners: []
        }
    },

    methods: {
        registerListener(target, event, fn) {
            if (target instanceof EventTarget) {
                this.vmListeners.push({ target, event, fn })
                target.addEventListener(event, fn)
            } else {
                throw TypeError(`Expected an instance of EventTarget interface.`)
            }
        },

        _removeListeners() {
            this.vmListeners.forEach(({ target, event, fn }) => target.removeEventListener(event, fn))
        }
    },

    destroyed() {
        this._removeListeners()
    }
}
```

tmorehouse:
Should we implement the mixin maybe for release v0.17.1?

tmorehouse:
Modal also listens for events on $root as well, and may need a tweak.

alexsasharegan:
I don't think we need to stuff anything else in `v0.17.0`. And with modals, this is exactly why a mixin seems like a good idea to me.

tmorehouse:
Yeah, v0.17.0 has a lot of changes

alexsasharegan:
So, if my latest PR looks good, you could merge that into here and use it for the collapse. I would say do separate PR's for modal & any other affected components.

I think you could keep the calls to register the listeners in the `created` hook. Probably best not to change what ain't broken.

PR #575, opened 2017-06-29
[pagination] fixed referenceerror: i is not defined

pimlie:
Was getting reference errors with nuxt during SSR while rendering the page, this commit solves those reference rerrors


tmorehouse:
Thanks! :+1:

PR #581, opened 2017-06-29
feat(mixin): Automate event registration & removal on root vm

alexsasharegan:
In response to Issue #569 and the conversation on PR #574


lib/mixins/listen-on-root.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Issue #569: collapse::toggle::state triggered too many times
+ * @link https://github.com/bootstrap-vue/bootstrap-vue/issues/569
+ */
+export default {
+    data() {
```
tmorehouse: I winder if this should this be a private prop where it is not "reactive"

alexsasharegan: I don't believe private props are non-reactive [anymore](https://github.com/vuejs/Discussion/issues/353), but perhaps private notation is best.

alexsasharegan: P.S. There's really an art to selecting the line number to comment on. This giant box can really get in the way!

alexsasharegan: I'm not a fan of [this method for non-reactive props](https://github.com/vuejs/vue/issues/1988) because it has the potential to overwrite data.

alexsasharegan: Also, where using Vue here, not Angular or React—if nothing is being rendered by that private prop, Vue won't do anything! Gotta love it

tmorehouse: True ;-)

alexsasharegan: A little more info on `_, $` prefixes and observation: https://github.com/vuejs/vue/issues/665#issuecomment-125265241

alexsasharegan:
I'm plugged this into my own project (seems like a useful tool), and it appears to be working nicely.

alexsasharegan:
As you can see, the object is still observed event with private notation. Maybe it is worth doing in the created hook? It's just a waste of observers and memory.
image

lib/mixins/listen-on-root.js
```diff
@@ -1,15 +1,13 @@
+if (!Array.isArray) {
+    Array.isArray = function(arg) {
+        return Object.prototype.toString.call(arg) === "[object Array]"
+    }
+}
```
alexsasharegan: Not sure if this is necessary, or if bublé will transpile `Array.isArray`

alexsasharegan:
Now there's no observation done. I think this is in the best interest of performance. We can deal with some more abstract code in the lib if it's transparent to our users.

image

lib/mixins/listen-on-root.js
tmorehouse: You could just make a utility function `isArray(...)` that tried for the native `Array.isArray` first, and if not available then use the `Object.prototype.toString(...)` method.

tmorehouse: https://buble.surge.sh/guide/#unsupported-features

alexsasharegan: Well, that answers that!

alexsasharegan:
It's interesting; a lot of Vue blog posts talk about instantiating an empty Vue component and calling it EventBus, but that assumes a way for every component to access the EventBus. Apart from multiple Vue roots, this pattern of using root as the global bus makes more sense as it doesn't assume/require anything (perfect for a library).

tmorehouse:
Yeah, I have seen that done before, mainly for app communication, and global state management (instead of vuex)

alexsasharegan:
I'm using vuex in my project, and while it largely eliminates the need for global events, there are still valid use-cases. For example, toggling a global "Are you sure?" modal (as you've seen screen captured). I also use this root event bussing to trigger the global login modal.

tmorehouse:
I think this is good now. Will merge shortly, then update my other pull request to take advantage of this.

Issue #587, opened 2017-06-30
Error: "Uncaught TypeError: t.includes is not a function"


valexiev:
Thank you for the new release!

I've just updated bootstrap-vue to 0.17.0 and when I run the project an error is thrown:
"Uncaught TypeError: t.includes is not a function", pointing to bootstrap-vue.esm.js.

Does anybody else have the same error?
Any suggestion for fixing it?

alexsasharegan:
Did it give you any line # column #? This sounds like Array.prototype.includes is in the source, and is not transpiled. What browser vendor and version were you using to produce the error?

alexsasharegan:
I believe I have found all the uses of `Array.prototype.includes` in the source. Bublé (transpiler) is more opinionated than babel, and does not transpile this. PR coming shortly.

alexsasharegan:
@tmorehouse, you don't think this will make it into 0.17.1?

tmorehouse:
0.17.1 is already out :smile:

alexsasharegan:
Haha, then I think you're right!

alexsasharegan:
Fix is in PR #589

valexiev:
Thanks for the prompts responses!

The browser is Chromium v46.

In the /dist/bootstrap-vue.esm.js this is the only occurrence of t.includes:
`function omitLinkProps() {
	for (var t = [], e = arguments.length; e--;) t[e] = arguments[e];
	return Object.keys(props).reduce(function(e, n) {
		return t.includes(n) || (e[n] = props[n]), e
	}, {})
}`

alexsasharegan:
Thanks for the update! Confirms my suspicions. Fix is merged and will appear in `v0.17.2`.

tmorehouse:
I'll close this issue.

v0.17.2 should be available soon.

invergent:
It's 2019, I am still getting the same error **"TypeError: t.includes is not a function"**. Although, there is a bit of a twist to it. I only get the error when using a mobile device to access the app.

tmorehouse:
@invergent what version are you using, and what browser?  if using IE mobile (or other older browsers), make sure you are using the latest polyfils.

PR #589, opened 2017-06-30
fix: Remove usage of es2015 Array.prototype.includes

alexsasharegan:
Resolves issue #587


alexsasharegan:
Darn es2015 is part of my regular language already!

tmorehouse:
Heheh.

mosinve:
maybe better use polyfill suggested at MDN?
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

alexsasharegan:
We don't currently have a place for polyfills to be injected, otherwise I'm a polyfill first kind of guy.

alexsasharegan:
And, unfortunately, they can be larger than just using utility methods.

PR #593, opened 2017-06-30
fix(modal): use listenOnRoot mixin

tmorehouse:
Uses new listenOnRoot mixin by @alexsasharegan to ensure `$root` event listeners are removed on destroy


tmorehouse:
the callbacks?  I was thinking about that.  just needed a name for them

tmorehouse:
moved the handlers into methods.

Issue #542, opened 2017-06-24
[form-*] Unused props on some form-controls


tmorehouse:
Some of the controls have props inherited from the form mixin that do not apply:

| prop | form-input | form-radio | form-checkbox | form-select | form-file
| ---- | ----- | ----- | -------- | ------ | ----
| `plain` | :x: | :white_check_mark:  | :white_check_mark:  | :white_check_mark:  | :white_check_mark: 
| `state` | :white_check_mark: | :white_check_mark: | :x:  | :x:  | :x: 
| `size` | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :x: 

I suggest that we migrate these props from the form mixin to the respective components, or create a modular import mixin similar to the new link mixin.

tmorehouse:
It appears that Bootstraps custom file input does not support contextual states... but perhaps the beta release will.

tmorehouse:
It also appears that b-form-radio and b-form-check custom inputs do not apply the state properties correctly.

tmorehouse:
I also think that we should add an `aria-invalid` prop that will set on textual inputs (adds `aria-invalid="true"` attribute).

tmorehouse:
PR #547 fixes form-radio and adds contextual state

tmorehouse:
`form-checkbox` looks like contextual state will only be applied when the parent container has the contextual state (`has-*), such as in side a form-fieldset

tmorehouse:
Merged

PR #604, opened 2017-07-01
feat(form controls): Optimize props

tmorehouse:
Optimize the props on all form input controls, removing unused props off of some controls.

This addresses #542 and should also optimize the docs


tmorehouse:
Also introduces a new mixin for custom form controls.

PR #609, opened 2017-07-01
fix(form controls): Apply only required props & classes

tmorehouse:
Made sure that only applicable classes and props are on each input type.

Will also improve documentation for prop listings.


tmorehouse:
This PR aligns the form-controls closer to Bootstrap V4.alpha.6

Issue #617, opened 2017-07-02
[dropdowns] Rename slot on nav-item-dropdown


yordis:
I would suggest to change `text` slot name to `trigger` on https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/nav-item-dropdown.vue#L15

If I put an image inside only, like an avatar, then it will be no longer a `text` and like another use cases ... so I think that `trigger` will make more sense

tmorehouse:
This would be a breaking change, so it might be a good candidate change for when BS V4.beta.` comes out.

yordis:
@tmorehouse it's better to do it now that the project is constantly updating that wait until more people use it and it's hard for them to switch, talking from my own experience

yordis:
@tmorehouse I would say over here as well https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/dropdown.vue#L14

tmorehouse:
@pi0, what do you think about changing the slot name from `text` to `trigger`? (this would need to be done in all the dropdown variants).

It might be possible to deprecate the slot `text` using a condition test for the slot name in `this.$slots['slotname']` and use v-if to render the slot.

tmorehouse:
Although trigger is still wrapped inside a button element, so it might be better to be called `trigger-content` or `button-content`, or similar. as some people might assume it is a replacement for the trigger button as a whole.

tmorehouse:
PR #627 addresses this issue, by adding the new slot `button-content`, while retaining support for the deprecated slot `text`

tmorehouse:
@yordis 

Release v0.18..0 is now available (latest), and there is a new slot named `button-content` that deprecates the old slot `text`

PR #627, opened 2017-07-03
feat(dropdowns): Various optimizations for dropdown components 

tmorehouse:
- Deprecate slot `text` in favour of slot `button-content`(addresses issue #617)
- Migrated data(), common props, methods & clickout into dropdown mixin
- Removed prop `link` as setting `variant="link"` does the same thing
- Removed undocumented `<b-button-group-dropdown>` in favour of using `<b-dropdown>`


tmorehouse:
@alexsasharegan & @pi0 

`<b-button-group-dropdown>` is basically identical to `<b-dropdown>`, with the exception of one additional class `.btn-group`.

The regular `<b-dropdown>` only uses this class when in split mode, but when dropdowns are in a button group, it needs the `btn-group` class always, regardless if it is in split mode or not.

So maybe we just always add `btn-group` class to `<b-dropdown>` and dispose of the `<b-button-group-dropdown>` component altogether? And then just update the button-group examples to include a dropdown?

<b-button-group-dropdown>` also had a few code issues that prevented it from working properly

tmorehouse:
Removing the `<b-button-group-dropdown>` would also save on some code space.

tmorehouse:
I have removed `<b-button-group-dropdown>`, as it wasn't documented and had bugs in it, and the regular `<b-dropdown>` will work inside `<b-button-group>` and `<b-button-toolbar>`

tmorehouse:
One question I need to check is if mixin's can import mixins, or can only components bring in mixins?

pi0:
> mixin's can import mixins

Both yes and not! :D We can import plain mixins and mixing them toghether using `Object.assign`

tmorehouse:
Hehe, that was my suspicion.

I'll switch to Object.assign.

tmorehouse:
dropdown mixn now imports clickout and merged with Object.assign where needed.

lib/mixins/dropdown.js
```diff
@@ -80,7 +81,7 @@ export default {
             }
         }
     },
-    methods: {
+    methods: Object.assign({}, clickOut.methods, {
         noop() {
```
pi0: Also we can spread methods into object like:
```
methods: { 
  ...clickout.methods,
  noop()
}
```

tmorehouse:
I wasn't sure if we could use spread here or not since IE sucks at it

image

tmorehouse:
@pi0 Is buble enabled to transpile object spread?

lib/mixins/dropdown.js
tmorehouse: I wasn't sure if we could use spread here or not since IE sucks at it without a transpiler.

image

tmorehouse: @pi0 Is buble enabled to transpile object spread?

tmorehouse: I just checked buble's docs and see that it does transpile object Spread

tmorehouse: Although it appears to not be enabled:
> (buble plugin) Object spread operator requires specified objectAssign option with 'Object.assign' or polyfill helper. (100:13)

tmorehouse:
It appears that object spread support is not enabled in Buble:
> (buble plugin) Object spread operator requires specified objectAssign option with 'Object.assign' or polyfill helper. (100:13)

mosinve:
about error at CI - https://github.com/vuejs/vue-hackernews-2.0/issues/87

tmorehouse:
And Object.assign in not supported in IE at all :(

image

But at least the babel polyfill will fix this for IE users

tmorehouse:
I wonder if this should be added to the build config?

```js
module.exports = {
	...
	buble: {
		objectAssign: 'Object.assign',
	},
}
```

mosinve:
I worth to try

tmorehouse:
Having the spread operator would be handy in a lot of places, and make the code more concise, and once IE dies.... we can turn it off ;-)

mosinve:
This works!

buble({objectAssbuble({objectAssign: 'Object.assign'}),ign: 'Object.assign'}),

tmorehouse:
Cool.... so that is in rollup.config.js?

tmorehouse:
I assume you meant: `buble({objectAssign: 'Object.assign'})`?

mosinve:
buble({objectAssign: 'Object.assign'}),


at rollup.config. ill make a commit to your branch

mosinve:
errr I got github banned at daywork :(((

tmorehouse:
I'll pop it into rollup.config.js

alexsasharegan:
Yay! 😃 :tada: I love object spread!

tmorehouse:
Yay, it worked! :smiley:

tmorehouse:
@alexsasharegan  once this is merged your link mixin stuff could take advantage of it :smile:

But maybe after 0.18.0 comes out.

alexsasharegan:
I'm tinkering with a functional link component right now. 
 😃 Spread definitely helps!

tmorehouse:
This PR saves us, net,  one line of code:
image

:stuck_out_tongue_closed_eyes:

alexsasharegan:
#winning

tmorehouse:
Although some of that is new comments ;-)

tmorehouse:
@alexsasharegan  have you had a chance to look over the changes in this PR?

alexsasharegan:
Is there a reason why the tests throw deprecation notices like it's the 4th of July? _(which it almost is)_ Is this just how the tests were written?

image

tmorehouse:
Ah, each time the component is instantiated it will generate that warn.

mosinve:
maybe rewrite tests to use actual prop instead deprecated?

tmorehouse:
Which test is using fom-fieldset?

alexsasharegan:
Looks like the table test

tmorehouse:
PR #630 fixes the fieldset deprecation warnings in the table tests.

tmorehouse:
Those console deprecation warnings should be gone now (table tests)

tmorehouse:
Safe to merge you think?

PR #630, opened 2017-07-04
fix(table demo.html): remove deprecated fieldset prop

tmorehouse:
Change deprecated fieldset prop from `label-size` to `label-cols`

To prevent erroneous console warnings during test


alexsasharegan:
Woot! Following our own upgrade guide!

PR #646, opened 2017-07-05
[WIP] perf(components): switch stateless components to functional components

alexsasharegan:
# To-do's
- [x] Refactor stateless components to functional components.
- [ ] Add test coverage for all refactored components.
- [ ] Document breaking changes

I figured I would work on this in the open so everyone can weigh in on/get familiar with a refactor to functional components for several stateless components. There are some key differences from single file `.vue` components that I'm still learning about, but the main info can be found here: https://vuejs.org/v2/guide/render-function.html#Functional-Components.

## Benefits
Functional Components (**FC**'s) are stateless (no `$data`) and instanceless (no `this`). As such, they incur no cost of observation, no time spent on component initialization, and consume less memory at runtime. 

They have slightly different render behavior than you may be accustomed to. FC's render _"eagerly"_ in that the are called again every time their parent re-renders. Virtual dom diffing optimizations may still be made, but there is currently no Vue mechanism in place to memoize the return of a FC and avoid calling the function with the same input twice. As a FC should be a pure function, it should theoretically be possible to memoize output, but FC's should be _"cheap enough"_ to render that the net perf improves regardless of memoization.

## Possible Gotchas
- No `data` (FC's are stateless)
- No `this` (FC's don't get instantiated, they are simply called as functions and rendered. They hold no instance memory)
  - Not able to emit Vue events because cannot because not a vue instance (still can listen and register callbacks on both vue and native events)
- Pure functions (cannot/should not create any side-effects)
- If the root of a FC is an HTML element, the vnode will not be a `vm`, but rather a simple element.
  - This means no `vm.$el`, just `el`
  - No need for `v-on:event.native` when root of FC is `HTMLElement`
- Prop and attribute merging is largely the burden of the component (especially depending on how many props the component overrides vs. merges)
- Not restricted to a single root node. Can return an array of vnodes at root level.

## Release
I say this could be a v1.0.0 release goal. I would technically place this under a breaking change because switching to FC's will change how the nodes are referenced and event handlers are attached.

## Component Todo List
A running list of BV UI components to refactor to FC's. Feel free to add to list!

- [x] `badge`
  - [x] test suite
- [x] `breadcrumb`
  - [x] test suite
- [x] `button-group`
  - [x] test suite
- [x] ~~`button-toolbar`~~ 
  - _Dropped due to focus event handling_
- [x] `button`
  - [x] test suite
- [x] `card-group`
  - [x] test suite
- [x] `card`
  - [x] test suite
- [x] `carousel-slide`
  - [x] test suite
- [x] `dropdown` testing
- [x] `dropdown-divider`
- [x] `dropdown-header`
- [x] `dropdown-item-button`
- [x] `dropdown-item`
- [x] `form`
- [x] `link`
  - [x] test suite
- [x] `input-group-addon`
  - [ ] test suite
- [x] `input-group-button`
  - [ ] test suite
- [x] `input-group`
  - [ ] test suite
- [x] `jumbotron`
  - [ ] test suite
- [x] `list-group`
  - [ ] test suite
- [x] `list-group-item`
  - [ ] test suite
- [x] `media`
  - [ ] test suite
- [x] `nav-form`
  - [ ] test suite
- [x] `nav-item`
  - [ ] test suite
- [x] `nav-text`
  - [ ] test suite
- [x] `nav`
  - [ ] test suite
- [x] `navbar`
  - [ ] test suite
- [ ] `navbar-brand`
  - [ ] test suite
- [x] ~~`progress`~~
  - _Dropped due to styling in FC context_

## Additional Resources
- [Emit event from functional component](https://forum.vuejs.org/t/emit-event-from-functional-component/13959)
- [[render functions] Multiple event handlers on the same event](https://forum.vuejs.org/t/render-functions-multiple-event-handlers-on-same-event-solved/13755)
- [Access router store from functional component](https://forum.vuejs.org/t/can-router-store-be-accessed-from-functional-component-solved/13641)
- [[render functions] Class merge strategy](https://forum.vuejs.org/t/render-function-class-merge-strategy-solved/13585)


mosinve:
can they support Vue event system? or they support only pure JS eventing?

alexsasharegan:
So I'm still experimenting and finding the limits, but I do not think that functional components can emit custom Vue events as this might be considered a side effect. They can bind listeners/register event handlers to both native events and other Vue events.

FC's are not limited to a single root node either.

alexsasharegan:
FC's do have access to their parent, so I suppose you could always grab the parent vnode and emit an event on root, but that feels hacky

alexsasharegan:
Yeah, since FC's are instanceless they don't have an emit method or a vnode target to bind to

tmorehouse:
https://github.com/vuejs/vue/blob/dev/src/core/vdom/vnode.js

It looks like a functional component can get a context (`vnode.functionalContext`), if you have access to the FC's root vnode.

lib/utils/assign.js
```diff
@@ -0,0 +1,3 @@
+// Aliasing Object.assign allows the minifier to shorten `Object.assign` to a single character variable.
+// Shortens 13 bytes to 1 at the cost of another allocation.
+export default Object.assign
```
tmorehouse: Iwonder if we stuck the polyfill code in here, and if Object.assign doesn't exist, we return our own object assign function?

The polyfill code is relatively small.:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill

Similar to what you did with `Array.includes` (although slightly more code).

tmorehouse: And once IE 11 dies, we can remove the extra code.

mosinve: @tmorehouse dream on...dream on )))

alexsasharegan: I kinda thought this affords that opportunity. I also just got tired of so many `Object.assign`'s.

alexsasharegan:
Troy, as far as I can see, the context is what is passed to the FC's render function as the second argument. It only has a reference to the parent: 
image

lib/utils/uid.js
```diff
@@ -4,8 +4,7 @@ let counter = 0
  * @link https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L16143
  * @param {string} prefix
  */
-export default function uid(prefix = "") {
+export default function uid(prefix = "bv_") {
     const id = ++counter
```
tmorehouse: With this be OK with SSR stuff? where the counter never gets reset as long as the server i running?  And will it conflict with any IDs generated by the client (which will start off at 0)?

alexsasharegan: The SSR stuff is a really good point. I would think an active node process would just keep incrementing the counter up to infinity (and beyond)!

I'm not sure if this will get used or not (and wiped), but I thought it might be handy to pluck this method off lodash for internal component keying in any lists. It's always nice to have one canonical way to do things. I was hearing some things about how vue tries to be efficient about rendering lists, and optimizes the process by reusing nodes. Long story short, I'm not sure if just using the index as a key really makes a node consistent (although it is unique). This is kind of pinned for more research later.

You may notice my dev style is to hoard my tools, get the job done, then trim everything down tight. If I end up implementing it anywhere, I'm gonna keep my eye open for SSR issues! That's a good point.

alexsasharegan: Much better now thanks to your idea, @tmorehouse !

tmorehouse: Excellent.

As long as the uids are only used for internal stuff, and not element id's, or other prop defaults, we should be safe on SSR

alexsasharegan: I've heard you talk about generating element id's before—what's the danger there?

tmorehouse: Since element IDs are a prop, and uses doesn't supply an ID and we try to generate one (i.e. similar to your uid function), the auto-id provided by the SSR rendering will most likely not match with the ID value generated by the client code, and they conflict, so the client bales rehydration of the server rendered markup and blows it away and regenerates it using the client generated ID.  so one different "default" value for a prop (between server and cli3ent) can cause the component (or evn the app) to re-render client side.

alexsasharegan: I see _(generally at least)_. I def need to keep SSR more in mind.

lib/components/col.js
```diff
@@ -0,0 +1,105 @@
...
+            default: "div"
+        },
+        col: {
+            type: Boolean,
+            default: false
+        },
+        size: {
+            type: [String, Number],
+            default: null
+        }
```
tmorehouse: I m thinking `size` prop should be called `cols`, so as to not confuse/conflict with the `size` prop used on many other components (i.e. button, button-group, etc) which refers to values of `sm`, `lg` and `md` for button sizes.

b-form-fieldset now uses the `cols` prop instead of `size` to prevent user confusion (a few issues were generated for-form-fieldset because of confusion as to the value)

lib/components/col.js
```diff
@@ -56,7 +56,7 @@ export const props = assign(
             type: Boolean,
             default: false
         },
-        size: {
+        cols: {
```
tmorehouse: :+1:

tmorehouse: Don't forget about line 98... still refers to props.size

lib/components/col.js
```diff
@@ -95,7 +95,7 @@ export default {
 
         classList.push({
             col: props.col || classList.length === 0,
-            [`col-${props.size}`]: props.size
+            [`col-${props.cols}`]: props.cols
```
tmorehouse: :+1:

Issue #650, opened 2017-07-05
Table+Pagination + Filter don't work


itielshwartz:
You can see it on the docs:
https://bootstrap-vue.js.org/docs/components/table/ has a bug with pagination + filter
-> when doing a filter the number of pages doesn't change.
this happens because when using the filter on the table the actual `items` won't change  -> causing a wrong number of pages.

I found the source of it:
```
                items = items.slice();

                // Apply local filter
                if (filter && !this.providerFiltering) {
                    if (filter instanceof Function) {
                        items = items.filter(filter);
                    } else {
                        let regex;
                        if (filter instanceof RegExp) {
                            regex = filter;
                        } else {
                            regex = new RegExp('.*' + filter + '.*', 'ig');
                        }
                        items = items.filter(item => {
                            const test = regex.test(recToString(item));
                            regex.lastIndex = 0;
                            return test;
                        });
                    }
                }
```

The table inner filter dosn't effect the real items,
Sadly im still new to Vue (and front end) so not sure how to fix...

tmorehouse:
Since pagination controls are one-way props **into**  `<b-table>`, unfortunately it can't update the number of pages in the parent component context.

I have been contemplating creating a new event to emit that returns the number of pages and current page number) when filter is applied, and deactivated.

tmorehouse:
We don't alter (mutate) the original items, as that wouldnt be nice to your parent component.  So we make a shallow copy of the array and manipulate that.

itielshwartz:
@tmorehouse can't the table expose the new items? so we can just use it from outside?
just add it as another data field (shallowCopyItems), and make the shallow copy change the data.shallowCopyItems? 

Sorry can't help more.

tmorehouse:
Vue 2.x only allows a single two-way prop (the v-model), which is currently bound to the value (the currently displayed items).  All other props are one-way (top down).  See http://vuejs.org/v2/guide/migration.html#twoWay-Prop-Option-removed

The only other way to communicate upwards is via emitting events.

mosinve:
It's strange, as i fixed this long ago... 
the reason of bug IMHO, here https://github.com/bootstrap-vue/bootstrap-vue/blob/master/lib/components/table.vue#L381
emitted filtered and paged data instead only filtered.

tmorehouse:
[...]

PR #652, opened 2017-07-06
feat(table): Emit event when local filtering changes number of result items - issue #650

tmorehouse:
Addresses issue #650 

When local filtering causes the number of pages to change, emits a `repaginate` event with a single argument of the total number of records after filtering (and before pagination slices the data).

Can be used to re-set an associated `<b-pagination>` so that it adjusts to the new number of pages.

This PR includes documentation of the new event in `meta.json`, and updates live example in docs

### DIFF
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -131,7 +131,7 @@ export default {
 
 ### `fields` prop
 The `fields` prop is used to display table columns. 
-keys are used to extract real value from each row.
+Keys are used to extract real value from each row.
 Example format:
 ```js
 {
@@ -358,7 +358,7 @@ possible to filter data based on custom rendering of virtual columns. The items 
 is stringified and the filter searches that stringified data (excluding any properties
 that begin with an underscore (`_`) and the deprecated property `state`.
 
-Thw `filter` can be a string, a `RegExp` or a `function` reference.  If a function
+The `filter` can be a string, a `RegExp` or a `function` reference.  If a function
 is provided, the first argument is the original item record data object. The
 function should return `true` if the record matches your criteria or `false` if
 the record is to be filtered out.
@@ -372,7 +372,7 @@ The built-in default `sort-compare` function sorts the specified field `key` bas
 on the data in the underlying record object. The field value is first stringified
 if it is an object, and then sorted.
 
-The default `sort-compare` routine **cannot** sort virtual columns, nor can it sort
+The default `sort-compare` routine **cannot** sort neither virtual columns, nor 
 based on the custom rendering of the field data (which is used only for presentation).
 For this reason, you can provide your own custom sort compare routine by passing a
 function reference to the prop `sort-compare`.
@@ -514,7 +514,7 @@ following `b-table` prop(s) to `true`:
 When `no-provider-paging` is `false` (default), you should only return at
 maximum, `perPage` number of records.
 
-Note that `<b-table>` needs refernce to your pagination and filtering values in order to
+Note that `<b-table>` needs reference to your pagination and filtering values in order to
 trigger the calling of the provider function.  So be sure to bind to the `per-page`,
 `current-page` and `filter` props on `b-table` to trigger the provider update function call
 (unless you have the respective `no-provider-*` prop set to `true`).
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -32,16 +32,18 @@
            :filter="filter"
            @repaginate="repaginate"
   >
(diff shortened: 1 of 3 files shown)


docs/components/table/README.md
```diff
@@ -46,37 +47,40 @@
...
-      }, {
-        isActive: false, age: 22, name: { first: 'Geneva', last: 'Wilson' }
-      }, {
-        isActive: true, age: 38, name: { first: 'Jami', last: 'Carney' }
-      }, {
-        isActive: false, age: 27, name: { first: 'Essie', last: 'Dunlap' }
-      }
-    ],
+    items: items,
+    totalRows: items.length,
```
mosinve: We dont even set initial number = items.length, as returned value from table without active filter will be equal to items.length. IMHO

lib/components/table.vue
```diff
@@ -375,11 +396,18 @@
...
                 // Apply local pagination
                 if (perPage && !this.providerPaging) {
+                    // Grab the current page of data (which may be past filtered items)
                     items = items.slice((currentPage - 1) * perPage, currentPage * perPage);
                 }
 
                 // Update the value model with the filtered/sorted/paginated data set
                 this.$emit('input', items);
+
+                if (numOriginalPages !== numFilteredPages) {
```
mosinve: Maybe move this to watcher? 
some as
```
    numFilteredItems(newVal, oldVal) {
        if (newVal !== oldVal)
            this.$emit('repaginate', newVal);
    }
```

tmorehouse: Hmmm that might be an idea, but I would have to make a few more data props to keep state while it is calculating it, since the only pace these values are available is in the _items computed prop.

tmorehouse:
I could also add an event that emits the actual filtered item list (maybe a `filtered` event, which has one argument which is a copy of the complete filtered items list)

mosinve:
thats exactly how i made it :) 
But do we really need it, if we'll have a total number emitted for pagination?

tmorehouse:
Yeah, it might just be a niche case.  There was one issue that was asking how to get the filtered data from table.  Easiest way (currently) to to set page size to null or 0 and then read the v-model value.

mosinve:
Then in order to not generate much events, make table emit 'filtered event with 'filtered items, and use it  to set pagination (items.length)?

tmorehouse:
Hmmm... That might kill 2 bids with one stone.

I was just thinking of a way to minimize the number of times events are fired.  By checking if the page count has changed, then if items went from 14 to 13 (with  page size of 5,), then the event wouldn't fire, because the page count hasn't changed, rather than emitting every time `_items` is called..

But, it might just be simpler to set pagination with the "true" filter items length every time.

I'll tweak this tomorrow morning,.

mosinve:
And is  there support of server-side generated pagination info, like from Laravel for example?

tmorehouse:
Yep, there is support for server side pagination built into `b-table` now, 

If you are using item provider function you still need to provide pagination props to b-table to trigger a refresh of the items provider function.

But if you provide an reference to an array to items (not a function), then no need to tie pagination props to b-table, as updating the array will trigger b-table to redraw.

tmorehouse:
@mosinve @pi0 @alexsasharegan 

Take a look at these changes/additions when you get the chance

lib/components/table.vue
```diff
@@ -344,10 +342,14 @@
...
                 }
 
+                // Shallow copy of items, so we don't mutate the original array order/size
                 items = items.slice();
 
                 // Apply local filter
                 if (filter && !this.providerFiltering) {
+                    // Number of Items before filtering
+                    let numOriginalItems = items.length;
+
```
alexsasharegan: Any reason why this isn't a `const` declaration?

lib/components/table.vue
```diff
@@ -363,6 +365,11 @@
                             return test;
                         });
                     }
+
+                    if (numOriginalItems !== items.length) {
+                        // Emit a filtered notification event, as number of items has changed
+                        this.$emit('filtered', items);
+                    }
```
alexsasharegan: This seems nice! Only emit when there's a difference.

lib/components/table.vue
tmorehouse: Ah, a leftover when I was overwriting the value in a previous commit.  I'll fix.

lib/components/table.vue
tmorehouse: i kinda wish it was possible to do multiple v-models on a component

PR #656, opened 2017-07-06
feat(utils): wrap-up as ES6 module

mosinve:

PR #658, opened 2017-07-06
feat(utils): transpiler friendly polyfills and methods

alexsasharegan:
This PR needs to wait for @mosinve's PR #656 so changes can be merged by hand.
- [x] merged

# Update
There's a little bit of formatting junk in here, but this is mainly a move to consolidating some modern language features into central modules. This allows for polyfilling and minification optimization.

## Introduces:
- `Object.assign` polyfill
- `Array.from` polyfill
- `Array.isArray` polyfill _(existed previously)_


lib/utils/array.js
```diff
@@ -0,0 +1,90 @@
+// Production steps of ECMA-262, Edition 6, 22.1.2.1
+// es6-ified by @alexsasharegan
+if (!Array.from) {
+    Array.from = (function() {
```
tmorehouse: Do we want to actually polyfill `Array.from`, or just assign `from` the code and export `from`?

lib/utils/object.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Aliasing Object[method] allows the minifier to shorten methods to a single character variable,
+ * as well as giving BV a chance to inject polyfills.
+ * As long as we avoid
+ * - import * as Object from "utils/object"
+ * all unused exports should be removed by tree-shaking.
+ */
+
+export const assign = Object.assign
```
tmorehouse: the `Object.assign` polyfill is quite short and could be assigned to `assign` if no native `Object.assign`

alexsasharegan: Yes, I'll add that in!

lib/utils/array.js
alexsasharegan: I think so. It's more powerful than other hacks like `Array.prototype.slice.call` Because it allows a mapping function to transform values. Plus, by using es6 like this, we will eventually be able to remove this entirely without any other refactoring.

lib/utils/object.js
```diff
@@ -6,6 +6,35 @@
...
+            // TypeError if undefined or null
+            throw new TypeError("Cannot convert undefined or null to object")
+        }
+
+        let to = Object(target)
+
+        for (let index = 1; index < arguments.length; index++) {
+            const nextSource = arguments[index]
+
+            if (nextSource != null) {
```
tmorehouse: If we turn on ESLint again, w'll probably need a lint comment to stop it from complaining about non strict inequality :stuck_out_tongue_closed_eyes:

alexsasharegan: I would actually prefer commenting the ignore line. I only do this when I have the explicit intention of using JS coercion!

alexsasharegan:
There's a little bit of formatting junk in here, but this is mainly a move to consolidating language features into a central location. This allows for polyfilling and minification optimization.

lib/components/breadcrumb.vue
```diff
@@ -18,9 +18,10 @@
 <script>
 import bLink from './link.vue';
 import { props as linkProps } from '../mixins/link';
-import {arrayIncludes} from '../utils';
+import { arrayIncludes } from '../utils/array';
```
mosinve: why direct import here? isn't it work from index? (just to know, for future)

lib/components/table.vue
```diff
@@ -76,14 +76,15 @@
 </template>
 
 <script>
-    import {warn} from '../utils';
+    import { warn } from '../utils';
+    import { keys } from '../utils/object';
```
mosinve: interesting, if we can do this like  `import { object.keys  as keys } from '../utils';`

mosinve: Guess not...

lib/mixins/index.js
```diff
@@ -1,11 +1,21 @@
-import clickOutMixin from './clickout';
-import dropdownMixin from './dropdown';
-import formMixin from './form';
-import formCheckBoxMixin from './form-checkbox';
-import formCustomMixin from './form-custom';
-import formOptionsMixin from './form-options';
-import linkMixin from './link';
-import listenOnRootMixin from './listen-on-root';
-import popoverMixin from './popover';
+import clickOutMixin from "./clickout"
```
mosinve: At upper files there were single quotes ;)

lib/components/table.vue
tmorehouse: What about nested destructuring:

```js
import { object: { keys} } from '../utils';
```

tmorehouse: Although I think you can only do one nested destructure (but as many sub properties as you want)

alexsasharegan: Es2015 imports are definitely not the same as object destructuring. I've been down that road. It stinks!

alexsasharegan: @tmorehouse, the destructuring is just the named export access syntax. Looks like a duck, walks like a duck, but it's a module.

lib/mixins/index.js
alexsasharegan: My darn editor tools! Like I said before, if I can configure the tool to format for me, I'm good. I guess prettier orders double quotes.

lib/components/table.vue
tmorehouse: Booo

alexsasharegan: Couldn't agree more 😞

lib/mixins/index.js
mosinve: i'll add rules for double quotes at eslint config :)

alexsasharegan:
Any last thoughts on this? I'd love to see this merged soon so I can pull in the changes and start using them in the FC PR.

tmorehouse:
I have no objections to merge.

Shall I press the button? :)

alexsasharegan:
I say go for it! I just never want to merge my own large changes. Keep it democratic.

tmorehouse:
We can always revert if there is an issue

Issue #624, opened 2017-07-03
HTML5 Microdata API support


pimlie:
As I could not find any references yet, are there any plans to support the [HTML5 Microdata API](http://www.w3.org/TR/microdata/)? It would be nice if we could say something like `<b-breadcrump :items="items" microdata/>` to add all relevant itemprop's etc. 

One possible complication could be that html markup should be different in some cases. E.g. if you look at the [ListItem](https://schema.org/ListItem#itemlist-1) example you would at least need an extra `<span>` within the `<a>`. So if you want to support this maybe it would be better to automatically generate an `json-ld` object, but this could introduce a slight overhead because ids/urls need to be repeated.

tmorehouse:
IF you use the attribute variant of microdata, you can provide those attributes directly on the component, as the attributes would appear on the root element rendered by the components, such as:
- itemscope 
- itemtype="http://schema.org/Person"
- itemprop="name"

etc.

pimlie:
Unfortunately that is only a partial solution, the full example for a breadcrump looks as follow:
```html
<ol itemscope itemtype="http://schema.org/BreadcrumbList">
  <li itemprop="itemListElement" itemscope
      itemtype="http://schema.org/ListItem">
    <a itemprop="item" href="https://example.com/dresses">
    <span itemprop="name">Dresses</span></a>
    <meta itemprop="position" content="1" />
  </li>
  <li itemprop="itemListElement" itemscope
      itemtype="http://schema.org/ListItem">
    <a itemprop="item" href="https://example.com/dresses/real">
    <span itemprop="name">Real Dresses</span></a>
    <meta itemprop="position" content="2" />
  </li>
</ol>
```
So all the items within the breadcrump should also have the correct itemprop's and it seems you can't access those at the moment.

For a full overview, the corresponding `json-ld` looks as follows:
```html
<script type="application/ld+json">
{
 "@context": "http://schema.org",
 "@type": "BreadcrumbList",
 "itemListElement":
 [
  {
   "@type": "ListItem",
   "position": 1,
   "item":
   {
    "@id": "https://example.com/dresses",
    "name": "Dresses"
    }
  },
  {
   "@type": "ListItem",
  "position": 2,
  "item":
   {
     "@id": "https://example.com/dresses/real",
     "name": "Real Dresses"
   }
  }
 ]
}
</script>
```

tmorehouse:
Ah,  you need to be able to add microdata to the inner elements of a component.

[...]

PR #674, opened 2017-07-08
feat(microdata): add html5 microdata 

pimlie:
This is a possible solution for #624. I believe using the `json+ld` variant instead of adding the itemscope attributes is preferable as its less intrusive. 

That said, I was looking at other components who could benefit from microdata but I couldnt find any actually. And as the `json+ld` variant can also be easily implemented by the users themselves, maybe bootstrap-vue does not need to support microdata after all?

### DIFF
--- a/lib/components/breadcrumb.vue
+++ b/lib/components/breadcrumb.vue
@@ -12,13 +12,17 @@
                     v-html="normalizedItem.text"></b-link>
         </li>
         <slot></slot>
+
+        <script v-if="addMicrodata" type="application/ld+json">
+          {{ microdata }}
+        </script>
     </ol>
 </template>
 
 <script>
 import bLink from './link.vue';
 import { props as linkProps } from '../mixins/link';
-import {arrayIncludes} from '../utils';
+import arrayIncludes from '../utils/arrayIncludes';
 
 const bLinkPropKeys = Object.keys(linkProps);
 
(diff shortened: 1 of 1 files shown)


tmorehouse:
What about a generic component that generates a json-ld script tag, which takes an object, or arrays of objects? Which could be used anywhere in the document, rather than making a niche use case on breadcrumb?

pimlie:
Sounds allright with me, _if_ that component can do a bit more then just printing out a json-ld script tag. E.g. see [jsonld](https://github.com/digitalbazaar/jsonld.js) for a nodejs lib that supports converting to different formats etc. Also supporting [signatures](https://github.com/digitalbazaar/jsonld-signatures) would be a nice feature.

Maybe at a later stage we could add integration support in other components like the breadcrump or e.g. a blogpost through a mixin? Those components could define a context which together with the current document results in a valid json+ld doc.

Also see the examples [here](https://json-ld.org/playground/)

tmorehouse:
Another option, since the `b-breadcrumb` component is basically a short'n'sweet generator, would be to simply create your own breadcrumb markup with all the required micro-data on the li's (and content, possibly using the meta tag method).

alexsasharegan:
You can also take a look at the way I've broken out breadcrumbs in the PR branch for functional components as the component may change in v1.0.0: https://github.com/alexsasharegan/bootstrap-vue/tree/feat/functional-components/lib/components

I would like to break up our components like this as much as possible since functional components are cheap. Even non-stateless components like `table` might be able to use functional components for the `thead`, `tbody`, and `tfoot` plus row components.

tmorehouse:
Closing this PR branch as it is stale ad may better be suited to being created as a directive which can be passed an object for configuration/sub-element targetting.

PR #675, opened 2017-07-08
feat(dropdown): add support for hover trigger

pimlie:
As mentioned in my comment in #417 I would like to contradict the decision of the bootstrap team of not supporting hover events for opening dropdowns. This commit adds support for that.

I have extracted the trigger logic from the `popover.js` mixin and put it in a new `triggers.js` mixin. Now if one should need to support triggers it should be as easy as importing the triggers mixin and:
- setting a `triggerName` in data
- add `hideComponent` and `showComponent` methods
- call the triggersCreated/triggersMounted and triggersBeforeDestroy methods

I have updated `dropdown` and `nav-item-dropdown` to support these new triggers. 
Also I updated the tooltip component with a triggerName so its triggering `shown::tooltip` events instead of `shown::popover`.

Couple of notable changes I had to implement (also good for the docs probably):
- `dropdown.js` is using two trigger elements, the button and the menu. The menu is required as it is absolutely positioned thus the button is not an `offsetParent` of the menu which means events are not propagated from the menu to the button
- `dropdown.js` lists a default debounce of 0. When you use a debounce value here you will probably be missing e.g. `mouseenter` events (on the menu) because you just triggered a `mouseleave` (on the button).  (the current debounce implementation I took from `popover.js` is not per event type, if we would change that that would probably be of some help)
- `dropdown.js` lists a default hide delay of 150ms, this is to overcome the default margin that exists on the dropdown-menu. 

If you enable both the `click` and `hover` triggers for a dropdown the menu opens on hover, and if you click stays open (`persistentState`) until you either clickout, click the menu or click the button again


tmorehouse:
It would have to be definitely optional, as hover makes menus almost impossible to use on mobile devices (the menus wrong open or get stuck),  providing a horrible user experience. This is part of the reason bootstrap official has stoped using hover to open menus. And most native operating system menus and html select elements do not exhibit this behaviour (I.e. opening on hover), as it would contradict the user experience, especially for users of assistive technologies (either blind, or keyboard only users not being able to know how they are supposed to open the menu).

In my personal opinion, and having experience with a userbase that uses assistive technologies, Hover is horrible as an interface to trigger interaction, or anything other than a decorative effect or triggering tooltips (as tooltips normally trigger screen readers to read the content, and users dont need to focus into the tooltip).

I would recommend this not be merged into the base/core components of bootstrap-vue, and would recommend creating a custom dropdown  component in the component-extras directory which suits your use case (Just remember you are going to isolate out some mobile users and users of assistive technologies who would be using your application)

tmorehouse:
I would recommend creating a custom component for this feature.

As our core compo nets should follow the bootstrap v4 standards.

pimlie:
Although I understand your reasons, it is a bit disappointing that I asked in #417 if you (in general, not personal) would accept a PR for this and I got only one 'I don't mind' reply...

That said, the implementation in this PR _is_ completely optional and configurable. Default behaviour is still a click-only trigger and you need to explicitly add hover to the triggers to change the behaviour. Also personally I would only use the combination of the 'hover' and 'click' triggers, not just the 'hover'. Combining those two triggers negate a lot of the drawbacks you mentioned.

tmorehouse:
Now that we have implemented the components-extra directory for non standard components, I would suggest creating this variant of dropdown there, as this pr does change functionality, and introduces changes to the default operation (timings, debkuces, etc).

Create  PR that creates a new component in components-extra, and not an alteration to the default dropdown components.

One never really knows how annoying hover as menu triggers are until you work with non sighted people and see their frustration with the app developers.  Not to mention the sighted users on mobile complaining even more (as hover isn't always available on mobile, or has annoying inconsistencies across mobile browsers).  

Bootstrap is "Mobile First" mantra, and that is why they wenth through a thourough review of this and have refused to enable hover as a trigger for dropdowns.

pimlie:
Thanks for your comments, I think I partially misunderstood. If I would change this PR to have only the trigger logic extracted from `popover.js` and then have `popover.js` use the `trigger.js` and rename the triggers for tooltip to `shown::tooltip` instead of popover, would you accept it then? So that means I will revert the changes to `dropdown.js`, `dropdown` and `nav-item-dropdown`.

At least we have the separate logic then that we can easily re-use for components-extra.

tmorehouse:
I would suggest you creating the altered components in components-extra (using different component names than the standard defsult ones, So they don't conflict).

Yes, optimizing the pop over trigger stuff is good, as long as it doesn't affect current functionality (except using the better event name you suggest).

PR #689, opened 2017-07-13
[WIP] feat(extra-components): new b-form-tinymce component

tmorehouse:
A new extra component that creates a tinymce based rich text form control input.

Can be used as a direct drop in for `b-form-input` (featuring rich text), or can also be used as an "on-demand" editor for "in-place" editing (by using an external trigger to turn on the editor) with optional save and cancel button support (using the standard tinymce `save` plugin and its associated two buttons).

In place mode (set by controlling the prop `disabled`, which renders the content in a div element when disabled), when the user clicks save or cancel, events are emitted, and the editor is hidden (disabled) and the regular div with content is shown. This hiding of the editor can be controlled by calling methods passed on the event object to the handler (i.e `cancel`, `revert`, `preventHide`, etc).

Also takes advantage of the new (once again) `.sync` modifier for props to update the `disabled` prop.

Designed for tinymce version 4.6.4 or better.


pi0:
@tmorehouse Can we retarget it against dev (1.x) ?

tmorehouse:
Sure.  Should be able to, or I can just copy the code to a new PR, since it didn't touch any other components

pi0:
@tmorehouse Sorry I had to ask on slack before this branch swap. Do what's the best you think.

tmorehouse:
This extra component is compatible with both v4.alpha and v4.beta, do doesn't really mater where it is targeted ;-)

pi0:
Closing due inactivity.

PR #700, opened 2017-07-16
feat(listen-on-root): Use a constant for private property name

tmorehouse:
Minor adjustment to use a constant to define the private property name used for storing the root listeners array.


tmorehouse:
Also improves minification.

PR #706, opened 2017-07-17
feat(modal) Make enforceFocus configurable

dedene:
In our project, we needed to launch a secondary modal with some form inputs from a Bootstrap-Vue Modal. Due to the "enforceFocus", this was not possible.

If we make this feature configurable, we can use the <b-modal> also when used with a secondary form modal launched from within the first modal.

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -197,6 +197,10 @@
                 type: Boolean,
                 default: false
             },
+            noEnforceFocus: {
+                type: Boolean,
+                default: false
+            },
             hideHeader: {
                 type: Boolean,
                 default: false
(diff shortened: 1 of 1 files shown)


tmorehouse:
Side note: multiple modals open at once are not supported in Bootstrap V4.  The first modal must be closed before the second modal opens.

Once caveat with disabling enforce focus is that non-sighted users (and keyboard-only users) may not be able to find the modal if tab focus leaves the modal content area. 

Native Bootstrap V4 alpha enforces focus on the active modal (i.. tabbing will cycle through the content of the modal), and is not configurable.

dedene:
I absolutely follow your comments about non-sighted users. 

Our use case was i.e. when using a WYSIWYG editor inside a modal, which opens another (non-bootstrap) modal for editing an image or link. I thought it might be convenient for others to also have the options to turn off the "enforceFocus" if necessary.

tmorehouse:
Ah... OK I see your use case.

So you could turn off enforce focus on the fly, and then re-enable it when the WYSIWYG editor modal closes.

tmorehouse:
I will need to update the documentation for this new prop once this is merged.

dedene:
👍 Awesome, thanks a lot!

Issue #705, opened 2017-07-17
Sticky Navbar - Nav Item Dropdown scrolls to top


SRichner:
When using a sticky navbar in combination with a large (scrollable) content, clicking on a nav-item-dropdown will cause the page to scroll to the top.
 
I have the following code:

```html
<div id="app">
  <b-navbar toggleable variant="success" type="light" sticky>
    <b-nav-toggle target="nav_collapse"></b-nav-toggle>
    <b-collapse is-nav id="nav_collapse">
      <b-nav is-nav-bar class="ml-auto">
        <b-nav-item-dropdown right>
          <template slot="button-content">
            <span>Foo</span>
          </template>
          <b-dropdown-item>Bar</b-dropdown-item>
        </b-nav-item-dropdown>
      </b-nav>
    </b-collapse>
  </b-navbar>
</div>
```

Whenever any Nav Item is clicked (here Foo), it scrolls to the top of the page.
I recreated the issue [in this JSFiddle](https://jsfiddle.net/7fwzwmgv/1/) and the behaviour is the same.

Not sure whether I'm using it the wrong way or it really is a bug and have not yet found a fix/workaround. Any help would be appreciated :)

tmorehouse:
This is probably because the nav items are rendered as links with an href of `#` (which the default action of `href="#"` causes a page to scroll to the top).

If you nav-items are to be actions, then add `@click.stop` handlers on them (if they are to run a method), or set the href prop to a page URL.

tmorehouse:
Also, make sure you load your CSS files first, then vue, then tether and then bootstrap-vue

SRichner:
@tmorehouse Thanks for your reply!

That makes sense, but what about the `button-content` slot, which should simply be a toggle for the actual dropdown menu? 
With a [modified example](https://jsfiddle.net/7fwzwmgv/3/) from the documentation (using `text="Lang"` and not the slot), when clicking the "Lang" navigation item, the dropdown is being toggled and the page scrolls to top.

Is there a way to prevent that slot from being rendered as a link?

tmorehouse:
Ah, let me check the code for nav-dropdown-item, to make sure it is issuing a `preventDevault()` on the click event (it should be but might have slipped through).

tmorehouse:
@SRichner Which version of the library are you using? Latest v0.18.0?

[...]

PR #709, opened 2017-07-18
fix(nav-item-dropdown): Prevent click from scrolling page (Issue #705)

tmorehouse:
Prevent click of trigger button of nav-item-dropdown from scrolling page to top.

Addresses Issue #705


tmorehouse:
Note: This fix may not be needed....

PR #710, opened 2017-07-18
feat(navbar-brand): New component <b-navbar-brand>

tmorehouse:
New component for rendering navbar brand content.

Setting props `to` or `href` will render a `b-link`. 

If both `to` and `href` are not present, then the navbar-brand will be resndered as a `div` by default (configurable via the `tag` prop)

### DIFF
--- a/lib/components/navbar-brand.vue
+++ b/lib/components/navbar-brand.vue
@@ -19,7 +19,7 @@ import { assign } from '../utils/object';
 // less the 'href', 'to', and 'tag' props
 // that we will reconstruct without any defaults
 // so our component functions properly
-const linkProps = assign(omitLinkProps('href', 'to'), {
+const linkProps = assign(omitLinkProps('href', 'to', 'tag'), {
     href: { type: originalLinkProps.href.type },
     to: { type: originalLinkProps.to.type },
     tag: { type: String, default: 'div' }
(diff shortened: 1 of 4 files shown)


lib/components/navbar-brand.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+     <b-link v-if="isLink"
+             class="navbar-brand"
+             v-bind="linkProps"
+             @click="$emit('click', $event)">
+        <slot></slot>
+    </b-link>
+    <div v-else :is="tag" class="navbar-brand">
+        <slot></slot>
+    </div>
```
alexsasharegan: Might make slightly more sense just using the `<component>` tag here. Just shows it off clearly as a dynamic tag.

lib/components/navbar-brand.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+     <b-link v-if="isLink"
+             class="navbar-brand"
+             v-bind="linkProps"
+             @click="$emit('click', $event)">
```
alexsasharegan: Once we hit v2.4 as a peer dependency with Vue, we can start using `v-bind="$attrs" v-on="$listeners"` to proxy everything off to link or whatever. And that is glorious!

lib/components/navbar-brand.vue
tmorehouse: Yeah, was thinking that

lib/components/navbar-brand.vue
tmorehouse: Excellent. Will be nice when that happens

PR #715, opened 2017-07-20
feat(button): Add pressed prop to place button in active state

tmorehouse:
Adds a new prop `pressed`, which defaults to null.

- If `pressed === true` then the `.active` class is added to the button and sets `aria-pressed="true"` attribute.
- If `pressed === false` then  the attribute `aria-pressed="false"` is set.
- If `pressed === null` (the default, or any other value) then neither the class `.active` nor the attribute `aria-pressed` are set.

Useful when creating toggleable buttons.

The pressed state can be synced by setting the `.sync` modifier on the prop:

```html
<template>
  <b-btn :pressed.sync="myPressedState">Toggle Me</b-btn>
</template>
<script>
export default {
  data: {
    myPressedState: false
  }
}
</script>
```

### DIFF
--- a/docs/components/button/README.md
+++ b/docs/components/button/README.md
@@ -5,12 +5,12 @@
 
 ```html
 <div class="row">
-<template v-for="variant in ['primary','secondary','success','outline-success','warning','danger','link']">
-    <div class="col-md-4 pb-2" v-for="size in ['sm','','lg']">
-    <b-button :size="size" :variant="variant" href="">
-        {{variant}} {{size}}
+<template v-for="var in ['primary','secondary','success','outline-success','warning','danger','link']">
+  <div class="col-md-4 pb-2" v-for="size in ['sm','','lg']">
+    <b-button :size="size" :variant="var">
+      {{variant}} {{size}}
     </b-button>
-    </div>
+  </div>
 </template>
 </div>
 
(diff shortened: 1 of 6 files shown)


tmorehouse:
One feature I need to add is set the `focus` class when the button is clicked (focused) and remove the class on blur.  This `focus` class would only occur if pressed is either `true` or `false`, but not ``null`.

This would be done by adding a conditional `focus` and `blur` listeners on the button element

mosinve:
About `focus` class, isn't theres already applies pseudo-class :focus on focus... and it goes away on click at another clickable element. it done by browser.

mosinve:
with this button, there's left a little step to another custom component - switch

tmorehouse:
I'm not sure why BS V4 adds the class. They don't document it, but if you inspect the button in pressed state they add 'focus' class when it has focus. maybe it is a just-in-case if the button is rendered as a link or label?

alexsasharegan:
This is looking good so far. I just pulled and ran the docs site to check it out. Maybe helpful to see two btn variants side-by-side to see the pressed stuff in action. Also, maybe a btn-group example?

tmorehouse:
Yeah, I have a toggle style, and a pressed style in the docs.  but could easily add a non-pressed one for comparison.

Not sure what CircleCI is complaining about.  Might need a cache clear.

tmorehouse:
Added in example of non-pressed button beside a pressed button.

Also added in short example for disabled state.

I suppose we could dd in an example of button rendered as a link (with prop href set)

tmorehouse:
Added a couple more examples (regular button and  button rendered as a link)

tmorehouse:
Removed the focus handler from within the component (negating the need to maintain hasFocus state) by creating a global handler for all instances of button which have the toggle/pressed option.

tmorehouse:
@alexsasharegan I have added in a button group example for toggle buttons

mosinve:
Maybe to add example with v-for generator of buttons?

**HTML**
```
<b-button-group>
    <b-button v-for="(button, index) in buttons" :pressed.sync="button.state" :variant="button.variant">
        {{button.caption}}
    </b-button>
</b-button-group>
{{buttons}}
```

**JS**

```
data: {
    buttons: [
      {
        variant: 'primary',
        caption: 'Toggle 1',
        state: null
      },
	  {
        variant: 'primary',
        caption: 'Toggle 2',
        state: false
      },
	  {
        variant: 'primary',
        caption: 'Toggle 3',
        state: false
      },
	  {
        variant: 'primary',
        caption: 'Toggle 4',
        state: false
      },
	  {
        variant: 'primary',
        caption: 'Toggle 5',
        state: false
      }
    ]
}
```

https://jsfiddle.net/cihkem/fja8kvdr/

tmorehouse:
@mosinve good idea.  I've update the button-group example.

docs/components/button/README.md
```diff
@@ -51,15 +61,86 @@ default padding and size of a button.
...
+  export default {
+    data: {
+      myToggle: false,
+      buttons: [
+        { variant: 'primary', caption: 'Toggle 1', state: true },
+        { variant: 'danger', caption: 'Toggle 2', state: false },
+        { variant: 'warning', caption: 'Toggle 3', state: true },
+        { variant: 'success', caption: 'Toggle 4', state: false },
+        { variant: 'outline-success', caption: 'Toggle 5', state: false },
+        { variant: 'outline-primary', caption: 'Toggle 6', state: false }
```
mosinve: One of them could have state = null to show that in this case state is not changing

tmorehouse: True...  although the name on the button should be `"can't toggle me"` instead of `"toggle #"` so people don't get confused.

mosinve: Yep 👍

tmorehouse: I changed the caption to "No Toggle" for this button :smiley:

mosinve: It's almost like "Can't toggle me", so nevermind 😃

tmorehouse: heheh!

Issue #716, opened 2017-07-20
prevent dropdown first item highlight after open


More4ever:
So, is it posible to prevent highlight first dropdown item after open?

tmorehouse:
For ARIA accessibility best practices, the first item in a dropdown should be focused when the dropdown is opened.  The current implementation emulated the functionality of a native select.  There is a change in the current master that ensures only one item has the highlight when the dropdown is opened.

We are currently re-working the focused state, where as previously the focus/hover state used the browser native outline (a blue box), which many people didn't like the look of, sowe went with making the drop down work more like native select element and native desktop menus.

More4ever:
I think such behavior would be useful in dropdown-select component. But for default dropdown I think it would be useful and more flexibile to have ability not highlight first item

tmorehouse:
Highlighting is part of the focusing (first item in dropdown is always focused for accessibility reasons).

Is it the look of the highlighting that you do not like?  We are addressing the "look" of the highlight, but will not remove the focusing of the first element (as is required by the W3 ARIA specifications for dropdown menus)

More4ever:
I don't expect this behavior, as it doesn't exist in twitter bootstrap dropdown. I can rewrite styles for "active" class, but if component API will give me ability to chose, it would be great

More4ever:
I can create pull request if it will be useful

tmorehouse:
Bootstrap V4 will be moving to using keyboard navigation for Dropdowns (to be AIA compliant) in beta release and will exhibit this same behavior (it is in there current mater code from dropdown).

The main thing we will be changing is the look of the highlight/focus to be that of hover, which is more subtle than the current blue box.

More4ever:
Ok, thanks a lot for your explanations

tmorehouse:
PR #717 should address the styling of the focused element.  The first element is still focused automatically (as per ARIA specs), but is styled different  than the `active` state.

zatziky:
Hi @tmorehouse ,

maybe I didn't understand quite well but Bootstrap doesn't have any item highligted upon opening a dropdown. Yet after hitting an arrow you can select the state. On item hover, the item item is highlighted but **without** the blue outline. Checkout the [docs](https://getbootstrap.com/docs/4.0/components/dropdowns/). 

The current behavior with the pre-selected blue outlined 1st item is really ugly and **not** the intended bootstrap way.

**After dropdown opens**

[...]

PR #717, opened 2017-07-20
fix(dropdowns): Change how dropdown items are highlighted

tmorehouse:
Changes the highlighting method of focused items (instead of using `active` class state).

Slight tweaks to the `:hover` and `:focus` states CSS to remove ugly blue-line outline on focus and add slightly darker background color (default BS V4 contrast ratio is waaay too low to notice the hover/focus state)

### DIFF
--- a/lib/components/dropdown.vue
+++ b/lib/components/dropdown.vue
@@ -45,7 +45,8 @@
 
 <style scoped>
     .dropdown-item:focus,
-    .dropdown-item:hover {
+    .dropdown-item:hover,
+    .dropdown-header:focus {
         background-color: #eaeaea;
         outline: none;
     }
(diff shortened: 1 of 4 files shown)


tmorehouse:
Preview of the new styling (hover, focus, and keyboard selection will appear in this style):
image

And when link is "active" (from clicking or navigation) it looks as so:
image

tmorehouse:
Addresses some concerns of #716 and other discussions with the @dev team

tmorehouse:
Depending on CSS loader, might need to add `!important` on the background color.

tmorehouse:
THe only issue that may crop up is that the `active` state takes precedence over hover/focus, so the focus state will not be as obvious due to the lack of outline.

mosinve:
Looks nice, better that before ) and there's no hardcoded '.active' class 👍

PR #724, opened 2017-07-22
feat(readme): add david dep badge

alexsasharegan:
Adds a badge that shows the status of our dependencies. Helps us stay honest and up to date with our small number of dependencies. _(far right badge)_

image


tmorehouse:
I wonder if these should be included in the documentation home page (readme) https://bootstrap-vue.js.org/

Or sing they are more related to the current master, maybe not....

PR #726, opened 2017-07-23
feat(Event): standard evt obj emulates native Event

alexsasharegan:
Stripped down version of browser `Event` object for use in component events that offer some kind of `preventDefault` behavior implemented by BootstrapVue.

- Adds a `BvEvent` class.

## Example usage

```js
// Inside modal.vue
{
  methods: {
    handleOkayClick(e) {
      this.$emit("ok", new BvEvent("ok", {
        nativeEvent: e,
        target: e.target,
        vueTarget: this,
      }))
    }
  }
}

// Listener
function handleOkay(e) {
  if (someCondition) {
    // Stops modal from closing
    e.preventDefault()
  }
}
```

### DIFF
--- a/lib/utils/BvEvent.js
+++ b/lib/utils/BvEvent.js
@@ -23,7 +23,9 @@ export class BvEvent {
         let defaultPrevented = false;
         // Recreate preventDefault method. One way setter.
         this.preventDefault = function preventDefault() {
-            defaultPrevented = true;
+            if (this.cancelable) {
+                defaultPrevented = true;
+            }
         };
         // Create 'defaultPrevented' publicly accessible prop
         // that can only be altered by the preventDefault method.
--- a/lib/utils/BvEvent.js
+++ b/lib/utils/BvEvent.js
@@ -8,11 +8,10 @@ export class BvEvent {
                 `Failed to construct '${this.constructor.name}'. 1 argument required, ${arguments.length} given.`
             );
         }
-        // Assign defaults first,
-        // the eventInit,
-        // and last the type and timestamp so they can't be overwritten.
+        // Assign defaults first, the eventInit,
+        // and the type last so it can't be overwritten.
         assign(this, BvEvent.defaults(), eventInit, { type });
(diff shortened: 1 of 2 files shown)


tmorehouse:
I'll be taking advantage of this in my b-form-tinyme control :smile:

alexsasharegan:
**Note:** I removed the `e.cancel` alias. I originally placed it in there because that's the modal event obj signature, but I think it's better if that alias is implemented in `modal.vue`. Then it can be deprecated for `e.preventDefault`.

tmorehouse:
Yep, I agree.  The `e.cancel()` is specific to modal.

Once I get the modal transition stuff finished, I'll update the pseudo event to use `BvEvent`

lib/utils/BvEvent.js
```diff
@@ -23,7 +23,9 @@ export class BvEvent {
         let defaultPrevented = false;
         // Recreate preventDefault method. One way setter.
         this.preventDefault = function preventDefault() {
-            defaultPrevented = true;
+            if (this.cancelable) {
```
tmorehouse: :+1:

I wonder if `cancelable` should default to `false`?

alexsasharegan: Possibly. I figured we would be using this mostly in circumstances where we need the `preventDefault` capabilities, so those are inherently cancelable UI events. Then it would be easy to forget to pass `cancelable: true`. Not sure until we use it I guess.

tmorehouse: Yeah, true, 99% of the time this is used it would be for a `preventDefault()` use case.

mosinve: However, i think it would be right to have boolean props switchers initialized as false.

tmorehouse: Although these are properties of the event object, not the component, so true is not as bad.

alexsasharegan: Right, in components were after that `<component bool-prop />` ease of use.

PR #728, opened 2017-07-23
feat(form-radio): Support button style radios

tmorehouse:
Adds new Boolean prop `buttons` (defaults to `false`) that wil enable the radios to be rendered with the look of buttons.
http://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons

Currently the buttons can only be rendered with the same variant specified by the prop `button-variant`.  Different variants for each radio button ar currently not supported (not without altering the `form-options.js` mixin).

`stacked` button radios are supported,, but `state` is not supported for button radios

### DIFF
--- a/docs/components/form-radio/README.md
+++ b/docs/components/form-radio/README.md
@@ -32,34 +32,93 @@ semantic and accessible markup, so it is a solid replacement for the default rad
 export default {
   data: {
     selected: 'first',
-    options: [{
-      text: 'Toggle this custom radio',
-      value: 'first'
-    }, {
-      text: 'Or toggle this other custom radio',
-      value: 'second'
-    }, {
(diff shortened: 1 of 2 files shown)


docs/components/form-radio/README.md
```diff
@@ -50,7 +50,7 @@ Please see options in [`<b-form-select>`](./form-select) docs for details on pas
 options (value array) to `<b-form-radio>`
 
 ### Size
-Constrol the size of the radio text by setting the prop `size` to either `sm` for small or
+Control the size of the radio text by setting the prop `size` to either `sm` for small or
```
tmorehouse: :+1:

tmorehouse:
Example was missing a comma, so not it is working.

Issue #619, opened 2017-07-02
[form-select] Does not work with "multiple" option


tmorehouse:
`b-form-select` does not work with multiple selections,

The `custom` control option might need to be disabled in order to displaya multi-select correctly, and a size/height may also need to be set to increase the height of the multi-selectbox.

Return value should be an array of objects (or values) when multiple is set to true.

tmorehouse:
native select also uses a size attribute for specifying how many lines to show.  I think we would have to make a new property select-size to act as this attribute

tmorehouse:
Currently custom-select doesn't support multiple:
https://github.com/twbs/bootstrap/issues/22065

PR #731, opened 2017-07-24
feat(form-select): Add multiple select support (issue #619)

tmorehouse:
Adds multiple select support to form-select (addresses issue #619)

new props:
- `multiple` enabled multiple select support when set to `true`
- `multiple-size` set the number of rows displayed in multiple select. Defaults to browser default size

`form-options.js` mixin updated to support multiple select.

Updated documentation

### DIFF
--- a/docs/components/form-select/README.md
+++ b/docs/components/form-select/README.md
@@ -1,46 +1,7 @@
 # Form Select
 
-> Bootstrap custom `<select>` using custom styles. Provide options based on an array, array of objects. or an object.
-
-```html
-<template>
-  <div>
-    <b-form-select v-model="selected" :options="options" class="mb-3">
-    </b-form-select>
-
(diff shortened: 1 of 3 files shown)


mosinve:
About `multiple-size` prop, why don't use common known html prop `size`, it will be easier to move from pure HTML to BV components.

tmorehouse:
Problem is, it shouldn't be applied to regular select (non multiple), and we already have a `size` prop that controls the text size of form-controls (`size="lg"` and `size="sm"`), so it would get confusing with the two.  this way it is more explicit as to what it is applied to.

In reality, the select `size` attribute really should have been called `rows` in native HTML (like textarea has `rows`).  Text inputs also have a native `size` attribute, which conflicts with the bootstrap form-control sizes (`lg` and `sm`), but since form-controls have the `.form-control` class, the `size` prop on text inputs isn't needed (as they will always be 100% of the width of their container).

PR #721, opened 2017-07-21
fix(alert): use v-model to update show value

alexsasharegan:
Fix for issue #718 

My initial work on the idea of using v-model. I haven't run my bug test on this yet because things are really busy for me. 
**Edit:** This does clear up my bug.

**Note:** ~~This may need to be tweaked as doing `<b-alert show dismissible>` doesn't allow the alert to be dismissed. The addition of `v-model` becomes necessary. It might be necessary to have an initial show either as a prop or as a data value local to alert. But if you think about it, it's fundamentally flawed to declare both `show` and `dismissible` at the same time while also making alerts reusable.~~

Feel free to contribute on this branch.

Issue #718, opened 2017-07-20
[b-alert] Dismissible alerts cannot be resused


alexsasharegan:
So I'm trying so use a single alert component in a parent vue component. Children pass up messages, and the parent resets the content and `show` value on the alert. I can't get the alert to re-show because the `show` value in the parent isn't updated when using the countdown. I try updating the show value with the dismiss countdown event, but that doesn't work after I dismiss an alert.

https://jsfiddle.net/alexsasharegan/zrdur2cp/

- Try dropping a msg into the input, letting it self-dismiss. All good.
- Now try the other btn. Should be good.
- Now fire a btn and dismiss. Should close.
- Now fire another btn. DOA

mosinve:
`bootstrapVue.prototype.Alert()`

tmorehouse:
@mosinve that is a cool idea!

alexsasharegan:
Weird, it seems to be working now...

mosinve:
it breaks only if

- Try dropping a msg into the input, letting it self-dismiss. All good.
- Now try the other btn. Should be good.
- Now fire a **GREEN** btn and dismiss. Should close. (for X times)
- Now fire a **RED** btn and dismiss. Should close.
- Now fire **GREEN** btn. DOA

alexsasharegan:
I have a suspicion this is related to using a number countdown in conjunction with early dismissal. Also, I don't think I like that the dismiss countdown doesn't accept a `value < 2`. That should be up to the dev to decide that's a silly time.

alexsasharegan:
Using `v-model` here seems to make much more sense. `show` is copied to a closure-scoped variable `dismissCountDown`, but as the countdown is decremented, `show` remains unchanged.

TonBits:
Able to replicate this on the main documentation page using Chrome

https://bootstrap-vue.js.org/docs/components/alert/

Steps to replicate easily
1.) Click on the "Show alert with count-down timer" button
2.) Click the X mark quickly (less than a second after clicking the button above)
3.) Clicking the button again will not display the alert anymore.

This has something to do with the timing with the dismiss countdown timer.

alexsasharegan:
@TonBits If you have the time, have a look at PR #721. Try pulling the branch down, run `yarn docs-dev`, and visit the docs page on [127.0.0.1:3000](http://127.0.0.1:3000). You'll notice my note in the PR. I'm curious how much that change would affect you and other users.

[...]

PR #734, opened 2017-07-24
[WIP] fix(alert): use v-model to update show value

tmorehouse:
Fix for issue #718

Alex's  initial work on the idea of using v-model to address issue #718

See PR #721 for discussion.

### DIFF
--- a/__tests__/components/alert.spec.js
+++ b/__tests__/components/alert.spec.js
@@ -3,7 +3,7 @@ import { loadFixture, testVM, nextTick, setData } from "../helpers";
 const variants = ["success", "info", "warning", "danger"].map(v => {
     return {
         variant: v,
-        class: `alert-${v}`,
+        variantClass: `alert-${v}`,
         ref: `variant_${v}`
     };
 });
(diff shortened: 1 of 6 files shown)


mosinve:
its about this bug report:
1.) Click on the "Show alert with count-down timer" button
2.) Click the X mark quickly (less than a second after clicking the button above)
3.) Clicking the button again will not display the alert anymore.

Problem in that case is that show value doesn't actually changed, and for all next clicks, watcher don't fire

mosinve:
as a solution, we can emit value=false, or completely destroy alert on close, but then we must create it again at open.

tmorehouse:
So it must be because the time is in the middle of being in first SetTimeout, and when value changes (dismissed, counter isn't being cleared in time... or maybe countdown time should check to see if show has been changed and cancel itself?

mosinve:
No, the problem exactly in not resetting value, used for show.

mosinve:
if to wait at least a 1 sec, for this value changed, the all works fine.

tmorehouse:
What if the first countdown happened immediately (i.e. setTimeout iof 0), and then subsequent countdowns happened 100ms later?

mosinve:
on other hand, maybe it not quite correct example...

mosinve:
having v-model show, i think we dont need to emit `dismiss-count-down` as we can return this countdown back by v-model

mosinve:
And i have Vue warning about direct mutate prop `show`

alexsasharegan:
I'll have to look into this deeper. I also see that I left a straggler in the data @ `vm.dissmissed`. It's no longer being used to compute anything.

tmorehouse:
We can probably delete this PR and branch, now that we have a "fix" (hiding the close button on auto-dismissing alerts)

PR #737, opened 2017-07-24
Add components to table heads

happilymarrieddad:
Hey Guys,

I really need a component in the header of a table. Is this something you guys might consider? If this isn't the best way to do it what can I do to get it approved or something like it so I can have custom thead is the table. Basically, there are a few columns in the rows that I need some dropdown buttons selects and what not that can't be accomplished easily. I could always put html in the v-html but it's dirty and Evan You says you shouldn't abuse v-html and you should be using components.

https://github.com/vuejs/vue/issues/1468
https://vuejs.org/v2/api/#is

Thanks!


tmorehouse:
Why not just place your component into the scoped named `HEAD_<field>` slot?

You can place anything in the slot, including components, just like you can with regular defsult slots.

tmorehouse:
And, since the field definition object is passed as one of the scoped variable properties, you can do your own `:is` inside the slot.

tmorehouse:
Check the b-table docs for the section on providing custom formatted header cells,

tmorehouse:
```html
<b-table ..... >
 <slot name="HEAD_actions" scope="fieldData">
   <component :is="fieldData.field.myComponent">
   </component>
  </slot>
  <slot name="HEAD_foo" scope="fieldData">
     <my-compoent>{{ fieldData.label }}</my-component>
  </slot>
  <slot name="HEAD_bar" scope="fieldData">
     <b-select @click.stop :options="..."></b-select>
  </slot>
  <slot name="HEAD_baz" scope="fieldData">
     <strong>{{ fieldData.label }}</strong>
     <small>( {{ fieldData.key }} )</small>
  </slot>
 ...
</b-table>
```

lib/components/table.vue
```diff
@@ -17,7 +17,8 @@
                     :tabindex="field.sortable?'0':null"
                 >
                   <slot :name="'HEAD_'+key" :label="field.label" :column="key" :field="field">
```
tmorehouse: This is a named slot.  you can reference the slot by name to place content in any of the header cells, which will override the default content inside this slot.

See https://vuejs.org/v2/guide/components.html#Scoped-Slots

lib/components/table.vue
```diff
@@ -17,7 +17,8 @@
...
-                    <div v-html="field.label"></div>
```
tmorehouse: This line will be rendered if the user does not provide content for the named slot.

If the user does provide a scoped slot, then its contents (HTML, text, components, etc) will replace this default.

lib/components/table.vue
```diff
@@ -17,7 +17,8 @@
...
+                    <component v-if='field.component' :is='field.component' :field='field' ></component>
```
tmorehouse: If you do this, your component will be very limited (as in impossible) for any `props` you may need to set on it.

Just  place your component inside a `<b-table ...><template slot="HEAD_<filedkey>">...</template></b-table>`

happilymarrieddad:
@tmorehouse Hey man thanks so much for putting up with me. I looked through that page and completely missed it. I also should have looked into the vuejs docs better. Thanks man!!!

PR #739, opened 2017-07-25
feat(table): add field data formatter prop

mosinve:
1. Add new prop to field - formatter, to allow custom formatting via callback function
2. Add api-url prop to pass to callback item provider function
3. Move sortBy and sortDesc props to Props instead Data, to allow define custom sorting on mount
4. Some refactor

TODO: modify docs, examples.


lib/components/table.vue
```diff
@@ -136,12 +135,24 @@
...
                 }
             },
+            sortBy: {
+                type: String,
+                default: null
+            },
+            sortDesc: {
+                type: Boolean,
+                default: true
+            },
```
tmorehouse: You might get prop mutation errors on `sortBy` and `sortDesc` due to the listeners on head click.

You might need local copies of these in data, and then use the new `.sync` modifiers on the props, and emit `updated:sort-by` and `update:sort-desc` events to update the users copy of the props.

tmorehouse: If you like, I can tweak the props/data to handle this.

tmorehouse:
Just thinking you might want to break out the changes into separate PRs... one for the formatter function, and one for the new props for sorting, and one for the api-url

tmorehouse:
@mosinve I've created a separate PR #742 for making  `sort-by` and `sort-desc` available as two-way  props (via `.sync` modifier).

lib/components/table.vue
```diff
@@ -104,25 +111,17 @@
...
-                return -1;
-            } else if (a[sortBy] > b[sortBy]) {
-                return 1;
-            }
-            return 0;
-        } else {
-            return toString(a[sortBy]).localeCompare(toString(b[sortBy]), undefined, {
-                numeric: true
-            });
+            return ((a[sortBy] < b[sortBy]) && -1) || ((a[sortBy] > b[sortBy]) && 1) || 0;
```
tmorehouse: :+1:

tmorehouse:
For your formatter function, you could have it default to a built in function that just returns the field data, and then the user can provide their own formatter if they want to do javascript based formatting.

Basically just pass the formatter function two parameters, the field `key` (or the field object), and the record data object.  The formatter probably only needs the field key and not ht field object, as the user would most likely have a copy of the field objects in their app anyway.

mosinve:
It was supposed as an alternative to slots, so if  it not defined, then slots used.... In case if i make a default builtin func, then could be that slots probably will never work?

tmorehouse:
Hmmm...

The output of the built in function could be passed to the default content of the slots

lib/components/table.vue
```diff
@@ -4,91 +4,98 @@
...
-                    <slot :name="key" :value="item[key]" :item="item" :index="index">{{item[key]}}</slot>
+        <tr v-for="(item,index) in _items"
+            :key="index"
+            :class="rowClass(item)"
+            @click="rowClicked($event,item,index)"
+            @hover="rowHovered($event,item,index)"
+        >
+            <template v-for="(field,key) in fields">
+                <td v-if="hasFormatter(field)" :key="key" :class="tdClass(field, item, key)"
+                    v-html="callFormatter(field, item, key)">
```
tmorehouse: This works well. You probably don't need to send the `field` object (as the user probably has a reference to it in their app.

But maybe you could pass `field` last, as most would probably only need `key` and the `item` record data.

mosinve: Yup, i understand, will remove field param, and check how it will work.

tmorehouse: Although it might be handy to have it still in there for some people, but as the 3rd argument.

mosinve: Yeah, ok, accepted))

lib/components/table.vue
```diff
@@ -524,14 +533,25 @@
...
+                    data.then(items => {
                         this._providerSetLocal(items);
                     });
                 } else {
                     // Provider returned Array data
                     this._providerSetLocal(data);
                 }
+            },
+            hasFormatter(field) {
+                return field.formatter && ((typeof (field.formatter) === 'function') || (typeof (field.formatter) === 'string'));
```
tmorehouse: Ah, so you are allowing each field to have it's own formatter function (via the fields objects)

I originally thought you were going to have a global field Formatter function (a prop), which would return a format based on the key, item, and field passed to it.  

Your solution allows for more flexibility. :)

lib/components/table.vue
```diff
@@ -524,14 +533,25 @@
...
+            },
+            callFormatter(item, key, field) {
+                if (field.formatter && (typeof (field.formatter) === 'function'))
+                    return field.formatter(item[key]);
+
+                if (field.formatter && (typeof (this.$parent[field.formatter]) === 'function'))
+                    return this.$parent[field.formatter](item[key]);
```
tmorehouse: So what do these two lines handle?

lib/components/table.vue
tmorehouse: and what is the option for `typeof (field.formatter) === 'string'` handle?

lib/components/table.vue
tmorehouse: So, just making a guess... does this allow calling a method defined in the parent (when field.formatter is a string)?

lib/components/table.vue
mosinve: I check if supplied value is string or function, to determine if remote formatter is correctly provided.

lib/components/table.vue
mosinve: Yep,  if formatter supplied as `String`, then we search for it in parent's methods, this solution have a potential bug, when we have table component in a some wrapper component, like b-table, but i don't find any more flexible solution.

lib/components/table.vue
mosinve: If provided formatter  func declared in global scope and provided as `Function` to formatter prop, then we directly call it.

lib/components/table.vue
tmorehouse: Could a user do this:

```js
{
  data: {
    fields: {
      foo: { sortble: true, formatter this.fooFormatter },
      bar: { sortble: true, formatter this.barFormatter }
    }
  },
  methods: {
    fooFormatter(item, key, field) {
      return item.a + item.b;
    },
    barFormatter(item, key, field) {
      return item.a - item.b;
    },
 }
}
```

tmorehouse: I think the field formatter calling signature should include `item`, `key`, `field` as args. 

That way a single formatter function could be used to handle multiple columns, based on the `key`.  That way a user could provide either unique formatters, or a single generic formatter method depending on their preference.

mosinve: then we must pass these vars from child component. 
and this declaration won't work, as parent functions aren't in children's scope. in case we want to use parent's methods, we must supply it as String, Function declaration only for global scoped which will be available to child
```
data: {
    fields: {
      foo: { sortble: true, formatter this.fooFormatter },
      bar: { sortble: true, formatter this.barFormatter }
    }
  },
```

mosinve: the main goal of this is to provide a way to modify single field by simple func... don;t like idea of all-in-one heavy variant

mosinve:
The real use case:
```
 const lastPostsFields = {
        id: {
            label: '#',
            sortable: 'id'
        },
        user: {
            label: 'Name',
            sortable: 'name',
            formatter: 'linkUser'
        },
        created_at: {
            label: 'Created at',
            sortable: true,
            class: 'text-center',
            formatter: 'formatDate'
        },
        type: {
            label: 'Type'
        },
        thread: {
            label: 'Title',
            formatter: 'linkThread'
        },
        text: {
            label: 'Text',
            formatter: 'formatPost'
        }
    }
```

lib/components/table.vue
tmorehouse: But making the extra args available would make rendering for "virtual" columns (columns that don't have a value in items) be possible (i.e. a virtual column that is comprised of data from different fields within the item record)

Maybe make the formatter function use this signature:

`formatter(value, item, key)`

And the user can then ignore the extra `item`, `key` args if not needed.

tmorehouse:
How would one use a formatter function for a virtual column?  i.e. a column that `appears` in `fields`but on in the raw `items` data?

tmorehouse:
ie.:
```js
data: (
  fields: {
    name: { sortable: true },
    birthdate: { sortable: true, formatter: 'formatDate')
    age: { formatter: 'calcAge' }
  },
  items: {
    { name: 'bob', birthdate: { year: 1967, month: 12, day: 7 } },
    { name: 'jane', birthdate: { year: 1992, month: 4, day: 20 }  )
  }
},
methods: {
  formatData(value) {
    return `${value.year}/${value.month}/${value.day}`;
  }
  calcAge(value, item) { // value is null/undefined in this case
    const d = item.birthdate;
    const bday = new Date(d.year, d.month-1, d.day);
    const diff = Date.now() - bday.getTime();
    return Math.abs((new Date(diff)).getUTCFullYear() - 1970);
  }
}
```

Without access to the `item` row data, it would be impossible to create a formatter for field `age`

docs/components/table/README.md
```diff
@@ -147,6 +148,7 @@ Supported field properties:
 | `thClass` | String or Array | Class name (or array of class names) to add to header/footer `<th>` cell
 | `tdClass` | String or Array | Class name (or array of class names) to add to data `<td>` cells in the column
 | `thStyle` | Object | JavaScript object representing CSS styles you would like to apply to the table field `<th>`
+| `formatter` | String or Function | A formatter callback function, can be used instead of slots
```
tmorehouse: Can be used instead of slots, but only for fields that have a real `value`.  Currently one cannot use this for creating data for a virtual column.

lib/components/table.vue
mosinve: Then we will have two ways of handling table data: slots and formatter callback.
Callbacks will give more reusability, as we can move formatter functions out to standalne module and use it as mixin at components and even as global mixin to app's Vueinstance, 
while slots will allow to insert another components into table cell and also format field value and create virtual columns in case we don't need to reuse it at whole app.

docs/components/table/README.md
mosinve: Yep, and by design these featues, slots and formatter are mutually exclusive. Ie. if we define formatter at items array, then we cant use slots, and slots are used **only** if formatter is not defined.

tmorehouse: So that should be added to the documentation ;-)

pi0:
Vue 2.4 SSR is incompatible with slots inside templates (or maybe because of v-if/v-else) and rendering raw `<slot>` tags like this:

image

tmorehouse:
Hmmmm....

pi0:
Workaround: dc8d23898d52a839fd3bad9ac1631b00026feb9b 😆  New SSR optimizations for `v-if` are causing many bugs like this.

PR #742, opened 2017-07-25
feat(table): Add syncable sort-by and sort-desc props

tmorehouse:
Adds two new props to allow user to specify sorting column and direction.

The new props can be made two way by adding the `.sync` modifier, and will be updated whenever sorting column or direction changes.

These props complement (or can be used instead of) the `sort-changed` event.

Includes documentation updates

### DIFF
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -7,20 +7,25 @@ custom rendering, events, and asynchronous data.
 <template>
 <div>
   <div class="my-1 row">
-    <div class="col-6">
+    <div class="col-md-6">
       <b-form-fieldset horizontal label="Rows per page" :label-cols="6">
-        <b-form-select :options="pageOptions" v-model="perPage"></b-form-select>
+        <b-form-select :options="pageOptions" v-model="perPage" />
       </b-form-fieldset>
     </div>
-    <div class="col-6">
+    <div class="col-md-6">
       <b-form-fieldset horizontal label="Filter" :label-cols="3">
-        <b-form-input v-model="filter" placeholder="Type to Search"></b-form-input>
+        <b-form-input v-model="filter" placeholder="Type to Search" />
       </b-form-fieldset>
     </div>
   </div>
 
-  <div class="my-1">
-    <b-pagination size="md" align="center" :total-rows="totalRows" :per-page="perPage" v-model="currentPage" />
+  <div class="row my-1">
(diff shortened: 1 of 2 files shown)


docs/components/table/README.md
```diff
@@ -382,12 +383,28 @@ will emit the `filtered` event, passing a single argument which is the complete
 items passing the filter routine. Treat this argument as read-only.
 
 ### Sorting
+As mentioned above in the **fields** section, you can make columns sortable. Clciking on
+sortable a column header will sort the column in ascending direction, while clicking
+on it again will switch the direction or sorting.  Clicking on a non-sortable column
```
mosinve: I removed a reset by clicking on non-sortable column, because it just set sortBy to `null`, but parameter sort=null|asc still sending to backend API.
On another tought, is may be up to Item provider func to deal with null sort target and remove it from query.
And on third thought, maybe better not expicitly to null `sortBy`, but reset it to default sortBy, defined at startup. If it not defined, then to null it.

tmorehouse: Although some may want to clear the sort... but you never know.  The original table (from before I started working on it) cleared the sort on a non-sortable column.

Some people may want to table to display in the order sent from the server (as not having a null sort by would case b-table to sort the data which is received).

mosinve: Yep, i think that this feature should be restored at my variant of <b-table>
But it needs to be reviewed.

mosinve: I think, we can merge this branch, and i will continue to work on mine.

tmorehouse: OK cool... so everything looks OK in this branch?

mosinve:
For a sec, Ill make a final test at local Playground doc, though

docs/components/table/README.md
mosinve: Mmm, Ill make a final test at local Playground doc, though

tmorehouse: OK good... The docs example should show the sort key and order under the table, using the `.sync` modifier.

tmorehouse: Alo, I could dd a button that says 'clear sort' (which is disabled if `sort-by` is `null`) to the example.

Clicking on the button would set `sort-by` to `null`, and if `sort-by` is `null` then the button could be disabled.  Sowing a bit more interactivity with with the table.

Issue #743, opened 2017-07-25
Carousel Navigation - Help


kaankucukx:
Hello ı have been trying to navigate carousel from outside of .carousel-indicatiors.

since in bootstraps code i can make it with defining "data-to-slide" atr. 

https://www.w3schools.com/bootstrap/tryit.asp?filename=trybs_carousel&stacked=h
`      <li data-target="#myCarousel" data-slide-to="0" class="active"></li>
      <li data-target="#myCarousel" data-slide-to="1"></li>
      <li data-target="#myCarousel" data-slide-to="2"></li>
    `

I couldnt make it work for bootstrap-vue. i tried to define all aria labels on extrenal <li aria-*> but no navigation happening.

I have a seperate/secondary navigation on my page apart from carousel but needs to navigate carousel slider too.

This navigation outside of the div of `#carousel` and have to navigate carousel slides as indicators.

I am using bundler, tried vue-webpack-simple and advanced so far for compat. issues bu no help so far.

tmorehouse:
The aria labels are for labelling the controls for sight impaired users that are using a screen reader to describe the page layout and controls.

Currently carousel doesn't provide a `v-model` binding for setting the current slide.

You can get around this and control the next and previous side by setting a `ref="myCarousel"` attribute on `<b-carousel ref="mycarousel">`  nd then calling the `next()` and `previous()` methods on that ref:

```js
// Goto next slide
 this.$refs.myCarousel.next();
// Goto previous slide
 this.$refs.myCarousel.prev();
```

kaankucukx:
@tmorehouse Hello, thank you for ur guidance

Instead of going back and forward i need to slide to exact slide "index(n)". I m digging in to source code but yet i cant find any solution.

To demonstrate what i need,think of a site navigation bar menus, `@click = navigate to index(n) of the slide` would be perfect :) 🥇

tmorehouse:
Try this:

`this.$refs.myCarousel.index = n`

where n is the slide you want to go to.

Note if the carousel is in the middle sliding when you specify the new index, it will not slide to that slide.

mosinve:
is it possible to know current carousel state (i.e. if it in transition state)?

tmorehouse:
There is a private property `.isSliding` which should be treated as read only.

mosinve:
So, one can check it by `this.$refs.myCarousel.isSliding`?

tmorehouse:
Yup.

[...]

PR #744, opened 2017-07-26
feat(carousel): Add v-model support (issue #743)

tmorehouse:
Add v-model support to allow external control of slides.

Addresses issue #743

### DIFF
--- a/lib/components/carousel.vue
+++ b/lib/components/carousel.vue
@@ -66,9 +66,9 @@
                 :aria-label="labelGotoSlide + ' ' + n"
                 :aria-describedby="slides[n-1].id || null"
                 :aria-controls="id ? (id + '__BV_inner_') : null"
-                @click="index=n-1"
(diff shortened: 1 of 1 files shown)


lib/components/carousel.vue
```diff
@@ -240,13 +260,16 @@
                 const currentSlide = this.slides[oldVal];
                 const nextSlide = this.slides[val];
 
+                // Don't do anything if there aren't any slides to slide to
                 if (!currentSlide || !nextSlide) {
```
mosinve: How about a variant of infinite loop carousel?

tmorehouse: Carousel will loop forever...

This basically makes the transitions happen if there isn't any slides (or if there is only one slide).

Not sure why the original author just didn't check if the slide's length === 1.  But there must have been a situation where they had to put this check in.

Issue #554, opened 2017-06-26
[pagination] allow usage as navigation (link support)


tmorehouse:
In some instances, it may be beneficial to use pagination as page navigation (i.e. links) rather than just component pagination, that would be SEO friendly.

Support could be added by adding a new prop for `navigation` and a `base-href` URL.

tmorehouse:
Based on work I have one with the current `<b-pagination>` component, I think a new component called `<b-nav-pagination>` would be best.

PR #746, opened 2017-07-27
feat(pagination-nav): New navigation based pagination (issue #554)

tmorehouse:
Addresses issue #554 

New pagination component for page navigation

Takes a `base-url`, which then the page number is appended to.

Supports generating `<router-link>` rather than regular links by setting the property `use-router`


mosinve:
Don't see a catch of this component.
What if i need to  define a custom array of links?
For ex, a typical vue-router config:
```
const routes = [
  { path: '/foo', component: Foo },
  { path: '/bar', component: Bar }
]
```
Is it possible to setup to navigate into these pages, or it just for navigating over enumerable resources, like pages or items?

tmorehouse:
This is for paginating over enumerable pages (say entries in a blog, or pages of an article/story/etc.

I was thinking about the option of providing an array like you suggested, for things that aren't sequential (i.e. 1, 2, 3, 4, 5, etc). But then wan't sure how to display the index number.

Another option might be to modify the baseUrl so that it can take a template (so the page number could be anywhere), or to provide a lookup table that converted the page number (1, 2, 3, etc) into URLs.

For router support, it is currently pretty basic... if `use-router` is set to true, then rather than an HREF ir places the URL into b-link's `to` prop.

Definitely open to suggestions to make this component flexible.(but not too complex to use).

Maybe if an array is provided (maybe a prop called `page-links`), it would use that, and if not st it would use the `number-of-pages` prop and `base-url`

The other option might be to provide a prop that lets people specify a URL/TO generator function (that overrides the built in makeUrl function), and it would be passed the page number and would return either an `href` or `to` prop.

mosinve:
We can add prop `items` just like for table, which can we either an Array, or a Function, but i think this will make component more complex.

tmorehouse:
We could use the idea of a url-maker function (which defaults to the current internal `makeUrl` function), which people could then either return a string (href) or an object (to).  then they can do their own lookup in their own array or whatever to make the individual links.  It would only be called for the displayed page link buttons (and the prev/next buttons).

tmorehouse:
I've added in a `link-gen` function prop, that allows finer grained control over the `href`/`to`

lib/components/pagination-nav.vue
```diff
@@ -90,7 +90,7 @@
             </li>
             <li v-else class="page-item" role="none presentation">
                 <b-link class="page-link"
-                        v-bind="linkProps(currentPage - 1)"
+                        v-bind="linkProps(numberOfPages)"
```
tmorehouse: Good catch!   a copy and paste error on my part ;)

mosinve: Everything else works fine, linkGen func awsome, but it need detailed docs. That part about using arrays of links and assigning length-1 to total pages, and pagenum-1, especially.

tmorehouse: Yeah,  I'll update the docs with a few examples, and explain that page num is a value from 1 to N, while arrays are 0 to N-1.

tmorehouse: @mosinve I've now added an example using an array of links to generate the page links.

PR #750, opened 2017-07-29
feat(form-input): Add autocomplete prop

paulgv:
It would be nice to be able to control browsers autocomplete features with an `autocomplete` prop on form inputs.


alexsasharegan:
I've been adding `autocomplete` into my `<b-form-input>`'s without any issue. Why does this need to be a prop?

Issue #761, opened 2017-08-01
Form Select "size" prop conflict with native size attr


r4fx:
In `b-form-select` there is no possibility to use native HTML5 `size` attribute, because of prop with the same name, so please consider to change the name for `size` property to something different to enable use of `size` attribute. Thanks!

```html
<select size="3">
  <option value="volvo">Volvo</option>
  <option value="saab">Saab</option>
  <option value="opel">Opel</option>
  <option value="audi">Audi</option>
</select>
```

See how native attr `size` work
https://www.w3schools.com/Tags/tryit.asp?filename=tryhtml_select_size

tmorehouse:
Currently `b-form-select` does not support the `multiple` attribute (neither does Bootstrap V4's custom select) which allows you to take advantage of the `size` attribute.

There is a PR #731 that introduces multiple select support, and also introduces the `multiple-size` property, which will set the `size` attribute on the select (and also removes custom styling from the select).

This PR has been merged. and will be available in v0.19.0 which should be released in the near future.

tmorehouse:
As a side note, the prop `size` is used by many of the BootstrapVue components to control the Bootstrap V4 form-control/button sizes.

r4fx:
yeah I seen this PR, but as you mentioned this will cover only dropdown with multiple mode,
there will not be an option to use `size` attribute in single mode. This is maybe a rare case to use it (as in my case) but i think there should be a possibility to use such native attr.

tmorehouse:
New merged PR #762 should address the issue to allow `select-size` to be set when not in `multiple` mode.

Sobient:
Avoid using W3Schools as a reference.

Here is a proper documentation: https://developer.mozilla.org/en/docs/Web/HTML/Element/select

PR #762, opened 2017-08-01
feat(form-select): Allow selectSize to be set when not in multiple mode (Issue #761)

tmorehouse:
Allows the select native `size` attribute to be set (via the prop `select-size`), even when not in multiple mode.  Only has any effect when `select-size > 1`.  Custom select styling is removed when in `multiple` mode or `select-size > 1`.

Changes `multiple-size` attribute to `select-size` (`multiple-size` was not available in v0.18.0, so deprecation is not needed)

Addresses issue #761

### DIFF
--- a/docs/components/form-select/README.md
+++ b/docs/components/form-select/README.md
@@ -1,7 +1,7 @@
 # Form Select
 
 > Bootstrap custom `<select>` using custom styles. Provide options based on an
-array, array of objects. or an object.
+array, array of objects, or an object.
 
 ### Options
 `options` can be an array or a key-value object. Available fields:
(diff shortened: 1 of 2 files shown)


docs/components/form-select/README.md
```diff
@@ -78,7 +78,7 @@ export default {
 <!-- form-select-1.vue -->
 ```
 
-#### Slect sizing (displayed rows):
+#### Select sizing (displayed rows):
```
tmorehouse: :+1:

lib/components/form-select.vue
```diff
@@ -4,7 +4,7 @@
             :id="id || null"
             v-model="localValue"
             :multiple="multiple || null"
-            :size="(multiple || selectSize > 1) ? multipleSize : null"
+            :size="(multiple || selectSize > 1) ? selectSize : null"
```
tmorehouse: Good catch! thanks!

mosinve: 😉

PR #768, opened 2017-08-03
fix(alert): add colon for binding `aria-label`

ReedSun:
I think this place maybe forget colon to bind attribute `aria-label` . So I make this PR.


tmorehouse:
Good Catch!  Thanks!

PR #776, opened 2017-08-08
feat(table): Add row-dblclicked event

tmorehouse:
Emit a `row-dblclicked` event when a row is double clicked.

Event has same argument signature as `row-clicked`


tmorehouse:
Another feature option I was thinking of adding was using double click on  sortable header field to clear the sort.

PR #777, opened 2017-08-09
feat(readme.md): Update Vuejs badge to 2.4.x

tmorehouse:
Make badge match current VueJS dependency version


pi0:
Hi @tmorehouse :) So fast. I'm on it will push commit now.

PR #786, opened 2017-08-09
Tmorehouse modal docs

tmorehouse:


tmorehouse:
Re-doing again dev branch

pi0:
@tmorehouse I've updated default branch to `dev` if you prefer :))

tmorehouse:
I was also in another branch when I made the first PR... which had other changes in it...  went back to a clean branch and remade the changes.

pi0:
That's a much better idea working on `dev` indeed. Also as a suggestion why don't you make single fixes with direct commits without working on your fork? I think this would save you lots of time :)

tmorehouse:
Using temp branches on this repo, rather than temp branches on my clone.

PR #792, opened 2017-08-10
feat(eslint): update settings to remove editor errors

alexsasharegan:
As stated, just trying to remove all the red squigglies from my editor.


.eslintrc.js
```diff
@@ -1,280 +1,214 @@
...
+        "operator-linebreak": ["error", "after"],
+        "padded-blocks": "off",
+        "prefer-arrow-callback": "off",
+        "prefer-const": "off",
+        "prefer-numeric-literals": "error",
+        "prefer-promise-reject-errors": "error",
+        "prefer-reflect": "off",
+        "prefer-template": "off",
+        "quote-props": "off",
+        quotes: "off",
```
mosinve: If you use doublequotes, maybe to enable this rule?
`quotes: [2, "double", "avoid-escape"]`

.eslintrc.js
```diff
@@ -1,280 +1,214 @@
 module.exports = {
-    'env': {
-        'browser': true,
-        'commonjs': true,
-        'es6': true
+    env: {
+        browser: true,
```
mosinve: Why here not using quotes?  
this breaks uniformity

alexsasharegan: Because Prettier did it. It only escapes illegal object key identifiers. I personally like this, but ultimately, I just go with whatever the tool does for me. Had it maintained consistency, I would have deferred to that. I will say this: prettier is a fairly accurate representation of the majority consensus of js style. In other words: you may not like it, but most people do.

alexsasharegan: My personal reason for appreciating dropping quoting object keys that are legal identifiers: it enables syntax highlighting. For me, string syntax highlighting gives me no indication of what the token is. I like being able to tell it's an object key. Not a big deal to me either way though. I just hit format, and move on. Guilty 😜

mosinve: there's eslint rule about this case - "quote-props"
currently it switched off, my offer to set it to: `"quote-props": [2, "as-needed", { "keywords": true }]`

tmorehouse: I use double quotes always in JSON object files, since quotes preferred over double quotes within source code around strings (unless the string has a lot of escaped single quotes), and literals for keys, unless they have unsupported characters.  

Everyone has a personal taste ;-)

I find single quotes make the file lighter than double quotes (less black ink, heheh)

alexsasharegan: I'm just trying to let my OCD go to my tooling. Of course, then my tooling OCD becomes a problem...

mosinve: doublequotes are default setting for eslint

PR #809, opened 2017-08-10
fix(carousel): Handle changes in slide content

tmorehouse:
Add `observedom` mixin to handle updates to slide content (and delayed rendering of `carousel-slide`)

No guarantee that carousel is bug-free


pi0:
I have temporarily removed carousel from docs sidebar. Do you think we are stable enough now to bring it back?

tmorehouse:
Now includes fixes for slightly messed up transitions (blank space between sliding slides is now gone)

Allows user to programmatically pause and restart the carousel by setting `interval` prop on the fly

tmorehouse:
I think this version fixes many issues (although not positive if it has added any new "bugs")

tmorehouse:
Also added updated docs/example for carousel

lib/components/carousel.vue
```diff
@@ -139,27 +140,35 @@
             },
             value: {
                 type: Number,
-                defrault: 0
```
tmorehouse: This was the main culprit of carousel not showing initially, as value defaulted to `null`/`undefined` if a v-model wasn't bound.

tmorehouse:
This should now be ready for testing before merge.

tmorehouse:
Incorporated changes by @pi0 in the v4-beta branch

PR #821, opened 2017-08-12
fix (form input) input reactivity

### LINKED ISSUE #817: b-form-input does not update reactively after first input
Hi. My problem is as follows. I have this in my template:

```html
<b-form-input type="text" v-model="someField"></b-form-input>
<b-button @click="add">Add<b-button>
```
And a method:
```js
add () {
    this.someField = 'someValue'
}
```
`this.someField` is initially set to `''` in `data`.

What happens: if i click on "Add" before inputing anything to the field, field starts to display "someValue". But if I input anything there and then click the button, field continues to display what I've entered. Vue Devtools inspector shows that `this.someField` is in fact updated to contain `'someValue'`, but `$0.value` in Chrome Devtools shows that actual DOM element still contains the value I've entered, _not_ the one I've set programatically.

If I use raw `<input v-model="someField"/>` in my template, everything works as expected.

Is this a bug in bootstrap-vue or I'm just doing something wrong?

See demonstration in fiddle: https://jsfiddle.net/nz23f2kg/

mosinve:
Supposed to fix #817

### DIFF
--- a/lib/components/form-input.vue
+++ b/lib/components/form-input.vue
@@ -3,6 +3,7 @@
            ref="input"
            :is="isTextArea ? 'textarea' : 'input'"
            :type="isTextArea ? null : type"
+           :value="value"
            :name="name"
            :id="id || null"
            :disabled="disabled"
--- a/lib/components/form-input.vue
+++ b/lib/components/form-input.vue
@@ -3,7 +3,6 @@
            ref="input"
            :is="isTextArea ? 'textarea' : 'input'"
            :type="isTextArea ? null : type"
-           :value="value"
            :name="name"
            :id="id || null"
            :disabled="disabled"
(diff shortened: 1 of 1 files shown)


alexsasharegan:
Would you be open to trying a refactor with v-model? From what I hear about watchers, it would be a more efficient solution.

mosinve:
v-model is just a sugar for :value/@input combination

alexsasharegan:
Agreed, but it's clear that using that works on a native input element. I'm really just looking to see if it's possible to avoid a watcher.

mosinve:
I'll try locally, but watcher is really working solution currently

mosinve:
```
> - <input :type="isTextArea ? null : type" v-model="localValue">:
> v-model does not support dynamic input types. Use v-if branches instead.
```

alexsasharegan:
That should be a `component` tag then. `<component :is="isTextArea ? 'textarea' : 'input' />`

tmorehouse:
Users can make input either a text area, or input with types (text, number, date, password, etc).

So I think it is complaining about input with a dynamic `type` attribute

tmorehouse:
So I think the watcher is the easiest solution for now.

alexsasharegan:
Agreed!

mosinve:
This works at earlier versions somehow... somewhat changes breaks and i dnt get what changes... and why `:value=value` doesn't work in internal input

mosinve:
with this watcher we even can remove `:value=value` from template. It still will work

alexsasharegan:
Well, thanks for giving it a shot, @mosinve! I learned something new about v-model in the process.

tmorehouse:
I'll merge in your changes to the 1.x branch as well.

PR #822, opened 2017-08-12
feat(btn-group): functional component

alexsasharegan:


alexsasharegan:
How does adding teddy coverage decrease test coverage? 🤔

lib/components/button-group.js
```diff
@@ -0,0 +1,41 @@
...
+    render(h, { props, data, children }) {
+        return h(
+            props.tag,
+            mergeData(data, {
+                staticClass: "btn-group",
+                class: {
+                    "btn-group-vertical": props.vertical,
+                    [`btn-group-${props.size}`]: Boolean(props.size)
+                },
+                attrs: { "aria-role": props.ariaRole }
```
tmorehouse: this should be 

```js
attrs: { "role": props.ariaRole }
```

Or perhaps, the prop should be just `role` instead of `ariaRole`

alexsasharegan: This is for the dom attribute that gets injected in the element. Key === attribute name.

alexsasharegan: Or are suggesting to change the component prop?

tmorehouse: the dom element attribute is called `role` (not `aria-role`)

alexsasharegan: Gotcha. I copied that from something else, so I'll see if I can track down the other incorrect case.

lib/components/button-group.js
```diff
@@ -33,7 +33,7 @@ export default {
                     "btn-group-vertical": props.vertical,
                     [`btn-group-${props.size}`]: Boolean(props.size)
                 },
-                attrs: { "aria-role": props.ariaRole }
+                attrs: { "role": props.ariaRole }
```
tmorehouse: 👍

PR #823, opened 2017-08-12
fix(dropdowns): Focus dropdown item on hover

tmorehouse:
Prevent two items from visually having the focus/hover states.

Also removed focusing of dropdown-header elements, as per discussion in https://github.com/twbs/bootstrap/issues/23336 (users should add an `aria-describedby` attribute on dropdown-items that point to their header element, if the dropdown-item needs context from the header)

Also name-spaced custom CSS (adding class `.b-dropdown` and `b-navdropdown` on the root elements)

### DIFF
--- a/lib/components/dropdown-header.vue
+++ b/lib/components/dropdown-header.vue
@@ -1,5 +1,5 @@
 <template>
-    <component :is="tag" tabindex="-1" class="dropdown-header">
+    <component :is="tag" class="dropdown-header">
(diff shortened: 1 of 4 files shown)


pi0:
Bootstrap JS still has this issue 😂

tmorehouse:
Heheh

tmorehouse:
Similar PR #825 for the dev branch

PR #827, opened 2017-08-12
feat(card): functional components for card 🎴

alexsasharegan:


pi0:
@alexsasharegan What's your idea moving `examples` ~> `tests/fixtures` as they are really fixtures now :))

tmorehouse:
Yeah.. they aren't really examples, and we wouldn't want people fiddling with them which would interfere with tests.

alexsasharegan:
Perfect! That seems right to me. Let me make that change once I get home!

tmorehouse:
Should `fixtures` be under the `tests` directory (i.e. `tests/fixtures/`)

alexsasharegan:
@tmorehouse I'll fix that on the next PR

PR #831, opened 2017-08-12
feat(carousel): Fallback support for browsers that dont support transitionend

tmorehouse:
Fallback to `setTimeout` for browsers that do not support `transitionend` event


tmorehouse:
This needs to be merged and reverted.....

PR #834, opened 2017-08-13
fix (input) workaround to Vue ":is" bug

### LINKED ISSUE #833: Textarea does not display the prefilled model
The content of the model is not added to textarea.

Example:
https://bootstrap-vue.js.org/docs/components/form-input
prefill the text3, then still in the textarea nothing is visible.

mosinve:
refactor to v-if/v-else structure
fix #833 and other issues with degraded input reactivity


tmorehouse:
Should we also merge `cih-input-refactor` into `1.x` branch, or just to a merge `dev` into `1.x`?

PR #835, opened 2017-08-13
refactor: move examples to fixtures

alexsasharegan:


tests/helpers.js
```diff
@@ -3,9 +3,9 @@ import { resolve } from 'path';
 import Vue from 'vue/dist/vue.common';
 import BootstrapVue from '../lib';
 
-const readFile = (path) => String(readFileSync(resolve(__dirname, '../examples', path)));
+const readFile = (path) => String(readFileSync(resolve(__dirname, 'fixtures', path)));
 const throwIfNotVueInstance = vm => {
-    if (!vm instanceof Vue) {
+    if (!(vm instanceof Vue)) {
```
alexsasharegan: If anyone is curious, this is an interesting javascript quirk. The `!` operator has a higher precedence than `instanceof` (not so in other languages).

PR #836, opened 2017-08-13
[WIP] feat(col): BS4 column component

alexsasharegan:


tmorehouse:
Bootstrap V4.beta.2 will be bringing the `.offeset-*-*` classes back (which were available in BSV3)

pi0:
Moved into #906

PR #837, opened 2017-08-14
feat(progress): Add height prop

tmorehouse:
Add support for setting the height of the progress bar

Also includes expanded documentation


mosinve:
Whoa!!! we have a progress bar? o_O

PR #839, opened 2017-08-14
fix(alert): Emit dismiss-count-down at 0 seconds

tmorehouse:
Alert never emitted a `dismiss-count-down` event at zero seconds. This PR remedies that.

Updated docs example

Updated tests

### DIFF
--- a/lib/components/alert.vue
+++ b/lib/components/alert.vue
@@ -76,6 +76,9 @@
         mounted() {
             this.showChanged();
         },
+        destroyed() {
+            this.clearCounter();
+        },
         methods: {
             dismiss() {
                 this.dismissed = true;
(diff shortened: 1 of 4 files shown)


lib/components/alert.vue
```diff
@@ -76,6 +76,9 @@
         mounted() {
             this.showChanged();
         },
+        destroyed() {
+            this.clearCounter();
+        },
```
alexsasharegan: This is a smart addition 👍

tmorehouse:
Gave us an extra 0.16% coverage too 😄

lib/components/alert.vue
alexsasharegan: I want to thumbs-up your thumbs-up of my thumbs-up.

PR #840, opened 2017-08-14
fix(pagination): Change classes from .hidden-* to .d-*

tmorehouse:
V4.beta has removed the `.hidden-*` classes, so we must use the `.d-*` classes for hiding on XS screens.

For pagination and pagination-nav we replaced `.hidden-sx-down` with `.d-none` + `.d-sm-flex`

### DIFF
--- a/lib/components/pagination-nav.vue
+++ b/lib/components/pagination-nav.vue
@@ -42,7 +42,7 @@
         </li>
 
         <!-- First Ellipsis Bookend -->
-        <li v-if="showFirstDots" class="page-item disabled hidden-xs-down" role="separator">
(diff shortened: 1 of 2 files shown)


alexsasharegan:
Well this is inconvenient.

tmorehouse:
Yeah.

Not sure of any other components that were using the `.hidden-<breakpoint>-down/up` classes (other than pagination/pagination-nav)

Thee removal of these classes now mean you have to add at least two classes to your elements (one for hiding, and one for displaying)

alexsasharegan:
It's a big deal for my production app. I'll be doing some find & replace by hand work when I upgrade...

PR #843, opened 2017-08-14
feat(card): add card-body functional component & card-img fixes

alexsasharegan:


lib/components/card.js
```diff
@@ -80,10 +84,10 @@ export default {
         // We are building up the markup.
         if (slots().img) {
             childNodes.push(slots().img);
-        } else if (props.img) {
+        } else if (props.img && !props.imgBottom) {
             childNodes.push(
                 h(CardImg, {
-                    props: { top: props.imgTop, bottom: props.imgBottom },
+                    props: { top: props.imgTop, bottom: props.imgBottom, imgFluid: props.imgFluid },
```
tmorehouse: What about building the card-img first in a var, then push it in the if sections? Save on repeating some code.

alexsasharegan: Have a look now. I changed it to accept the img slot as priority, then to dump it up top if top is specified, or if bottom is not specified (making top the default).

lib/components/card.js
```diff
@@ -78,19 +78,18 @@ export default {
...
                 h(CardImg, {
                     props: { top: props.imgTop, bottom: props.imgBottom, imgFluid: props.imgFluid },
                     attrs: { src: props.img, alt: props.imgAlt }
-                })
-            );
+                });
+
+        // The order of these conditionals matter.
+        // We are building up the markup.
+        if (props.imgTop || !props.imgBottom) {
```
tmorehouse: Should this be:
```js
if (props.img || !props.imgBottom) {
```

And always assume img-top if img-bottom not present?

alexsasharegan: I shouldn't need to check `props.img` because of this:
```js
let img =
    slots().img ||
    h(CardImg, {
        props: { top: props.imgTop, bottom: props.imgBottom, imgFluid: props.imgFluid },
        attrs: { src: props.img, alt: props.imgAlt }
    });
```
And yes, img-top if specified, or if bottom is not specified.

alexsasharegan:
Any last words, anybody? I think we're good to go here.

mosinve:
I'll check within 15 mins

tests/components/card.spec.js
```diff
@@ -118,9 +118,8 @@ describe("card", async () => {
         expect($refs.card_group).toBeElement("section");
     });
 
-    it('CardBody should have assigned class', async () => {
-        const {app: {$refs}} = window;
-        const body = $refs.card_body.children[0];
-        expect(body.classList.contains('card-text')).toBe(true);
+    it("CardBody should have assigned class", async () => {
+        const { app: { $refs } } = window;
+        expect($refs.body).toHaveClass("card-text");
```
mosinve: Whoa, dont know that it can be used that way )) 👍

alexsasharegan: Yeah, and it works across FC's and VM's!

PR #847, opened 2017-08-15
fix(forms): BS4.beta form/input validation styles & components

tmorehouse:
Support for the new validation styles and states, and new/changed/removed components.

## Components

**New functional components:**
 - `<b-form-text>` help text block
 - `<b-form-feedback>` only shown when validation fails! or `state="invalid"`, as must be a sibling at the same level as the input and follow after the input
 - `<b-form-row>` for creating tighter `rows` with tighter margins inside forms

**New components:**
 - `<b-form-textarea>` Dedicated component for text area, with additional features

**Breaking change components:**
 - `<b-form-input>` Removed textarea support. `static` prop is now `plaintext` and renders an input with new styles

**Renamed components:**
 - `<b-form-fieldset>` is now `<b-form-group>` (although `<b-form-fieldset>` is available as an alias of `form-group`)

**Removed Components:**
 - `<b-form-static>` removed in favor of new `plaintext` prop on `<bform-input>` and `<b-form-textarea>`

## Validation states:
the `state` props now accepts one of three values: `'valid'` (green color), `'invalid'` (red color), or `null` (default text color)

The new `<b-form-feedback>` component (used in the `<form-group>` component is only shown when validation is `invalid` state.  It is not shown for `valid` states! (BS4 change in behavior)

## Validation can be performed one of three ways.

Two methods involve using browser HTML5 validation.  Inputs must be wrapped in an element that has the class `was-validated` (such as `b-form` and `b-form-group` which have the `validated` prop set (which adds the `was validated` class to the wrapper ( `<b-form-radio>` also supports the `validated` prop).

There are two ways to e control this feature.  One is to use a `b-form` wrapper with `novalidate` **not** set.  The form components will immediately be validated (and display browser native error messages).  The second option is to set `novalidate` and then trap on submit and run the browser native `form.checkValidity()` method and then set `validated` prop on `<b-form>`.  `<b-form>` will emit the native `form submit` event. 

The 3rd option (and probably the preferred method for Vue)  is what BS4 calls server side validation, but it can be run on client.  it applies `is-valid` or `is-invalid` to the inputs via the `state` prop (set to `valid` or `invalid`) which will trigger contextual styles on the form controls.  any `<b-form-feedback>` element taht is a sibling on the invalid control will be shown.


tmorehouse:
This should be pretty much ready for review before merging.

tmorehouse:
@pi0 @alexsasharegan and @mosinve 

Could you guys go over the changes?

PR #848, opened 2017-08-15
feat(dropdowns): functional dropdown components and testing

alexsasharegan:


alexsasharegan:
Moving the code coverage needle `0.67%` at a time 👍

tmorehouse:
That might actually be a `.666%` rounded up to two decimals :p

lib/components/dropdown-header.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default {
+    functional: true,
+    props,
+    render(h, { props, data, children }) {
+        return h(
+            props.tag,
+            mergeData(data, {
+                staticClass: "dropdown-header",
+                attrs: { tabindex: "-1" }
```
tmorehouse: TabIndex no longer needed.

lib/components/index.js
```diff
@@ -78,11 +78,17 @@ export {
     bCardGroup,
     bCardImg,
     bDropdown,
+    bDropdown as bDd,
     bDropdownItem,
+    bDropdownItem as bDdItem,
```
tmorehouse: 👍

lib/components/dropdown-header.js
tmorehouse: The main dropdown mixin doesn't focus headers any more.  users will need to point aria-describedby attribute to a  header with an ID.

alexsasharegan: Should I remove this then?

alexsasharegan: Oh, you did!

PR #864, opened 2017-08-15
feat(nav): new functional nav components

alexsasharegan:


lib/components/form.js
```diff
@@ -0,0 +1,14 @@
...
+    props: {
+        inline: {
+            type: Boolean,
+            default: false
+        }
+    },
+    render(h, { props, data, children }) {
+        return h("form", mergeData(data, { class: { "form-inline": props.inline } }), children);
+    }
+};
```
tmorehouse: Thi will probably conflict with the updated form.vue component in the penndig form-validation branch

tmorehouse:
👍

lib/components/navbar-brand.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+});
+
+export default {
+    functional: true,
+    props: {
+        tag: {
+            type: String,
+            default: null
+        }
```
tmorehouse: Should this be the props as exported above?

alexsasharegan: Hmm... I set it to div! Let me see how that didn't copy over.

alexsasharegan: I see now. I moved props to an export and then didn't use them.

PR #878, opened 2017-08-17
fix(form-group): use new prop labelCols over old computedLabelCols

charkins:


alexsasharegan:
This looks like a slam dunk. Thanks for taking the initiative and just fixing this!

PR #882, opened 2017-08-18
feat(forms): Auto ID generation client side if no id prop provided

tmorehouse:
This is a client side only ID generator, which supplies an ID attribute on form control when no ID prop is provided.

ID generation only runs on client side, during the mounted() hook (after SSR rehydration, if applicable).

The presence if IDs, even if not in the initial SSR rendered HTML will greatly improve accessibility.

Other components that can take advantage of IDs (i.e. dropdowns, carousel, etc) can use the new `id.js` mixin

### DIFF
--- a/lib/mixins/id.js
+++ b/lib/mixins/id.js
@@ -24,5 +24,10 @@ export default {
         safeId() {
             return this.id || this.localId_ || null;
         }
+    },
+    methods: {
+        safeIdPlus(str) {
+             return this.safeId ? `${this.safeId}_${str)` : null;
+        }
     }
 };
--- a/lib/mixins/id.js
+++ b/lib/mixins/id.js
@@ -27,7 +27,7 @@ export default {
     },
     methods: {
         safeIdPlus(str) {
-            str = String(str || '').replace(/\s+/g,'_');
+            str = String(str || '').replace(/\s+/g, '_');
             return (this.safeId && Boolean(str)) ? `${this.safeId}_${str)` : null;
         }
     }
--- a/lib/mixins/id.js
+++ b/lib/mixins/id.js
@@ -21,12 +21,12 @@ export default {
         }
     },
     methods: {
-        safeId(suffix) {
+        safeId(suffix = '') {
             const id = this.id || this.localId_ || null;
             if (!id) {
                 return null;
             }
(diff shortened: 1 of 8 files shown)


lib/mixins/id.js
```diff
@@ -24,5 +24,11 @@ export default {
...
             return this.id || this.localId_ || null;
         }
+    },
+    methods: {
+        safeIdPlus(str) {
+            str = String(str || '').replace(/\s+/g, '_');
+            return (this.safeId && Boolean(str)) ? `${this.safeId}_${str}` : null;
+        }
     }
 };
```
mosinve: IMHO will be more flexible to have one universal methods at this API, which accepts two optional params: `generateId(String suffix = '', Boolean isPrefix=false)`

It could be something like that:
```
generateId(str='', isPrefix=false) {
            str = String(str).replace(/\s+/g, '_');
            if(!!str && this.safeId){
                     return isPrefix? `${str}_${this.safeId}`: `${this.safeId}_${str}`
            }else if(this.safeId){   
            return this.safeId
            }
      return null
}
```

mosinve: https://codepen.io/cih/pen/eEVpEP

tmorehouse: True, but in most cases, this is an internal ID's will have suffix only.

Originally I had planned on a single method, which appended the suffix if provided (similar to what you have above), but I wanted wanted the suffixed ID to return null if the suffix was invalid (i.e. a boolean, undefined, etc).

tmorehouse: but I could convert `safeId` to a method, and optionally append the first argument

lib/mixins/id.js
```diff
@@ -17,12 +17,17 @@ export default {
...
-    computed: {
-        safeId() {
-            return this.id || this.localId_ || null;
+    methods: {
+        safeId(suffix) {
+            const id = this.id || this.localId_ || null;
+            if (!id) {
+                return null;
+            }
+            suffix = String(suffix || '').replace(/\s+/g, '_');
```
mosinve: With default value at function definition you do not need to do `suffix || ''`

mosinve: on the whole, this PR now looks better with one flexible method 👍

PR #886, opened 2017-08-18
feat:(input-group): Convert sub components to functional components

tmorehouse:
Converted `input-group-addon` and `input-group-button` to functional components


lib/components/input-group.vue
```diff
@@ -1,19 +1,25 @@
 <template>
-    <component :is="tag" :class="classObject" role="group">
+    <component :is="tag" :id="id || null" :class="classObject" role="group">
```
tmorehouse: In this case ID is not required.  Would the validator crap out (invalid when passed null or am empty string?

Vue typically will add the attribute if passed an empty string.  The `|| null` ensures that the attribute will not appear if anything falsey is given (`null`, `''`. `0`, etc)

alexsasharegan: The validator would issue a prop warning in dev mode, but wouldn't throw an actual exception.

I see you're right about passing the attr with an empty string:
image

alexsasharegan: It seems potentially beneficial to have that prop warning in place since I can't think of a good reason for issuing an element id as an empty string.

tmorehouse: So does the validator only run if the default is not selected?

alexsasharegan: It runs when you pass a prop down.

tmorehouse: OK, so if no value provide it doesn't run, but only runs if a value is passed.

alexsasharegan: Exactly 👍

PR #887, opened 2017-08-18
Fix links to <b-collapse> and <b-dropdown> in Navbar docs

johnfraney:
The links could be broken in other places as well. I'll continue to send in PRs as I use the docs. Let me know if they get annoying and I can group them all together!


johnfraney:
It's possible that there's a conflict between links in the alpha and beta docs. The links appear to be working properly in the [beta docs](https://bootstrap-vue.now.sh/docs/components/navbar)

tmorehouse:
The links wont work in the `README.md` fails when viewing on github, but they work correctly when the docs pages are rendered, as teh page slugs are rendered without a trailing slash.

mosinve:
Hey, links work well at docs site https://bootstrap-vue.js.org/docs/components/navbar/ - just check em. Changes in this PR aren't needed

johnfraney:
@tmorehouse @mosinve Thanks for checking into that. The links are still broken for me in the Navbar docs. See the link to `<b-dropdown>` in this example: https://bootstrap-vue.js.org/docs/components/navbar/#-b-nav-item-dropdown-

mosinve:
Try to remove  `#-b-nav-item-dropdown-` from url, manually scroll to 'troubled' link and check it

johnfraney:
@mosinve It doesn't work for me with or without the hash. Is it working for you?

mosinve:
yes, it work for me

tmorehouse:
Remove the trailing slash after `navbar/`

Pages are rendered without the trailing slash.

tmorehouse:
i.e.
https://bootstrap-vue.js.org/docs/components/alert
https://bootstrap-vue.js.org/docs/components/navbar
https://bootstrap-vue.js.org/docs/components/collapse
https://bootstrap-vue.js.org/docs/components/button
etc.

tmorehouse:
Hmm 

I think this is more of a bug in the Vue router routines when  rendering pages.  If you go direct to a link, vue router is adding the trailing slash.

But if you go to a page via the side navbar, a trailing slash is not added.

We need to create an issue to have the docs rendered updated, as this PR is not the way to fix it.

johnfraney:
@tmorehouse Good catch! I was starting to wonder if I was going crazy there for a second.

PR #901, opened 2017-08-21
fix(tab): transitions

### LINKED ISSUE #504: Tabs not fading
It is stated in the docs that fade is enabled by default but the tabs don't fade. This can be seen in the corresponding [docs page](https://bootstrap-vue.github.io/docs/components/tabs) itself.

Maybe something changed in bootstrap that broke the fading? (Is something like that even possible?)

mosinve:
Make it transition again
Attempt to fix #504

### DIFF
--- a/lib/components/tab.vue
+++ b/lib/components/tab.vue
@@ -1,5 +1,5 @@
 <template>
-    <transition @enter="enter" @before-leave="beforeLeave" mode="out-in">
+    <transition @after-enter="afterEnter" @before-leave="beforeLeave" mode="out-in">
         <component :is="tag"
                    :id="id || null"
                    role="tabpanel"
@@ -19,7 +19,7 @@
 <script>
     export default {
         methods: {
-            enter() {
+            afterEnter() {
                 this.show = true;
             },
             beforeLeave() {
--- a/lib/components/tab.vue
+++ b/lib/components/tab.vue
@@ -7,8 +7,7 @@
                    :aria-hidden="localActive ? 'false' : 'true'"
                    :aria-expanded="localActive ? 'true' : 'false'"
                    :aria-lablelledby="controlledBy || null"
-                   v-if="localActive || !lazy"
-                   v-show="localActive || lazy"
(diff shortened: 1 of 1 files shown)


lib/components/tab.vue
```diff
@@ -7,8 +7,7 @@
                    :aria-hidden="localActive ? 'false' : 'true'"
                    :aria-expanded="localActive ? 'true' : 'false'"
                    :aria-lablelledby="controlledBy || null"
-                   v-if="localActive || !lazy"
-                   v-show="localActive || lazy"
+                   v-show="localActive"
```
tmorehouse: what if this was:

```html
v-if="localActive && lazy"
v-show="localActive"
```
Would that make a difference?

Some people may not want any child components of a tab rendered if not visible.

mosinve: if there's a v-show, then element will be rendered

mosinve: but if only v-if stay... transitions are buggy...

mosinve: then .. judgement is simple- v-show is the only case when this component works just like original BS4

tmorehouse: But removing the lazy part will always render with v-show (but just have `display:none;` set), so the content of the tabs will always be in the dom (and potentially active) even though not visible, which may not be the desired state for some users

tmorehouse: Maybe if we move the lazy part to the tab content slot?  So the tab itself is always present 9but hiding with v-show), but the content inside can be taken out of DOM via a v-if?

tmorehouse: What aaaaabout this:

````
                   v-show="localActive"
                   ref="panel"
        >
             <template v-if="localActive || !lazy">
                 <slot></slot>
             </template>
</component>
```

mosinve: Check updated file... seems to work, at least at playground

mosinve: the catch is we must spread active and show classes in time...

Issue #868, opened 2017-08-16
[b-tabs] setting first tab to active not working


pimlie:
Probably a regression from #687

If you set the first tab explicitly as active, the tab is _not_ shown on load although the correct index is set. See https://jsfiddle.net/25jwyoqu/1/

tmorehouse:
Have you been abl;e to confirm if this is till the case in v1.0.0.beta as well?

tmorehouse:
Just played with your fiddle.. and can see that it does have issue, but only with first tab.

pimlie:
Unfortunately I have not found the time yet to test v1.0.0-beta. 

Actually I took the fiddle from issue #687 and only moved the active property from the second to the first tab .

The workaround for now is to just not set the first tab as active, as it is the first tab it is active by default and then it works!

tmorehouse:
The code for both c0.22.x and v1.0.0.beta.3 should be almost identical.

I'll take a look at this later today.

pimlie:
@tmorehouse  Thanks, i noticed that removing all active properties as workaround messes up server side rendering as no tab content is generated on ssr at all. That missing content is quite an issue for search engines..

tmorehouse:
Content should be rendered based on the 'lazy' prop not being set (and just hidden)

tmorehouse:
This is fixed now in V1.0.0.beta.5 https://bootstrap-vue.js.org/docs/components/tabs

PR #903, opened 2017-08-21
fix(tabs): Better handling of active tab and transitions

tmorehouse:
Better handling of tabs with `active` state (issue #868)

Incorporates transition fixes from #901 

Child tabs now sniff parent `lazy` and `fade` props


lib/components/tabs.vue
```diff
@@ -146,28 +142,28 @@
...
-                this.setTab(this.currentTab, false, 1);
+                this.setTab(this.currentTab + 1, false, 1);
             },
 
             /**
              * Move to previous tab
              */
             previousTab() {
-                this.setTab(this.currentTab, false, -1);
+                this.setTab(this.currentTab - 1, false, -1);
```
mosinve: why to change index? i think it calculates inside `setTab` method

tmorehouse: I refactored how set tab works with the offset direction value (3rd argument to st tab), which now specifies a direction to move if the requested tab is disabled (skipping over disabled)

mosinve: Ok, i think i get it )

tmorehouse: :)

lib/components/tabs.vue
```diff
@@ -146,28 +142,28 @@
...
              * Set active tab on the tabs collection and the child 'tab' component
              */
-            setTab(index, force, offset) {
-                offset = offset || 0;
+            setTab(index, force, direction) {
+                direction = this.sign(direction || 0);
 
                 // Prevent setting same tab!
-                if (!force && (index + offset) === this.currentTab) {
+                if (!force && index === this.currentTab) {
```
mosinve: `index === this.currentTab` but this part is never fired... because `index = this.currentTab+[1|-1]`

lib/components/tabs.vue
tmorehouse: Also, this removes the new for people to call previous and next methods directly during external control.  They can now just increment the value and it will look for the next tab if available.,

lib/components/tabs.vue
tmorehouse: Not any more ;-)

mosinve: and we have this expression at fact:
`if (!force && (this.currentTab+[1|-1] === this.currentTab)) {`

tmorehouse: Index is now the requested tab, and offset (which is now called direction) is a direction to move if the requested tab is disabled.

mosinve: is there a reason to be this in code?

lib/components/tabs.vue
mosinve: but as i recall.. v-model changing tabs works earlier... but i'm not sure )

lib/components/tabs.vue
tmorehouse: the variable offset doesn't exist anymore.

If the tab is already active, we can save some processing power.  But maybe it might be safer to always run the code below, just to make sure only one tab is ever active at a time.

lib/components/tabs.vue
tmorehouse: It did, but it couldn't handle wrapping around properly, and skipping disabled tabs

Issue #911, opened 2017-08-22
b-tabs active property hiding content


nc-csl:
Hi. 
When I use the b-tabs component and add the "active" property, the content of the tab is not displayed. When removing the active property, content is shown again.

mosinve:
Please provide an example or description of steps to reproduce this issue

mosinve:
You can change active state of tabs by pass index to v-model

tmorehouse:
Is this also happening in V1.0.0.beta.5?

mosinve:
Yep, quick fix is to set show: true at declaration

mosinve:
This fixed in both alpha & beta branches. Will be avalable in next releases.

tmorehouse:
Should now fixed in v1.0.0.beta.6 release.

nc-csl:
Thank you for confirming the bug and for a solution!

PR #912, opened 2017-08-22
fix(tabs): Show first tab when set to active

tmorehouse:
May address issue #911

### DIFF
--- a/lib/components/tab.vue
+++ b/lib/components/tab.vue
@@ -2,7 +2,6 @@
     <transition @before-enter="beforeEnter"
                 @after-enter="afterEnter"
                 @after-leave="afterLeave"
-                :duration="computedFade ? 150 : 0"
                 mode="out-in">
         <component :is="tag"
                    :id="safeId()"
@@ -46,7 +45,7 @@
             tabClasses() {
                 return [
                     'tab-pane',
-                    this.show ? 'show' : '',
+                    (this.show || !this.computedFade) ? 'show' : '',
                     this.computedFade ? 'fade' : '',
                     this.disabled ? 'disabled' : '',
                     this.localActive ? 'active' : ''
--- a/lib/components/tab.vue
+++ b/lib/components/tab.vue
@@ -36,7 +36,7 @@
         data() {
             return {
                 localActive: this.active && !this.disabled,
-                show: Boolean(this.active && !this.disabled)
+                show: this.active && !this.disabled
             };
         },
         computed: {
--- a/lib/components/tab.vue
+++ b/lib/components/tab.vue
@@ -1,5 +1,10 @@
 <template>
-    <transition @before-enter="beforeEnter" @after-enter="afterEnter" @after-leave="afterLeave" mode="out-in">
+    <transition @before-enter="beforeEnter"
+                @after-enter="afterEnter"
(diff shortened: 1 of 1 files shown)


lib/components/tab.vue
```diff
@@ -36,7 +36,7 @@
         data() {
             return {
                 localActive: this.active && !this.disabled,
-                show: false
+                show: Boolean(this.active && !this.disabled)
```
mosinve: why not just localActive?

tmorehouse: We don't have access to `this.localActive` yet at this stage, as it hasn't been returned to the VM.

mosinve: check this https://codepen.io/cih/pen/eErKvm

mosinve: ah.... but then at mounted? i dont like this similar code.....

mosinve:
Troy... i thought... isn't localActive supposed to be a computed value? if it's a data - then it will be valuated only once.. upon creation of component

tmorehouse:
b-tabs sets the `localActive` data in `b-tab`, as it controls which tabs are visible

mosinve:
but computed vars may have a setter..

tmorehouse:
b-tabs still need to be able to control which tab is active through:
https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/components/tabs.vue#L189-L196

lib/components/tab.vue
tmorehouse: Hmmm... mounted may work, although it should maybe look at localActive on mounted (depending on if b-tabs is set to `lazy` or not.

mosinve:
and if we assign active prop not at design, but at runtime? will it handle it?

tmorehouse:
Someone can manually set active on all their tabs, but we ensure that only one tab should be visible at a time (we select the last 'active" tab found).

maybe show should be a computed prop (or an `inShow` computed prop that looks at the `show` data and the localActive data?

mosinve:
tabs not dynamically following tab's `active` prop. it looks at localActive, which sets up at mount... so if we set `tab.active` at runtime - this wont work.

tmorehouse:
I tried setting active tab on fiirst tab (with `currentTab` set to index 0, and the tab is there, but missing the `show` property.,

mosinve:
at runtime?

mosinve:
let's try with computed show, i'll check all my doubts at local playground...

tmorehouse:
In the b-tabs doc page.

I set first tab as active (with active prop), and left the app data `tabIndex` as 0

Tab is definitely showing, but with opacity still at 0 because the `show` class was missing on the active tab.

It appears the first `active` tab  isn't triggering the transition triggers, which then add the `show` cl;ass to the `b-tab`

tmorehouse:
You can see above, that the tab is `active` but missing the `show` class, so opacity is still `0`

mosinve:
Ah... I saw this.. and after that i tried to set show at true ... and so on... 
But it wont help in case active will set true at runtime...

tmorehouse:
We could change the css transition sniffing and set a duration on the transition to 150ms (if fade, 0 if noFade)

PR #913, opened 2017-08-22
fix(dropdowns): Migration to popper.js positioning

tmorehouse:
Migration of dropdown and nav-item-dropdown to use popper.js

Migration wasn't too bad.  a couple of new optional props (`offset` & `no-flip`), but I was able to make it work with our existing props (`dropup` & `right`)

By design, popper is disabled for menus in navbars (because of collapsing behavior of navbars)

### DIFF
--- a/lib/mixins/dropdown.js
+++ b/lib/mixins/dropdown.js
@@ -107,7 +107,7 @@ export default {
             return this.$refs.button.$el || this.$refs.button;
         }
     },
-    destroy() {
+    destroyed() {
         if (this._popper) {
             this_popper.destroy();
         }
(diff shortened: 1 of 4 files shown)


lib/mixins/dropdown.js
```diff
@@ -13,14 +15,24 @@ function filterVisible(els) {
...
+    // Dropdown Right Align
+    BOTTOMEND: 'bottom-end'
+};
+
 export default {
     mixins: [clickoutMixin, listenOnRootMixin],
     props: {
-        id: {
-            type: String
-        },
```
tmorehouse: this comes from the `id` mixin now

mosinve:
Checked local playground. It's very buggy.

tmorehouse:
What kind of bug are you seeing?  I wonder if we are using the correct version of popper?

Bootstrap recently upgraded (or downgraded) their popper version

mosinve:
i uploaded screencast to slack

tmorehouse:
This should now be ready for merging.

PR #914, opened 2017-08-22
[wip] feat(table): easier usage

pi0:
While working on tables on some quick cases, the current API is little hard to use and some times there are really some extra works for easy tasks (like changing column titles for each field or even when need a quick representation of data) The solution was using computed props.

Example of new usage:
```vue
<template>
  <b-table :items="logs"/>
</templte>

<script>
export default {
 data: () => ({
  logs: [
    {log_type: 'CHECK_IN', deviceId: '123', user: 'A'},
    {log_type: 'CHECK_OUT', deviceId: '44', user: 'B'},
    {log_type: 'CHECK_IN', deviceId: '3343434', user: 'C'}
  ]
  })
}
```

Renders:
image

Actually, with example above, fields is auto sampled from first record and it's column names are auto formatted using [startCase](https://lodash.com/docs/4.17.4#startCase)

Fields property can now be either an array (of needed fields) or a more flexible object:

```js
// will render two columns "Log Type" and "User Name"
let fieldsArray = [ 'log_type', 'user_name' ]

fields = {
 // Old style style possible
 a: { label: 'Foo' },
 // Auto generate column with label "B ID"
 bId: true,
 // Hide c 
 c: false, 
 // Column "User Name" with username value
 userName: r => r && r.user && r.user.username 
}
```

#### TODO

- [x] Currently, somewhere callFormatter is being called with null value

- [x] We may also simplify at least first example in Table docs maybe moving current main example to the end as "Advanced Usage"

- [x] There is a hidden performance typo currently in Table implementation of `hasFormatter` and `callFormatter`
Adding `console.log` or this to one of those functions for debug:
```js
if(!window.hasFormatter) window.hasFormatter = {} ; if(!window.hasFormatter[field])window.hasFormatter[field]=0 ; window.hasFormatter[field]++ 
```
shows a huge number of hasFormatter calls:
```
window.hasFormatter
02:02:26.058 {[object Object]: 312}
```

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -418,7 +418,7 @@
                     }
                     // Label shortcut
                     if (typeof fields[k] === 'string') {
-                        fields[k] =  { label: fields[k] }
+                        fields[k] =  { label: startCase(k) }
                     }
                 })
 
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -498,7 +498,7 @@
         },
         methods: {
             keys,
-            fieldClass(field, key) {
+            fieldClass (field, key) {
                 return [
                     field.sortable ? 'sorting' : '',
                     (field.sortable && this.localSortBy === key) ? 'sorting_' + (this.localSortDesc ? 'desc' : 'asc') : '',
@@ -507,7 +507,7 @@
                     field.thClass ? field.thClass : ''
                 ];
             },
-            tdClass(field, item, key) {
+            tdClass (field, item, key) {
                 let cellVariant = '';
                 if (item._cellVariants && item._cellVariants[key]) {
                     cellVariant = (this.inverse ? 'bg-' : 'table-') + item._cellVariants[key];
@@ -519,12 +519,12 @@
                     field.tdClass ? field.tdClass : ''
                 ];
             },
-            rowClass(item) {
+            rowClass (item) {
                 return [
                     item._rowVariant ? ((this.inverse ? 'bg-' : 'table-') + item._rowVariant) : ''
                 ];
             },
-            rowClicked(e, item, index) {
+            rowClicked (e, item, index) {
                 if (this.computedBusy) {
                     // If table is busy (via provider) then don't propagate
                     e.preventDefault();
@@ -533,7 +533,7 @@
                 }
                 this.$emit('row-clicked', item, index, e);
             },
-            rowDblClicked(e, item, index) {
+            rowDblClicked (e, item, index) {
                 if (this.computedBusy) {
                     // If table is busy (via provider) then don't propagate
                     e.preventDefault();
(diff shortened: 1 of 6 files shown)


tmorehouse:
This sounds like a great idea.

alexsasharegan:
Agreed. 👍

mosinve:
I like it. Having ability to compute fields from items array allows to use single table for different data sources, we have only to change api-url.

Whats to do with formatter related things?

tmorehouse:
Made a first stab t documentation updates.

pi0:
@tmorehouse  Nice job :)

@mosinve I think the best way would be calling formatters inside `computedItems` so vue can optimised for minimum calls.

mosinve:
@pi0 , tried your suggestion and it seems that `computedItems` is not good place, because we need to access to `computedFields.formatter` prop, but `computedFields` depends on `computedItems` itself, so we get a loop...

tmorehouse:
@pi0 @mosinve Maybe if computedItems checked if a fields prop was passed?  and if so, then c all the formatter, else, assume no formatter?

mosinve:
@tmorehouse @pi0 good idea, will try it

mosinve:
Well, partially works, but it seems, that ` items = items.slice();` patr doesn't work as expected... original `this.items` got mutated.

lib/components/table.vue
```diff
@@ -441,7 +438,18 @@
                 }
 
                 // Shallow copy of items, so we don't mutate the original array order/size
-                items = items.slice();
+
+                items = JSON.parse(JSON.stringify(items))
```
tmorehouse: This would no longer be a shallow copy of items, but a deep copy (which means much more memory usage as a side effect, and slightly slower than a slice)

lib/components/table.vue
```diff
@@ -441,7 +438,18 @@
...
+
+                // Apply formatter only if fields object defined
+                if (this.fields && this.fields.toString() === '[object Object]'){
+                    const keysToFormat = keys(this.fields).filter((k, i, a) => this.hasFormatter(this.fields[k]))
+                    keysToFormat.forEach(k =>{
+                        items.forEach(item =>{
+                            item[k] = this.callFormatter(item, k, this.fields[k])
+                        }, this)
```
tmorehouse: So I see why using JSON.stringy now... since you are altering the raw data, rather than at the render level (in the template)

mosinve: yep, this is about @pi0 wrote, i think... at least i understand it this way)))

mosinve: And it needs to cancel all Vue reactivity

mosinve:
So, can it be merged?

pi0:
@mosinve Yeah seems already good.

tmorehouse:
Docs could maybe use a bit more examples (live), but we can tweak that later.  We have the essentials in there.

Issue #747, opened 2017-07-28
[Informational] Bootstrap V4.beta 1, 2 & 3 ship list


tmorehouse:
Changes / features in Bootstrap V4.beta:

V4.beta.1: https://github.com/twbs/bootstrap/issues/21568 ✔️ 
V4.beta.2: https://github.com/twbs/bootstrap/issues/23278 ✔️ 
V4.beta.3: https://github.com/twbs/bootstrap/issues/24439

We will use this issue to track progress for Bootstrap-Vue v1.0.0.x

Bootstrap-Vue components affected (we'll ad more as we find out more):
- [x] `b-tooltip` popper.js (PR #923) implemented as a directive & component
 - [x] `b-popover` popper.js (PR #923) implemented as a directive & component
 - [x] `b-badge`:
 - - [x] `b-badge` class `.badge-default` removed in beta.1
 - - [x] V4.beta.2 supports "actionable" badges (i.e. links with HREF) https://github.com/twbs/bootstrap/pull/24422
 - [ ] `b-dropdown`:
 - - [x] popper.js positioning (PR #913)
 - - [ ] Dropleft and dropright support (See https://github.com/twbs/bootstrap/pull/23860)
 - - [ ] ~Allow `container` option for dropdowns (see https://github.com/twbs/bootstrap/pull/24257)~ Canned as it breaks keyboard navigation
 - - [ ] Support for forms in dropdown (see https://github.com/twbs/bootstrap/issues/22591 and https://github.com/twbs/bootstrap/pull/22857)
 - [x] `b-nav-item-dropdown`:
 - - [x] `popper.js` - exclude positioning (see https://github.com/twbs/bootstrap/issues/23553) (PR #913)
https://github.com/twbs/bootstrap/pull/22857)
 - [x] `b-table`:
 - - [x] `b-table` renamed `.table-inverse`, `.thead-inverse`, and `.thead-default` to `.*-dark` and `.*-light`, matching BSV4 color schemes used elsewhere. See https://github.com/twbs/bootstrap/pull/23560. `b-table` docs: Bootstrap-Vue already supports any variant name on b-table, but needs updated documentation when V4.beta.2 ships. (PR #1222)
[...]

PR #923, opened 2017-08-24
feat: New popper.js based tooltip/popover directives and components

tmorehouse:
New tooltip directive `v-b-tooltip`, applied to any element/component that one would like a tooltip on.

**Directive syntax:**
```
v-b-tooltip:[container].[mod1].[mod2].[...].[modN]="<value>"
```
Where `<value>` can be (optional):
 - A string (content of the tooltip, Optional HTML supported)
 - A function reference (to generate the content of the tooltip, received one arg of the element reference)
 - An object containing more complex configuration of tooltip, See Bootstrap docs for possible values/structure, supports most configuration params, including custom templates)

Where [modX] can be (all optional):
 - Positioning: `top`, `bottom`, `left`, `right`, or `auto` (last one found wins, defaults to `top`)
 - Event trigger: `click`, `hover`, `focus` (if none specified, defaults to `focus` and `hover`)
 - A delay value in the format of `d###` (where `###` is in ms, defaults to 0);
 - An offset value in pixels in the format of `o###` (where `###` is the number of pixels, defaults to 0)
 - `nofade` to turn off animation
 - `html` to enable rendering raw HTML & DOM node. by default HTML or DOM node is escaped and converted to text

Where [container] can be (optional)
 - An element ID (minus the #) to place the tooltip markup in
 - If not provided, tooltips are appended to the body

**Simplest usage:**
```
v-b-tooltip="'This is a Tooltip!'"
```
or use the element's `title` attribute for the tooltip content:
```
v-b-tooltip title="This is a title Tooltip"
```
**Placement examples:**
```
v-b-tooltip.bottom
v-b-tooltip.right
```
**Trigger examples:**
```
v-b-tooltip.hover  => Hover only
v-b-tooltip.click  => Click only
v-b-tooltip.hover.focus => Both hover and focus
```

**Combo:**
```
v-b-tooltip.hover.bottom  => Show on hover and place at bottom
v-b-tooltip.bottom.hover  => Same as above
```

If a title/message is not provided, then the tooltip will not display. For popover, either the title or content must be provided.

**Popover:**
`v-b-popover` directive works similar, except that `<value>`, if a string (or function) is the content of the popover, and the `title` attribute will be used for the popover heading.  You can also provide both title and content via an object passed as value:
```
v-b-tooltip="{title='This is title', content:'This is the content'}"
```

_Popover extends the ToolTip class object, overriding some of it's methods and static getters._

Based on https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js and https://github.com/twbs/bootstrap/blob/v4-dev/js/src/popover.js

Also added replacement `<b-popover>` and `<b-tooltip>` components.. **Breaking Change Warning:** these replaced components now have different usage signatures.

Addresses BSV4.beta shiplist #747


pi0:
@tmorehouse Having some SSR issues:
image

I'm checking for possible workarounds :)

tmorehouse:
Hmmm This should only be on client side

tmorehouse:
Ah, probably should check for inBrowser on unbind

tmorehouse:
Would this be colliding with existing tooltip component?

pi0:
@tmorehouse (Stilll :D) Don't know or maybe just a web pack bug with prototypes!

tmorehouse:
It's Basically the same structure as scrollspy (function with prototypes applied)

pi0:
b204ff7d28fd0452a4584f84a0d0861bdc3b9cef 😆

lib/directives/tooltip.js
```diff
@@ -108,7 +108,7 @@ if (inBrowser && window.Element && !Element.prototype.closest) {
 /*
  * ToolTip Class
  */
-function Tooltip(el, bindings, vnode) {
+function ToolTip(el, bindings, vnode) {
```
tmorehouse: 👍

tmorehouse:
Now the question is, does this work? 😆

pi0:
Still some errors with playground:

image


@tmorehouse Are you testing without docs dev or watch?

lib/directives/tooltip.js
```diff
@@ -0,0 +1,779 @@
...
+    // Sanitize Title
+    let title = this.$config.title || '';
+    if (title && typeof title === 'number') {
+        title = config.title.toString();
+    }
+    this.$config.title = title;
+    // Title can also be the element's 'title' attribute if not provided here
+
+    // Special handling of event trigger modifiers Trigger is a space serated list
+    let triggers = this.config.trigger;
```
mosinve: Still an error here at playground, maybe to add nullcheck here?

tmorehouse:
I'm now working on migrating the "guts" of this directive over to a class object file, which will be able to be re-used with popovers (since Popover inherits much of the guts of ToolTip.

mosinve:
it's not workin at all currently, i try to fix it.

mosinve:
There's an error at compileHtml when try to show TT

> Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.

https://stackoverflow.com/questions/21926083/failed-to-execute-removechild-on-node

tmorehouse:
working on the compileHtml routine in the new class based tooltip/popover

lib/directives/tooltip.js
```diff
@@ -79,7 +79,7 @@ function generateId() {
 function compileHtml(str) {
     let div = document.createElement('div');
     div.innerHTML = String(str || '').trim();
-    const node = div.firstElementChild ? div.removeChild(div.FirstElementChild) : null;
+    const node = div.firstElementChild ? div.removeChild(div.firstElementChild) : null;
```
mosinve: My point is that this line isn't needed at all, as this var at functions scope and will be handles by GC after function ends

mosinve: I'd refactor whole func to:
```
// Compile a string of HTML and return the DOM element
function compileHtml(str) {
    let div = document.createElement('div');
    div.innerHTML = String(str || '').trim();
    return div.firstElementChild;
}
```

tmorehouse: But I need to remove the compiled template from the DIV, otherwise it still has a `parentElement` (which is the outer holder `div`).

parentElement is used to detect if the tooltip is still in the document DOM.

lib/classes/tooltip.js
```diff
@@ -1,11 +1,10 @@
 import Popper from 'popper.js';
 import { assign, keys } from '../utils/object';
 import { isArray, from as arrayFrom } from '../utils/array';
-import { assign, keys } from '../utils/object';
 
 const NAME = 'v-b-tooltp';
 const CLASS_PREFIX = 'bs-tooltip';
-const BSCLS_PREFIX_REGEX = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g');
+const BSCLS_PREFIX_REGEX = new RegExp(`\\b${CLASS_PREFIX}\\S+`, 'g');
```
tmorehouse: See https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L37

mosinve: it matches with space at start

mosinve: why to blindly follow things, that could be done better?

mosinve: https://regex101.com/r/Kocwnl/1

tmorehouse: I know theirs works. Just trying to keep as close to their code (although changing all jQuery to native DOM functions), as it will make it easier to include any fixes that TWBS makes (they just made a fwe commit changes to Tooltip/Popover today)

mosinve: and this is BS's 
https://regex101.com/r/OgnXoS/1
String.match returns full match

mosinve: I got almost whole this PR fixed... one thing left - it does not respect position

mosinve: I suggest to do a PR to BS repo :)

tmorehouse: jQuery allows removing a bunch of class in a single `$(el).remoceClass(...)` operation (space separated), whereas native DOM `el.classList.remove(...)` only allows one class name (no spaces) to be removed per call... that is why they were capturing the space before the class name (there can be multiple classes with the prefix on the tooltip element)

lib/directives/tooltip.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+export default {
+    bind(el, bindings, vnode) {
+        applyBVTT(el, bindings, vnode);
+    },
+    inserted(el, bindings, vnode) {
+        applyBVTT(el, bindings, vnode);
+    },
+    update(el, bindings, vnode) {
+        if (bindings.oldValue !== bindsings.oldValue) {
```
mosinve: ??? maybe 
`if (bindings.value !== bindings.oldValue) {`

tmorehouse: ooops... Yep.  So used to watchers

lib/classes/tooltip.js
```diff
@@ -0,0 +1,632 @@
...
+
+    static get NAME() {
+        return NAME;
+    }
+
+
+    // Update config
+    updateConfig(config) {
+        // Merge config into defaults
+        let updatedConfig = assign({}, ToolTip.constructor.Default, config);
```
tmorehouse: we have to use `this..constructor.Default` rather than `ToolTip.constructor.Default`, otherwise Popover will look like a tool tip.  so `this` is used in order to ensure that popover, when it overrides defaults gets the right template

lib/classes/tooltip.js
```diff
@@ -373,7 +373,7 @@ class ToolTip {
     getTipElement() {
         if (!this.$tip) {
             // Try and compile user supplied template, or fallback to default template
-            this.$tip = this.compileTemplate(this.$config.template) || this.compileTemplate(this.constructor.Default.template);
+            this.$tip = ToolTip.compileTemplate(this.$config.template) || ToolTip.compileTemplate(this.constructor.Default.template);
```
tmorehouse: We cant do this, otherwie Popover becomes a tooltip..  This needs to be `this.` references

lib/classes/tooltip.js
```diff
@@ -272,14 +272,18 @@ class ToolTip {
         };
 
         // Show tip
-        this.$tip.classList.add(ClassName.SHOW);
+        tip.classList.add(ClassName.SHOW);
         this.setOnTouchStartListener(true);
 
-        if (this.$tip.classList.contains(ClassName.FADE)) {
+        this.transition(tip, transEvent, complete, ClassName.SHOW);
```
tmorehouse: Shouldn't this be `ClassName.FADE` ?

tmorehouse: As the transition only runs if FADE is enabled

lib/classes/tooltip.js
```diff
@@ -345,12 +313,40 @@ class ToolTip {
...
 
-        this.transition(tip, transEvent, complete, ClassName.FADE);
+        this.transition(tip, transEvent, complete);
 
         this.$hoverState = '';
     };
 
-    update() {
+    transition (tip, transEvent, complete) {
+        if (tip.classList.contains(ClassName.FADE)) {
```
mosinve: Why you remove 4rd arg, which allow to select ftansition class?

tmorehouse: The class is always FADE...  It is the only one that triggers transitions. SHOW doesn't

tmorehouse: Adding or removing SHOW triggers the transition, but fad stats that the tooltip has animation.

mosinve: K

lib/directives/popover.js
```diff
@@ -32,7 +32,7 @@ function parseBindings(bindings) {
         config.content = bindings.value;
     } else if (typeof bindings.value === 'object') {
         // Value is config object, so merge
-        config = assign(this.$config.value);
+        config = assign(bindings.value);
```
tmorehouse: 👍

lib/directives/popover.js
```diff
@@ -50,8 +50,8 @@ function parseBindings(bindings) {
             // no animation
             config.animation = false;
         } else if (/^(auto|top|bottom|left|right)$/.test(mod)) {
-            // placement of tooltip
-            config.position = mod;
+            // placement of popover
+            config.placement= mod;
```
tmorehouse: 👍

tmorehouse:
Added in events hooks to ToolTip class, so that our `<b-tooltip>` and `<b-popover>` components can emit `show`, `shown`, `hide`, `hidden` events

docs/components/popover/README.md
```diff
@@ -87,14 +87,14 @@ element via the `<b-popover>` component or `v-b-popover` directive.
     },
     methods: {
       onCancel() {
-        this.$parent.$emit('close');
+        this.$refs.popover.$emit('close');
```
tmorehouse: Yeah..  I kept thinking app was a child of b-popover, but its not...  good catch!

lib/classes/popover.js
```diff
@@ -67,6 +67,17 @@ class PopOver extends ToolTip {
         }
     };
 
+    getTitle() {
+        let title = this.$config.title || '';
+        if (typeof title === 'function') {
+            title = title.call(this.$element).toString();
+        }
+        if (!title) {
```
tmorehouse: I think we need to chek here to see if title is also a DOM node (`typof title === 'object' && title.nodeType && title.innerHTML.trim())`), and if so, just return the title as a node

lib/components/popover.vue
```diff
@@ -123,7 +123,7 @@
                 const cfg = assign({}, this.baseConfig);
                 if (this.$refs.title.innerHTML.trim()) {
                     // We pass the DOM element to preserve components
-                    cfg.title = this.$refs.title;
+                    cfg.title = this.$refs.title.innerHTML;
```
tmorehouse: If we do this, we can't pass components to the title (i.e. say someone wants top place a close button in the title)

lib/classes/popover.js
```diff
@@ -67,6 +67,17 @@ class PopOver extends ToolTip {
...
+            title = this.$element.getAttribute('title') || this.$element.getAttribute('data-original-title') || '';
+        }
+        return title.trim();
```
tmorehouse: Actually, moving the `.trim()` method up above fixes this.

lib/mixins/dropdown.js
```diff
@@ -1,49 +1,50 @@
...
+import { assign } from "../utils/object";
 
 // Determine if an HTML element is visible - Faster than CSS check
-const isVisible = el => el && (el.offsetWidth > 0 || el.offsetHeight > 0);
+function isVisible(el) {
+    return el && (el.offsetWidth > 0 || el.offsetHeight > 0);
+}
 
 // Return an Array of visible items
-const filterVisible = els => (els ? els.filter(el => isVisible(el)) : []);
```
tmorehouse: Ah, this is what was causing the `this$1 = this` warning

alexsasharegan: Yeah, the arrow functions are trying to capture `this`, which according to es 2015 spec at the top level of s module is undefined. Rollup has to do that manually to emulate the module environment accurately.

mosinve: Will know :)

lib/classes/popover.js
mosinve: This isnt work with current example. Function returns object, and then it stringified to `[object Object]`

lib/classes/popover.js
```diff
@@ -70,20 +70,32 @@ class PopOver extends ToolTip {
...
 
     // New methods
 
     getContent() {
-        return typeof this.$config.content === 'function' ?
-              this.$config.content.call(this.$element) :
-              this.$config.content;
+        let content = this.$config.content || '';
+        if (typeof content === 'function') {
+            content = content.call(this.$element).toString.trim();
```
mosinve: This isnt work with current example. Function returns object, and then it stringified to [object Object]

tmorehouse: This when setting content to a function, or passing a dom object?

lib/components/popover.vue
mosinve: okay. agree

lib/classes/popover.js
mosinve: when function returns object

lib/mixins/dropdown.js
alexsasharegan: The warning is still issued, but I can't find any straggler arrow functions at top-level module scope. I'm thinking it might be in popper.js itself since the warning comes right where popper is imported.

lib/classes/tooltip.js
```diff
@@ -0,0 +1,690 @@
...
+    MozTransition: ['transitionend'],
+    OTransition: ['otransitionend', 'oTransitionEnd'],
+    transition: ['transitionend']
+};
+
+// Client Side Tip ID counter for aria-describedby attribute
+// Could use Alex's uid generator util
+// Each tooltip requires a unique client side ID
+let NEXTID = 1;
+const generateId = name => `__BV_${name}_${NEXTID++}__`;
```
mosinve: Maybe this is reason of `this` warning?

lib/mixins/dropdown.js
mosinve: see lower comment )

lib/classes/tooltip.js
alexsasharegan: Yeah, that would do it!

lib/mixins/dropdown.js
tmorehouse: Just changed the `generateId` in tooltip.js class  back into a named function...

alexsasharegan: image

Just pulled, and it's still warning. It's one of those things that _should be fine_, but can potentially introduce bugs.

mosinve: where did you get that warnings? i dont have any

mosinve: So it's not arrow functions gens this warning?

mosinve: image
Also i have this error

alexsasharegan: I get it running `yarn` which does the build after install.

I was asking Troy about that Popper error the other day too. It doesn't make much sense to me... The headless browser the jest test runner uses should have something like that in it's environment.

mosinve: https://stackoverflow.com/questions/42213522/mocking-document-createrange-for-jest

alexsasharegan:
Moved the `BvEvent` into the classes directory with all its friends. 😄

lib/classes/popover.js
```diff
@@ -70,7 +70,7 @@ class PopOver extends ToolTip {
     getTitle() {
         let title = this.$config.title || '';
         if (typeof title === 'function') {
-            title = title.call(this.$element).toString().trim();
+            title = title(this.$element).toString().trim();
```
mosinve: this pass element as argument to function... while previous variant executed function with context of element. R U sure that this will work?

tmorehouse: I changed this to allow the function that produces content to maintain its own `this` context (i.e. the app context, so functions can reference app methods and data using `this`)

tmorehouse: And I need to remove the toString stuff from here

lib/mixins/dropdown.js
tmorehouse: I wonder if we can import popper at a different point, within the tooltip class? maybe in the constructor function, or win the show method (where be instantiate popper?  Or, if it is a global, then do we need to do the import at all?

lib/components/tooltip.vue
```diff
@@ -11,10 +11,13 @@
...
     import { isArray } from '../utils/array';
     import observeDom from '../utils/observe-dom';
+    import { assign } from '../utils/object';
     
     export default {
         data() {
-            toolTip: null
+            return {
+                toolTip: null
+            }
```
tmorehouse: 👍

lib/mixins/dropdown.js
mosinve: but source of warning is dropdown, not tooltip....

tmorehouse: or use a `const Popper = require('popper.js')` in the `constructor` or in the `show()` method?

tmorehouse: True...

Maybe in dropdown, we get Popper on create/show?

PR #940, opened 2017-08-28
fix(link): if to is an object and to.path exists, use to.path has href fallback

tmorehouse:


alexsasharegan:
It's this necessary? View router already does this for us I thought.

tmorehouse:
Maybe it does...  so should we even set `href` at all if `to` is provided? (or maybe always set to `'#'` ?

tmorehouse:
This used to be in the original b-link code from way back (before my time).

I think it was a fallback in case vue-router wasn't available

alexsasharegan:
Then maybe it should check for vue-router first. Maybe it could use `Boolean(parent.$router)`?

tmorehouse:
Is there a `$isRouter` prop on the vue instance?

tmorehouse:
Note... just `vm.$router`

tmorehouse:
Actually it appears that router-link ignores the href prop.

pi0:
@tmorehouse We have initially made `b-link` abstraction to make BV more portable not depending on special libraries like vue-router and also have a tight integration with the router if exists. This made many big challenges. First that router-link itself creates href too so having both on our custom component makes many problems specially with SSR renders. But i think with the The `parent.$router` guard in this PR is safe as it just happens when there is **no** router :)

tmorehouse:
Should we event set an `href` (right now defaults to `#` if `to` is provided and there is `parent.$router` defined).  Should it maybe return null or empty string?

pi0:
@tmorehouse, To be honest, can't remind the exact story and reason. But I think it was '#' to show a correct cursor on drop-down, btn etc when router is available but `to` is not specified

pi0:
Ah it's getting more complicated :)) we need a psudo like code to specify b-link's behavior 😆

tmorehouse:
maybe if `$router` is defined and `to` provided, we return `null` for HREF, and let router link generate the HREF. but if `to` is provided and `$router` not avail, return `to/to.path`,

lib/components/link.js
```diff
@@ -89,12 +89,16 @@ function computeTag(props) {
     return props.to && !props.disabled ? "router-link" : "a";
```
tmorehouse: Maybe this should check for `parent.$router` as well?

pi0: Yes i think it should too.

tmorehouse: Just need to make sure its pass `parent` as the next arg

tmorehouse:
Closing s this has been superseded by #1016

PR #941, opened 2017-08-28
feat(layout): alignment utilities 🛠

alexsasharegan:


mosinve:
whats purpose of col attribute? played out with docs and dont find one...

alexsasharegan:
`b-col` sets `.col` only when no `col-*` class is given. It's the most basic, equal width flexbox class from the BS4 grid system. The reason it sets that automatically is that plain `<b-col></b-col>` tags should do something by default. However, the moment you change that up to do this: `<b-col cols="6"></b-col>` that default behavior becomes unexpected.

mosinve:
KK, so what's the real needed use case of col prop?

alexsasharegan:
Just following the BS4 library: https://getbootstrap.com/docs/4.0/layout/grid/#equal-width

mosinve:
In case using `col-*` classes, `.col` is not needed. So we have to set `.col` class only when we have not any other modifiers. And this done by `b-col` itself. At BS examples, i think it excessive, you can check it from browser console by yourself, the reason they do it, only to show that div is col, it does not make any real work in case other `col-*` classes used.

alexsasharegan:
I'm not sure I follow. `b-col` offers out-of-the-box responsive, flexible row columns. The moment you add in a specific column width, `b-col` adapts to that and chucks the `.col` class. This feels really nice as you develop because you just start using `<b-col>`s everywhere and then add props as your responsive layout needs change.

mosinve:
I just wanted to know the purpose of col attribute of `b-col`, but already figured out it.... it's' used as fallback.

alexsasharegan:
Ah, then yeah. To be honest, the flex grid system is a little confusing. I followed their "spec" as closely as possible, and while I know how to use it well, I'm not super confident on the flex basis and all that behind it.

mosinve:
`.col-*` works when size breakpoint classes not in action to retain flex ability. 
`<b-col col lg="2">1 of 3</b-col>` Here at sizes lower than lg it fallback jus to .col and still use flexbox, otherwise, without  `.col`, at sizes lower lg, it will be just block with `width:100%`

alexsasharegan:
Exactly.

tmorehouse:
Maybe we should mention that in the docs under b-col?

Issue #944, opened 2017-08-29
Where's the usage docs?


muhajirframe:

muhajirframe:
Hi, thanks for this great project.
But I can't find the docs. I know there's a documentation on how to install it https://bootstrap-vue.js.org/docs . But how do i use it?

mosinve:
There is a left navbar with links pages describing the usage of components.
![image](https://user-images.githubusercontent.com/20986566/29831267-bfb3dd2a-8d0e-11e7-8415-ed23784424c2.png)

muhajirframe:
Ah, I am sorry. I was browsing with my phone, and can't see that. Perhaps we can't make it not hidden on small screen? Maybe put it at the bottom?

mosinve:
Probably, it's a bug in docs. Thanks for report. :)

muhajirframe:
Okay, good works btw 😊. Shall we close this issue?

mosinve:
We will close, when problem will be fixed ;)

PR #946, opened 2017-08-29
fix(docs): Side bar navigation not accessible on small screens

tmorehouse:
The sidebar is not accessible when screen are XS size.

fixes issue #944


docs/nuxt/layouts/docs.vue
```diff
@@ -5,6 +5,7 @@
         <div class="container-fluid">
             <div class="row flex-xl-nowrap2">
                 <div class="col-12 col-md-3 col-xl-2 bd-sidebar">
+                    <div class="d-block d-sm-none py-4 mt-3"></div>
```
tmorehouse: Temporary hack to push search & toggler in view on XS screens

mosinve: no positive effect

mosinve: image
toggler showed, but its not clickable

tmorehouse: ah may need to add v-b-toggle directive...

tmorehouse: My mistake.  I had component `<nav-toggle>` instead of `<b-nav-toggle>`

tmorehouse: Note search currently doesn't work from what I can see.  Its just there for when search will be applied.

I wonder if that should be hidden, or disabled?

tmorehouse:
Is the toggle working correctly now?

PR #951, opened 2017-08-29
[WIP] feat(carousel): enable mobile swipe to control next/prev

tmorehouse:
Swipe right or left to go to the next or prev slide.

User has to swipe at least 25 pixels or 1/2 slide width to trigger swipe

### DIFF
--- a/lib/components/carousel.vue
+++ b/lib/components/carousel.vue
@@ -281,10 +281,10 @@
                     if (Math.abs(dX) > Math.abs(dY)) {
                         // Direction is more horizontal
                         evt.preventDefault(); // prevent scrolling
-                        slideWidth = parseInt(window.getComputedStyle(this.$el).width,10);
+                        const slideWidth = parseInt(window.getComputedStyle(this.$el).width,10);
                         if (dX > slideWidth/2 || dX > 25) {
                             // We have moved enough to trigger a slide, so go next or prev
-                            this.[dX > 0 ? 'next' : 'prev']();
+                            this[dX > 0 ? 'next' : 'prev']();
                             // Stop reactiving to touchmove
                             this.touchSlid = true;
                         }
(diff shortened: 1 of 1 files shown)


lib/components/carousel.vue
```diff
@@ -246,6 +250,48 @@
                 }
             },
 
+            // handle swiping on mobile
```
jackmu95: I haven't tested the code yet, but I think it can be a bit more simplified like this:

```js
// Handle swiping on mobile
onSwipe(evt) {
  if (!evt.touches || evt.touches && evt.touches.length > 1) {
    return;
  }

  const type = evt.type;
  const touch = {
    x: evt.touches[0].pageX,
    y: evt.touches[0].pageY
  };

  if (type === 'touchstart') {
    // Start of possible swipe.
    this.touchData = touch;
    this.pause();
  } else if (type === 'touchmove') {
    if (!this.touchData) {
        return;
    }

    // User is moving finger. Has it moved horizontally enoough?
    const dX = touch.x - this.touchData.x;
    const dY = touch.y - this.touchData.y;

    // Is a scroll or a left/right?
    if (Math.abs(dX) > Math.abs(dY)) {
      // Direction is more horizontal
      evt.preventDefault(); // Prevent scrolling

      const slideWidth = parseInt(window.getComputedStyle(this.$el).width, 10);
      if (Math.abs(dX) > slideWidth / 2 || Math.abs(dX) > 25) {
        // We have moved enough to trigger a slide, so go next or prev
        this[dX > 0 ? 'prev' : 'next']();

        // Stop reactiving to touchmove
        this.touchData = null;
      }
    }
  } else if (type === 'touchend') {
      this.touchData = null;
  }
}
```

tmorehouse: Thanks!  this PR is a bit outdated now, but we'll be revisiting this later.

pi0:
Closing due inactivity.

PR #976, opened 2017-08-31
Allow formatter to recieve full row instead of the prop named by the …

SergioReis97:
…field if field has 'useCustomFunction'

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -555,6 +555,12 @@
                 return field.formatter && ((typeof (field.formatter) === 'function') || (typeof (field.formatter) === 'string'));
             },
             callFormatter(item, key, field) {
+                if(field.useCustomFunction){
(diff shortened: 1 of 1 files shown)


SergioReis97:
Using [this](https://jsfiddle.net/ecjcgf18/) pretty much solves my problem, this allows more flexibility into custom formatting i guess. At least in my use case would help.

mosinve:
Why u target `master` branch?  the actual release branch is `dev` 
If you'll look at source then you'll see that your proposed fix already implemented in bootstrap-vue@beta.6
https://github.com/bootstrap-vue/bootstrap-vue/blob/2777380f98f0508aff855f397f1910649a15416b/lib/components/table.vue#L622

SergioReis97:
I'm sorry, this was just my bad of pulling and pushing from master :sweat_smile: This is a way better fix than my quick fix, will pull and use this. Thanks

PR #977, opened 2017-08-31
fixing multiple modal

mirzaakhena:
I know bootstrap not supported multiple modal (yet). But who can resist someone to create multiple modal. This actually works well.

This pull request is to handle bug when we have multiple modal at the same time. The issue is occur when the second modal is hiding. The first modal is freezing and cannot scrolled because the second modal was remove the modal-open in the body class.

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -104,6 +104,22 @@
         '[tabindex]:not([disabled]):not(.disabled)'
     ].join(',');
 
+    // to handle multiple modal
+    var modalGlobalCounter = {
+      info: {
+        count: 0
+      },
+      incCount () {
+        this.info.count++;
+      },
+      decCount () {
+        this.info.count--;
+      },
+      getCount () {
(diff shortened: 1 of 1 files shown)


tmorehouse:
Modals are currently undergoing a re-write to conform closer to the Bootstrap specs, and to handle the open transitions better.

lib/components/modal.vue
```diff
@@ -104,6 +104,22 @@
         '[tabindex]:not([disabled]):not(.disabled)'
     ].join(',');
 
+    // to handle multiple modal
+    var modalGlobalCounter = {
```
tmorehouse: Note in Server Side Rendering (SSR), this global variable is shared between ALL requests.  So if a modal is set to be initially show from server side, this will affect all requests to every user of the website.

tmorehouse:
I would recommend that you make a separate multi-modal component incited of altering the standard modal.

PR #984, opened 2017-09-02
[WIP] feat: Utility directives (proposal)

tmorehouse:
Example Ideas:

- `v-b-text.*.*.*` text (variant, align, transform, nowrap)
- `v-b-bg.*` background (variant)
- `v-b-border.*.*.*` border (variant, rounding, sides)
- `v-b-clearfix` (clearfix util)
- `v-b-float.*.*` Float (left/right, breakpoint)
- `v-b-d.*.*` Display (type, breakpoint)

These mainly map to these at https://getbootstrap.com/docs/4.0/utilities/

Thoughts? Is this a silly idea? or useful?

All these would do would be to add appropriate classes to the element they are applied on.

### DIFF
--- a/lib/directives/bg.js
+++ b/lib/directives/bg.js
@@ -1,0 +1,25 @@
+/*
+ * v-b-bg directive
+ *
+ * Usage: 
+ *  Variants: v-b-bg.danger
+ */
+
+import { variants } from '../utils/enums';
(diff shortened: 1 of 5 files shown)


tmorehouse:
Canning this PR, as these only run on client side and not SSR (not without explicit enabling them in SSR).

PR #996, opened 2017-09-06
Removed ="md" from toggleable="md"

pmorong:
toggleable is expected to be a boolean so assigning it a string "md" causing a type mismatch error.

### DIFF
--- a/docs/components/navbar/README.md
+++ b/docs/components/navbar/README.md
@@ -4,7 +4,7 @@
 It’s easily extensible and thanks to our Collapse plugin, it can easily integrate responsive behaviors.
 
 ```html
-<b-navbar toggleable="md" type="dark" variant="info">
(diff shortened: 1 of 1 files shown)


tmorehouse:
In the latest release, `toggleable` can be a `Boolean` or a `String`:

https://github.com/bootstrap-vue/bootstrap-vue/blob/d765976d7286c0379f31d1f2209b787cb1704617/lib/components/navbar.js#L15-L18

Which, when used as a Boolean specifies the smallest breakpoint, and while specified as a String allows you to set the breakpoint at which the navbar will collapse (i.e. `sm`, `md`, `lg`, `xl)

tmorehouse:
Which version of Bootsrtap-Vue are you using when you see this error?

pmorong:
Ah I was only using 18, just updated to 20.2 and its working -- sorry about that.

PR #1004, opened 2017-09-06
feat(modal): Add props to change the variant of the default modal buttons

tzagorni:
This provides an easy way to have other variants than primary for Ok and secondary for Close without creating a custom modal footer


tmorehouse:
Thanks!, we will need to have documentation updated to mention the new props

PR #1024, opened 2017-09-10
feat(modal): Improve modal transitions, padding adjustments, and aditional features

tmorehouse:
Various modal improvements:

- [x] Improve modal transitions.
- [x] Remove need for scoped CSS.
- [x] Incorporate `BvEvent` object in modal events (`hide`, `hidden`, `show`, `shown`).
- [x] Incorporate scrollbar adjustments.
- [x] Add listeners for resize/orientationchange to adjust modal padding if necessary
- [x] Use observeDom util to adjust modal when content size changes.
- [x] Option to not show backdrop (although still supports close on backdrop click, when backdrop not shown).
- [x] Emit on `$root` new namespaced events.
- [x] Listen on `$root` new namedspaced events (**Minor breaking change**).
- [x] Update `v-b-modal` directive to use new namespaced `$root` event.
- [x] Add optional bg, border, and text variants (header, footer, body).
_Note: current Bootstrap V4.beta.1 CSS causes header and footer to loose rounded corners when a background variant is applied. This PR includes a fix for that using the `rounded-*` classes_
- [x] New functional component `<b-button-close>` for header close button (also added to `<b-alert>`).
- [x] New slot for header-close button content
- [x] Change footer close button text to `Cancel` to better reflect slot name and event name
- [x] More dom utils for better minification (used by modal and tooltips/popovers).
- [x] Auto append tooltip/popovers to `.modal` `<div>` (instead of `<body>`) for proper z-indexing, if container not specified on tooltip/popover (check is done within tooltip.js class)
- [x] Update docs & examples.

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -15,6 +15,7 @@
         >
             <div :class="modalClasses"
                  :id="safeId()"
+                 :aria-hidden="is_visible ? '' : 'true'"
                  role="dialog"
                  ref="modal"
                  v-show="is_visible"
@@ -106,6 +107,10 @@
         }
         return arrayFind(selectAll(selector, root), isVisible) || null;
     }
+    
+    function reflow(el) {
+        return el.offsetHeight;
+    }
 
     export default {
         mixins: [idMixin, listenOnRootMixin],
@@ -117,6 +122,7 @@
                 is_show: false,
                 is_block: false,
                 scrollbarWidth: 0,
+                isBodyOverflowing: false,
                 return_focus: this.returnFocus || null
             };
         },
@@ -258,10 +264,11 @@
                 });
                 this.emitEvent(showEvt);
                 // Show if not canceled
-                if (!showEvt.defaultPrevented) {
-                    this.is_visible = true;
-                    this.$emit('change', this.is_visible);
+                if (showEvt.defaultPrevented || this.is_visible) {
+                    return;
                 }
+                this.is_visible = true;
+                this.$emit('change', this.is_visible);
             },
             hide(trigger) {
                 if (!this.is_visible) {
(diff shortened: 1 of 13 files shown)


tmorehouse:
I think everything is now ready for the revamped modal. component

tmorehouse:
Added additional dom utils methods for better minification

tmorehouse:
OK... now I think this PR is ready for use.

jackmu95:
@tmorehouse Nice work!

alexsasharegan:
It would be nice to work all the examples (turn most into live play ex), but I think this is good now. Examples could be another pr.

lib/components/modal.vue
```diff
@@ -249,49 +249,61 @@
...
                     // Rounding is needed to fix a bug in bootstrap V4.beta.1 CSS
-                    Boolean(this.headerBgVariant) ? 'rounded-top' : '',
-                    Boolean(this.headerBgVariant) ? `bg-${this.headerBgVariant}` : '',
-                    Boolean(this.headerTextVariant) ? `text-${this.headerTextVariant}` : '',
-                    Boolean(this.headerBorderVariant) ? `border-${this.headerBorderVariant}` : ''
+                    {
+                        'rounded-top': this.headerBgVariant,
+                        [`bg-${this.headerBgVariant}`]: this.headerBgVariant,
+                        [`text-${this.headerTextVariant}`]: this.headerTextVariant,
+                        [`border-${this.headerBorderVariant}`]: this.headerBorderVariant,
```
tmorehouse: Should these be wrapping the variants in `Boolean(...)` (since they are strings)

tmorehouse: i.e
```js
[`bg-${this.headerBgVariant}`]: Boolean(this.headerBgVariant),
```

tmorehouse: And similar for the other classes sections as well...

alexsasharegan: Like we discussed, Vue checks these in an `if` condition, so it isn't necessary. You're thought and them getting cached in the computed props is smart though, so best to cast them.

Issue #1028, opened 2017-09-11
Module not found: Error: Can't resolve 'lodash.startcase'


michalbujalski:
I've added BootstrapVue to my project. When trying to build i keep getting:
`Module not found: Error: Can't resolve 'lodash.startcase` error.

followed by:
` @ ./~/babel-loader/lib!./~/vue-loader/lib/selector.js?type=script&index=0!./~/bootstrap-vue/lib/components/table.vue 91:0-41
 @ ./~/bootstrap-vue/lib/components/table.vue
 @ ./~/bootstrap-vue/lib/components/index.js`

I'm using https://github.com/egoist/vuepack

tmorehouse:
PR #1030 may address this issue.

greg-harrison:
I'm seeing this issue, as well, even after it was addressed in the latest Beta release.

I notice the Lodash and Lodash.startcase are in the DevDependencies, what's the reasoning there?

tmorehouse:
lodash is used by the docs, while lodash.startcase is used by b-table.  lodash.startcase has been added as a dependancies in PR #1030, which will be available in the next release (v1.0.0.beta.8)

PR #1030, opened 2017-09-11
fix: Add lodash.startcase to dependancies

tmorehouse:
`b-table` requires `lodash.startcase`

Re: issue #1028


tmorehouse:
@pi0, not sure if there are other locations that we need to alter this (maybe in rollup cnofig)?

pi0:
@tmorehouse That problem shouldn't happen in normal cases (tiny startcase is inlined in dist bundles) unless users try to build `.vue` files. I've merged it.

PR #1039, opened 2017-09-12
Revert "fix(observedom): Callback not being called for changes other than childList changes"

tmorehouse:
Reverts bootstrap-vue/bootstrap-vue#1025


tmorehouse:
This appears to be causing a time out issue with modal.

Reverting for now, and will revisit.

PR #1067, opened 2017-09-16
chore(docs): dynamic links to alpha or dev branch source files 🔗

alexsasharegan:
Links to the appropriate source file branch based on client url.


alexsasharegan:
I guess I would need to PR this to alpha too, wouldn't I?

tmorehouse:
Yeah, probably so...  Or just hardcode the link base URL in each branch.

alexsasharegan:
I don't suppose there's any way to read the version tag dynamically and reference an exact commit hash. Then you could see the version source code and not bleeding edge code.

tmorehouse:
Hmmm not sure.

tmorehouse:
Added in detection of component source extension (`.js` or `.vue`)

pi0:
Alpha docs has a completely different structure from the current state. So I think it would be better adding multi-version docs for next releases.

PR #1094, opened 2017-09-22
fix(tooltip): Emit correct $root event name

jacobgunnarsson:
I was wondering why `this.$root.$emit('bv::hide::tooltip')` wasn't working and found this typo.


tmorehouse:
👍 Thanks for finding this!

Will be available in the next release

PR #1095, opened 2017-09-22
Update ToolTip content when config.title changes in updateConfig

jacobgunnarsson:
I had to support a use case where the tooltip content needs to update when the `config.title` binding changes (and the tooltip is already instantiated). This seems to do the trick.

### DIFF
--- a/lib/classes/tooltip.js
+++ b/lib/classes/tooltip.js
@@ -166,6 +166,21 @@ class ToolTip {
         // Stop/Restart listening
         this.unListen();
         this.listen();
+
+        if (typeof updatedConfig.title === 'undefined') {
+            // We've recieved a new config without a title, hide the tooltip
+            this.hide();
+        } else if (prevConfig.title !== updatedConfig.title) {
+            // We've recieved a new title that differs from the previous one,
+            // set the new element content and update positioning
+            this.setElementContent(select(Selector.TOOLTIP_INNER, this.$tip), this.getTitle());
+            this.update();
+
+            if (typeof prevConfig.title === 'undefined') {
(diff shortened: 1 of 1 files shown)


lib/classes/tooltip.js
```diff
@@ -166,6 +166,21 @@ class ToolTip {
...
+            this.hide();
+        } else if (prevConfig.title !== updatedConfig.title) {
+            // We've recieved a new title that differs from the previous one,
+            // set the new element content and update positioning
+            this.setElementContent(select(Selector.TOOLTIP_INNER, this.$tip), this.getTitle());
+            this.update();
+
+            if (typeof prevConfig.title === 'undefined') {
+                // There was no previous title set, we also need to show the tooltip
+                this.show();
```
tmorehouse: This might end up showing a tooltip (or popover, as popover uses this same base class) when it shouldn't be, as well as it doesn't set a trigger type.

The design intent was to only update the title when hidden and re-shown - except in the case of the popover/tooltip components, which allow for reactive content.

tmorehouse: Also, popover also has a content (besides title), and this doesn't address this.  Content and title can be a string, a function, or an element reference.

I think there might be too much breaking change in this case, and may have unexpected side effects in both tooltips and popovers.

tmorehouse:
Your best bet is to use the component version of tooltip, which allows for dynamic update of the title while the tip is open (when using the default slot).

The current implementation (design intent) is only to update the title (from the title attribute or config.title when it is a string) is when the tip closes and reopens.  The tooltip is rebuilt each time it opens.

jacobgunnarsson:
Gotcha, cheers!

PR #1129, opened 2017-09-28
feat(form-group): Switch to fieldset+legend for better semantic/ARIA markup

tmorehouse:
Fieldset + label automatically associates with any containing input element.

Also provides better ARIA support when there are multiple inputs (i.e. radio's and checkboxes) in a form-group

### DIFF
--- a/docs/components/form-group/README.md
+++ b/docs/components/form-group/README.md
@@ -1,8 +1,8 @@
 # Form group
 
 > The `<b-form-group>` component is the easiest way to add some structure to forms. Its
-purpose is to provide a label and control pairing, help text and feedback text, as well
-as contextual state visual feedback.
+purpose is to provide a pairing between controls and a label, help text and feedback text,
+as well as contextual state visual feedback.
 
 <div class="alert alert-warning">
  <p class="my-0">
@@ -49,11 +49,11 @@ export default {
 ```
 
 ## Label
-Use the prop `label` to set the content of the generated `<label>` element (html supported),
-or by using the named slot `label`, You may optionally visually hide the label by setting
-the prop `label-sr-only`.
+Use the prop `label` to set the content of the `<fieldset>` generated `<legend>`
+element (html supported), or by using the named slot `label`, You may optionally
+visually hide the label text by setting the prop `label-sr-only`.
 
-By default, the label appears above the input element, but you may optionally set
+By default, the label appears above the input element(s), but you may optionally set
 the prop `horizontal` to place the label on the same line, and control the width
 of the label by setting `label-cols` to the number of columns (default of `3`,
 valid range of 1 through 11). `label-cols` has no effect if the layout is
@@ -70,13 +70,13 @@ Optional descriptive text which is always shown with the `.text-muted` class
 (html supported) by setting the `description` prop or using the named slot `description`.
 The description text is rendered using the <`b-form-text>` component.
 
-## Feedback
+## Invalid feedback
 Show optional feedback text to provide textual state feedback (html supported)
 by setting the prop `feedback` or using the named slot `feedback`.
 
-Note that the feedback **will not be shown** unless the `invalid` state is set on the
-`<b-form-group>` and it's child(ren) input(s) or just on the input (`b-form-input>`, `b-form-textarea>`, `<b-form-select>`,
-`<b-form-checkbox>`, `<b-form-radio>`, or `<b-form-file>`).
+Note that the feedback **will not be shown** unless the invalid `state` is set on the
+`<b-form-group>` and it's child(ren) input(s) or just on the input (`b-form-input>`,
+`b-form-textarea>`, `<b-form-select>`, `<b-form-checkbox>`, `<b-form-radio>`, or `<b-form-file>`).
 
 Also feedback will be shown if the parent `<b-form>` component does not have the
 `novalidate` prop set (or set to `false`) along with `vadidated`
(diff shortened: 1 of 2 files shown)


tornography:
Colud you please revert this change, because in my opinion this is kind of breaking the internet.

If you remove the label, that corresponds with an input, your "win" of ARIA is obsolete. I want to click on a label to set focus on the input.
Additionally I dont want every label+input beeing wrapped in a fieldset. I can understand this is wanted with groups of radios or checkboxes. But not with single text inputs.

Please stick to the original bootstrap version. https://getbootstrap.com/docs/4.0/components/forms/

tmorehouse:
The problem is that the label is not easily associated with input (especially if there are multiple inputs). which is what triggers the focusing of the input.

How about if a click handler was added to the legend, which then finds the first input within the content, and if one found, it focuses it.

tornography:
Ok, I see the issue.
For things like that we use a unique/random like id/hash that's generated for each label + input pair.
(not for radios/checkboxes, as they are wrapped inside the label)

The second thing is, that the input is kind of described by the label, which mast be manually added tuo the input, pointing to the legend.

Third, still every input now has its own fieldset, what is not what one might want.
It's more likely to wrap a set of inputs (like you said with the checkboxes/radios) to group them semantically.

So my suggestion is to separate two components:
1. b-form-group => with `label + input` grouped in a div.form-group
2. b-formfield => to wrap sets of inputs semantically


Maybe off topic in this issue: Is it necessary to wrap each form group in rows and cols?
I see the use when having a horizontal label/input layout. 
But it's bloating code when having a useless `row > col-12`.

tmorehouse:
`b-form-group` doesn't limit one to placing a single input inside it (i.e. you could have multiple text fields representing an address, which is labeled by the fieldset legend).

For semantics, the legend describes the content of the fieldset, much in the same way a label can describe an input, but without the need to use the `for` attribute (and `for` can only point to a _single_ input element)

We might be able to alter the render function to not wrap the content in rows/cols when not in horizontal mode.

tmorehouse:
How about this compromise:

If we add a prop `label-for`, and one passes an input ID to the `label-for` prop, we generate a `<label>` element instead of a `<legend>` element?

tmorehouse:
@tornography 

PR #1423 adds in the `label-for` prop.  If specified, `<b-form-group>` will render  a `<label for={id}">` element.  It assumes the ID provided exists on the input that is inside the `form-group`.

If an ID is not passed to `label-for`, then the default `<legend>` will be rendered with a click handler that focuses the first non-disabled visible input found in the form-group (emulating the label click, when a label has a `for` attribute.

I think this is a good compromise. If we were to auto 'for' the label, then we would need to add in additional DOM change observers, etc, for picking  the first input ID, And adding in additional code for dealing with multiple inputs, etc, which would end up bloating the component code.

The 'div' wrapper around the input(s) is needed for grouping/alignment purpose (for the input+feedback+description), and to provide a container $refs, so it is not something we can remove easily without increasing the code size.

tornography:
This compromise is something I can live with, or at least one can work with. Thank you.

But I'm a little afraid, that users could get confused when using `b-form-group` expecting a `form-group` and getting a `fieldset + legend`. This one component servers two different purposes. And "hiding" this switch behind a `label-for` does not help to tell what is is rendered.

So another idea: `<b-form-group variant="fieldset">` so the user can explicitly choose to use fieldset.

Thanks for getting back to me with your quick responses.

tmorehouse:
The documentation quite clearly explains what is used (i.e. fieldset and legend) in `b-form-group`,  see: https://bootstrap-vue.js.org/docs/components/form-group#label

So, tell me this... without the 'label-for` switch, how are you going to associate the label with the input?

The label, if it has no `for` attribute, **will never be associated** with the input (as it doesn't wrap the input).  Clicking on the `label` (if we just has label instead of legend) will not activate the input.  So your point is moot.

So because of this, we chose to use fieldset and legend to provide ARIA accessibility (i.e. for sight-impaired users, as a alegend inside a field set is always associated with the contents (and is spoken out as so by screen readers).

tornography:
I don't want to omit the `label-for`. That's absolutely necessary. Just wanted to point out, that the code `<b-form-group label-for="id">` might not indicate that the group is rendered differently. That's why I thought an explicit property could make this clear in the code. Like
`<b-form-group label-for="id">` for label, vs. `<b-form-group variation="fieldset">`.
I find it strange to switch expected behavior and just pointing to the documentation.

But I guess we can use it is now. I just need to get everyone in the company aware of this.

Thanks.

tmorehouse:
If you look at PR #1423, you will notice the documentation update included, which explains this.

`variation`, as a prop name, does not explain what `variation` is, and would be confused with `variant` most likely.

If someone doesn't specify a `label-for` prop, then the label ** will never be associated with an input**, and becomes useless.  Legend on the other hand, does by nature of markup provide better ARIA accessibility than a dangling label with no association.

tornography:
Oh, ok. I guess I did't make myself clear. Sorry. I'll try by giving extended examples.

```
<b-form-group label-for="id" label="text">
  <b-form-input />
</b-form-group>
```
 would render
```
<div class="form-group">
  <label for="id">text</label>
  <input id="id" />
</div>
```
So `label-for` is need. As you said.
Now if we don't want a `<label>` we would need to use a boolean to switch to fieldset element, instead of checking if `label-for` is set. 
```
<b-form-group fieldset="true" label="one, two">
  <b-form-input />
  <b-form-input />
</b-form-group>
```
 would render
```
<fieldset class="form-group" label="text">
  <legend>text</legend>
  <input />
  <input />
</fieldset>
```
The logic is the same, but the user is more aware of his choice. But ether way, the checking the existence of `label-for` to switch will work fine. So all is good.

tmorehouse:
Your inputs are missing IDs, and hence **the label will not be associated to the input**

setting `label-for` **will not automatically ID the inputs**.

tornography:
Oh, sure. Updated the example and added the id. Of course the id is not generated automatically.
And you're right, the fallback to fielset will still be accessible. I'm ok with the solution.

Thanks again for your feedback and changes.

PR #1182, opened 2017-10-09
fix(modal): Modal jumps when dialog height changes

DabeDotCom:
Building on #1058...  If the modal dialog's content height changed (e.g., "v-if" conditionally displayed an error message), the `padding-right` could be applied inconsistently.

twbs/bootstrap#23672 fixed this by calling `this.adjustDialog()` during `onBeforeEnter()`, as well

### DIFF
--- a/lib/components/modal.vue
+++ b/lib/components/modal.vue
@@ -387,6 +387,7 @@
                 this.is_transitioning = true;
                 this.checkScrollbar();
                 this.setScrollbar();
+                this.adjustDialog();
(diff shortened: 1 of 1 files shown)


tmorehouse:
The DOM observer should catch most changes in modal content (i.e. sections being conditionally hidden/shown), although this observation only occurs after the modal has opened (and not to mention that internet explorer sucks when it comes to DOM observation).

But this is a good addition.  Thanks!

Issue #1195, opened 2017-10-12
Server side rendered b-modal causes runtime error


JamieLottering:
Using Nuxt with our app and some pages simply output a modal, as seen with #1167 .

These pages render modals on the server side and default to `visible` as such:

```
<b-modal visible>
   Some content
</b-modal>
```

However this causes a runtime error in dev because the server side html doesn't match the client side html. It looks like it may be caused by the close button in the header.

This is a screenshot of the error.
![screenshot](https://www.dropbox.com/s/jth1rkt6tb5l8cc/Screen%20Shot%202017-10-12%20at%203.23.11%20PM.png?raw=1)

Now of course wrapping these in a `<no-ssr />` tag works fine but this is not ideal as I'd like the HTML for these to be visible in the initial source.

I'm not sure if this bug is specific to bootstrap-vue or an error in Vue itself.

JamieLottering:
FWIW If I add the `hide-header-close` prop on the `<b-modal />` it renders correctly without an error. But this is not the behavior I want.

In addition, manually putting a 
`<button type="button" name="button">&times;</button>` in the modal body slot it doesn't experience this client/server html mismatch so it must be something in the `<b-modal />` itself causing this.

tmorehouse:
This has been fixed in the latest dev branch already.

JamieLottering:
I’m curious what the fix was for that then. I still see the hard coded x symbol in button-close on dev.

JamieLottering:
Never mind. Just saw your other comment. Thanks!!

PR #1196, opened 2017-10-12
[fix] Fix server-side / client side html mismatch caused by button-close.js

JamieLottering:
This should fix the mismatch in server/client HTML by using an HTML entity.

As seen in #1195

### DIFF
--- a/lib/components/button-close.js
+++ b/lib/components/button-close.js
@@ -41,7 +41,7 @@ export default {
         };
         // Careful not to override the slot with innerHTML
         if (!children.length) {
-            componentData.domProps = { innerHTML: "×" };
+            componentData.domProps = { innerHTML: "&times;" };
         }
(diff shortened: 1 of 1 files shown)


lib/components/button-close.js
```diff
@@ -41,7 +41,7 @@ export default {
         };
         // Careful not to override the slot with innerHTML
         if (!children.length) {
-            componentData.domProps = { innerHTML: "×" };
+            componentData.domProps = { innerHTML: "&times;" };
```
tmorehouse: If you look at the history for this file, we were originally inserting `&times;`, which ended up causing the SSR error.

Changing it to the actual UTF-* character `x` works.

JamieLottering:
Ok that would make sense then. Just need to wait for the new release tag

PR #1197, opened 2017-10-13
feat(dropdown): optionally hide the dropdown toggle caret

jon-walton:
Hi

This allows the dropdown toggle to be hidden when the dropdown is not set to split. Useful for when the dropdown is just an icon.

usage

```html
            <b-dropdown variant="link" hide-toggle size="sm">
              <template slot="button-content">
                <icon icon="avatar"></icon>
              </template>

              <b-dropdown-item @click="$store.dispatch('logout')">Logout</b-dropdown-item>
            </b-dropdown>
```

thanks!

### DIFF
--- a/docs/components/dropdown/README.md
+++ b/docs/components/dropdown/README.md
@@ -188,6 +188,26 @@ Create a split dropdown button, where the left button provides standard
 <!-- dropdown-split.vue -->
 ```
 
+## Hidden Toggle
+The dropdown can be created with the toggle hidden by setting the `hide-toggle` prop to `true`. This is useful when the dropdown is to be displayed as an icon.
+
+**Note:** The toggle will always be shown when using `split`
+
+```html
+<div>
+  <b-dropdown variant="link" size="sm" hide-toggle>
+    <template slot="button-content">
+      <icon icon="avatar"></icon>
+    </template>
+
+    <b-dropdown-item href="#">Action</b-dropdown-item>
+    <b-dropdown-item href="#">Another action</b-dropdown-item>
+    <b-dropdown-item href="#">Something else here...</b-dropdown-item>
+  </b-dropdown>
+</div>
+
+<!-- dropdown-hidden-toggle.vue -->
+```
 
 ## Sizing
 Dropdowns work with trigger buttons of all sizes, including default and split
--- a/docs/components/dropdown/README.md
+++ b/docs/components/dropdown/README.md
@@ -189,7 +189,8 @@ Create a split dropdown button, where the left button provides standard
 ```
 
 ## Hidden Caret
-The dropdown can be created with the caret hidden by setting the `no-caret` prop to `true`. This is useful when the dropdown is to be displayed as an icon.
+The dropdown can be created with the caret hidden by setting the `no-caret` prop to `true`.
+This is useful when the dropdown is to be displayed as an icon.
 
 **Note:** The caret will always be shown when using `split`
 
@@ -197,7 +198,7 @@ The dropdown can be created with the caret hidden by setting the `no-caret` prop
 <div>
   <b-dropdown variant="link" size="lg" no-caret>
     <template slot="button-content">
-      &#x1f50d;
+      &#x1f50d;<span class="sr-only">Search</span>
     </template>
 
     <b-dropdown-item href="#">Action</b-dropdown-item>
--- a/docs/components/dropdown/README.md
+++ b/docs/components/dropdown/README.md
@@ -188,14 +188,14 @@ Create a split dropdown button, where the left button provides standard
 <!-- dropdown-split.vue -->
 ```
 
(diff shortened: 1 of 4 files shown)


tmorehouse:
Could you provide updated documentation for this feature?

jon-walton:
i wasn't sure where to put it, how is the props documentation created? or should it go in the dropdown/README.md somewhere?

thanks

jon-walton:
i saw what you meant now 😄 docs updated :+1:

jon-walton:
oh, i'll get the tests updated too

tests/components/dropdown.spec.js
```diff
@@ -28,7 +28,7 @@ describe("dropdown", async () => {
         // Without async iterators, just use a for loop.
         for (let i = 0; i < dds.length; i++) {
             Array.from(dds[i].children)
-                .find(node => node.tagName === "BUTTON" && node.classList.contains("dropdown-toggle"))
+                .find(node => node.tagName === "BUTTON" && node.id.includes("_BV_toggle_"))
```
jon-walton: I had to change this to look for the id instead of the class name due to the new `hide-toggle` prop. i'm not familiar with how bootstrap vue generates the id so hopefully this is suitable?

tmorehouse: Here's how it is done:

If an ID is passed to the `id` prop of the `<b-dropdown>`, then that ID is used as the base and `__BV_toggle_` is appended.  If no ID is provided, then Bootstrap-Vue (on client only), we use the Vue unique UID (each vue VM is assigned a unique instance I on the client) to generate an ID string and `__BV_toggle_`is ppended to that.  The ID generation is handled by the id mixin's `safeId()` method.

docs/components/dropdown/README.md
```diff
@@ -188,6 +188,26 @@ Create a split dropdown button, where the left button provides standard
 <!-- dropdown-split.vue -->
 ```
 
+## Hidden Toggle
```
tmorehouse: Technically this isn't really a hide toggle, as toggle refers to the button that triggers the dopdown.

This probably should be called `hide-caret`, as removing the class `dropdown-toggle` stops the caret from being rendered (which is generated by CSS pseudo elements to add the dropdown/dropup caret)

lib/components/dropdown.vue
```diff
@@ -70,6 +69,10 @@
                 type: String,
                 default: null
             },
+            hideToggle: {
```
tmorehouse: Can we change this to `hide-caret` ?

tmorehouse: Yeah, `no-caret` would be good

docs/components/dropdown/README.md
tmorehouse: Actually as @mosinve suggested, `no-caret` would be better as the prop name.

docs/components/dropdown/README.md
```diff
@@ -188,6 +188,26 @@ Create a split dropdown button, where the left button provides standard
...
+## Hidden Caret
+The dropdown can be created with the caret hidden by setting the `no-caret` prop to `true`. This is useful when the dropdown is to be displayed as an icon.
+
+**Note:** The caret will always be shown when using `split`
+
+```html
+<div>
+  <b-dropdown variant="link" size="sm" no-caret>
+    <template slot="button-content">
+      <icon icon="avatar"></icon>
```
jon-walton: will using `<icon />` here cause issues with the doc generation due to the component not being registered?

jon-walton:
@tmorehouse hopefully i've addressed your comments. I pushed as another small commit, let me know when/if you want them squashed.

just one last question from me above regarding the use of `<icon />` in the doc example

docs/components/dropdown/README.md
tmorehouse: It won't cause an issue, but it will not show up. 

Since this is a live example (any html section that ends with a comment `<!-- some-name.vue -->` turns into a live example on the docs) it won't render correctly.

I would suggest just using a word, or an HTML entity code for a symbol, in place of the icon component (the docs to not use any icon libraries at the moment)

docs/components/dropdown/README.md
```diff
@@ -195,9 +195,9 @@ The dropdown can be created with the caret hidden by setting the `no-caret` prop
 
 ```html
 <div>
-  <b-dropdown variant="link" size="sm" no-caret>
+  <b-dropdown variant="link" size="lg" no-caret>
     <template slot="button-content">
-      <icon icon="avatar"></icon>
+      &#x1f50d;
```
tmorehouse: 👍

docs/components/dropdown/README.md
jon-walton: ok thanks. updated to be a unicode emoji icon (magnifying glass)

Issue #1121, opened 2017-09-27
Problem with sub dependencies after doing npm install


art-mironoff:
Hi guys,
I have a problem. I'm doing `npm install` and after I try to start the project with `npm start` I get the following error:
```
Failed to compile with 2 errors                                                                                                                                          00:08:40

These dependencies were not found:

* vue-functional-data-merge/dist/lib.common.js in ./~/bootstrap-vue/lib/utils/index.js
* popper.js in ./~/bootstrap-vue/lib/mixins/dropdown.js

To install them, you can run: npm install --save vue-functional-data-merge/dist/lib.common.js popper.js
> Listening at http://localhost:8080
```

After the investigation, I found that the `node_modules/bootstrap-vue/node_modules` folder does not exist. If I go to the `node_modules/bootstrap-vue` folder and do `npm install` in this folder, I get an error: 
```
sh: 1: build/build: not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! bootstrap-vue@1.0.0-beta.9 build: `build/build`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the bootstrap-vue@1.0.0-beta.9 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```
This error is not critical and the project can be started with `npm start`.

Can someone help me, why are sub dependencies not installed without manual installation?

My package.json dependencies:
```
"dependencies": {
  "bootstrap": "4.0.0-beta",
  "bootstrap-vue": "^1.0.0-beta.9",
  "chart.js": "^2.7.0",
  "vue": "^2.4.2",
  "vue-chartjs": "^2.8.7",
  "vue-moment": "^2.0.2",
  "vue-router": "^2.7.0"
}
```

My Node version: v8.5.0
My NPM version: 5.4.2
My OS: Linux Mint

alexsasharegan:
Are you compiling from source? These dependencies are included in the build files. If you want to compile from the source files, you'll have to install the dependencies mentioned.

```bash
npm i vue-functional-data-merge popper.js
```

art-mironoff:
@alexsasharegan Problem was in direct importing for each component

```
import { bContainer, bRow, bCol } from 'bootstrap-vue/lib/components'
```

After removing this row all began work perfect. Thanks.

PR #1204, opened 2017-10-17
Fix dependencies to import components individually

tabe1hands:
Issue #1121 should be addressed. I'd like to import components individually because I'd like to reduce my package's size with dynamic importing. So, I created this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,10 @@
     "vue": "^2.4.2"
   },
   "dependencies": {
-    "bootstrap": "^4.0.0-beta"
+    "bootstrap": "^4.0.0-beta",
+    "lodash.startcase": "^4.4.0",
+    "popper.js": "^1.12.5",
+    "vue-functional-data-merge": "^1.0.6"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,7 +62,10 @@
     "vue": "^2.4.2"
   },
   "dependencies": {
-    "bootstrap": "^4.0.0-beta"
+    "bootstrap": "^4.0.0-beta",
```
tmorehouse: The only issue about having Bootstrap placed here as a dependancy is that Bootstrap Vue will use the official Bootstrap v4 CSS as is.

If one wants to use a customized Bootstrap CSS in their project, Bootstrap-Vue may not use it their custom compiled CSS.

tmorehouse:
@pi0 @alexsasharegan 

should Bootstrap be listed in the dependencies?  Or should it be a peer dependency or moved back to dev dependancies?, We don;'t actually use anything directly from Bootstrap (other than in the docs)

PR #1217, opened 2017-10-21
fix(Progress): Bootstrap V4.beta.2 CSS height prop change

### LINKED ISSUE #1216: b-progress should apply the heigth prop to itself
Otherwise, it keeps its `1rem` default value.
https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/components/progress.vue

mosinve:
Move height prop from progress-bar to progress according to current Bootstrap4-beta2.
fixes #1216

### DIFF
--- a/lib/components/progress-bar.vue
+++ b/lib/components/progress-bar.vue
@@ -27,7 +27,7 @@
             },
             progressBarStyles() {
                 return {
-                    width: (100 * (this.value / this.computedMax)) + '%',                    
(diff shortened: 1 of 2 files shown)


tmorehouse:
Will need to adjust b-progress-bar as it also sets it's own height based on parent b-progress height value.

Basically just need to remove it as new CSS sets height === 100% on the progress-bar

PR #1222, opened 2017-10-22
feat(table): BS V4.beta.2 New responsive breakpoints and table-dark class

tmorehouse:
Bootstrap-V4-beta.2 adds responsive breakpoint support (`.table-responsive-{sm|md|lg|xl}`), and changes class `table-inverse` to `table-dark` for "inverse" tables.

For backwards compatibility, the `inverse` prop is still available, but the prop `dark` deprecates `inverse`.

Addresses issue #747

### DIFF
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -234,9 +234,15 @@ export default {
             default: false
         },
         dark: {
-            // Replaces prop `inverse`
             type: Boolean,
-            default: false
+            default() {
+                if (this && typeof this.inverse === 'boolean') {
+                    // Deprecate inverse
+                    warn("b-table: prop 'inverse' has been deprecated. Use 'dark' instead");
+                    return this.bark;
+                }
+                return false;
+            }
         },
         inverse: {
             // Deprecated in v1.0.0.beta.10 in favor of `dark`
@@ -413,13 +419,6 @@ export default {
         });
     },
     computed: {
-        computedDark() {
-            if (typeof this.inverse === 'boolean') {
-                warn("b-table: prop 'inverse' is deprecated. Please use prop 'dark'.");
-                return this.inverse;
-            }
-            return this.dark;
-        },
         tableClasses() {
             const responsive = this.responsive === '' ? true : this.responsive;
             return [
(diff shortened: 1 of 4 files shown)


tmorehouse:
Should the `inverse` prop be deprecated and replaced with a `dark` prop, to better reflect the new class/wording in Bootstrap official?

mosinve:
I think we could add deprecation notice in docs and at runtime as usual we do with deprecated things, and after a pair of versions delete it.

lib/components/table.vue
```diff
@@ -407,14 +413,22 @@ export default {
         });
     },
     computed: {
+        computedDark() {
```
mosinve: At carousel-slide more laconic deprecation procedure at `img-src` and `src` props definition. And without extra variable.

mosinve: something like this

```
dark: {
            // Replaces prop `inverse`
            type: Boolean,
            default() {
                if (this && this.inverse !== null) {
                    // Deprecate inverse
                    warn("b-table: prop 'inverse' has been deprecated. Use 'dark' instead");
                    return this.inverse;
                }
                return false;
            }
        },
        inverse: {
            // Deprecated in v1.0.0.beta.10 in favor of `dark`
            type: Boolean,
            default: null
        },
```

lib/components/table.vue
```diff
@@ -148,7 +148,7 @@ function recToString(obj) {
 
     return toString(keys(obj).reduce((o, k) => {
         // Ignore fields that start with _
-        if (!/^_/.test(k)) {
+        if (!'/^_/'.test(k)) {
```
tmorehouse: Can't have quotes around a literal regex

mosinve: Yep, my bad english...  eslint recommended parens not quotes ))

lib/components/table.vue
```diff
@@ -148,7 +148,7 @@ function recToString(obj) {
 
     return toString(keys(obj).reduce((o, k) => {
         // Ignore fields that start with _
-        if (!/^_/.test(k)) {
+        if (!(/^_/).test(k)) {
```
tmorehouse: Literal RegExpr shouldn't need wrapping in brackets.

If ES Lint is complaining, then try this:
```js
if (!(/^_/.test(k)))
```

mosinve: it works good with this `if (!(/^_/).test(k)) {`

mosinve:
As for me, it good now.

PR #1224, opened 2017-10-22
fix(dropdown): Bootstrap V4.beta.2 now has better hover/focus styling

tmorehouse:
Removed custom CSS for focus/hover styling as Bootstrap V4.beta.2 now has better hover/focus styling, although still not ideal.

### DIFF
--- a/lib/components/dropdown.vue
+++ b/lib/components/dropdown.vue
@@ -98,14 +98,3 @@
         }
     };
 </script>
-
-<style>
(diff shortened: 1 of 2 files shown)


pi0:
YEAH Finally 😍

tmorehouse:
The new BS V4 CSS is not perfect.  the hover state is still very light, but they now enabled the "blue box" focus ring styling (same style for button and link versions of dropdown items)

PR #1228, opened 2017-10-23
fix(package.json): Move Bootstrap back to dev deps and update popper.js version

tmorehouse:
As bootstrap is only needed to build the documentation site, move it back to `devDependencies` from `dependencies`, as the final bootstrap-vue JavaScript dist files import nothing from Bootstrap.

We should update the getting started docs to use bootstrap v4.0.0-beta.2, and possibly a few of the examples too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "vue": "^2.4.2"
   },
   "dependencies": {
-    "bootstrap": "^4.0.0-beta.2",
     "lodash.startcase": "^4.4.0",
     "popper.js": "^1.12.5",
     "vue-functional-data-merge": "^1.0.6"
@@ -71,6 +70,7 @@
     "@nuxtjs/pwa": "^0.2.1",
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-istanbul": "^4.1.5",
+    "bootstrap": "^4.0.0-beta.2",
     "clean-css": "^4.1.9",
     "codecov": "^2.3.0",
     "codemirror": "^5.30.0",
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   },
   "dependencies": {
     "lodash.startcase": "^4.4.0",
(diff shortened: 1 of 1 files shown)


pi0:
Hmm but there is a fact that users usually use bootstrap sass/css files from it's npm package. So they need adding another dependency. I'm totally OK with this change. Maybe would be better if we change setup docs to something like this: 

```bash
npm install bootstrap bootstrap-vue
```

pi0:
Also as it is somehow a BREAKING CHANGE, we should land it in real 1.0 not rcs :)) Maybe we can make it with next release ;)

tmorehouse:
Yeah..

In the current v1.0.0-beta.9 release, BS is in the devDeps, and not dependencies, so it's technically not a breaking change.

tmorehouse:
Maybe we should update the getting started docs to show how bootstrap and bootstrap-vue should be loaded into their project.

mosinve:
Currently we have this:
```bash
# With NPM:
npm i bootstrap-vue bootstrap@4.0.0-beta
# With Yarn:
yarn add bootstrap-vue bootstrap@4.0.0-beta
```
Is it fundamentally differs from what @pi0 offered?

tmorehouse:
That should be updated to:

```bash
# With NPM:
npm i bootstrap-vue bootstrap@4.0.0-beta.2
# With Yarn:
yarn add bootstrap-vue bootstrap@4.0.0-beta.2
```

for release V1.0.0..beta.10

mosinve:
ah, yep.
also, text about whitelisting `node_modules/bootstrap_vue` at `webpack.conf`
```
                    path.resolve(__dirname, "src"), // white-list your app source files
                    require.resolve("bootstrap-vue"), // white-list bootstrap-vue

```
I, generally, use just `resolve("bootstrap-vue")`. ` require.resolve` -  is from webpack 1

tmorehouse:
Bumped the popper.js version as well.

PR #1231, opened 2017-10-23
feat(navbar-nav): New b-navbar-nav component

tmorehouse:
The class 'nav' is no longer used in BS V4.beta.2 when the nav is a navbar-nav.

Introduce a new lightweight functional `<b-navbar-nav>` component.

Deprecate `is-nav-bar` prop on `<b-nav>`

### DIFF
--- a/docs/components/navbar/meta.json
+++ b/docs/components/navbar/meta.json
@@ -2,11 +2,12 @@
   "title": "Navbar",
   "component": "bNavbar",
   "components": [
-    "bNavItemDropdown",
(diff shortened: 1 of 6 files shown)


docs/components/navbar/meta.json
```diff
@@ -3,11 +3,11 @@
   "component": "bNavbar",
   "components": [
     "bNavbarNav",
-    "bNavbarBrand"
+    "bNavbarBrand",
```
tmorehouse: 👍

PR #1246, opened 2017-10-26
feat(modal): Add Bootstrap V4 anticipated verticaly centered modal

tmorehouse:
Vertically centered modals will be coming in Bootstrap V4

https://github.com/twbs/bootstrap/pull/24510

This adds a new prop `centered` which renders the modal centered vertically.

Note vertically centered modals are **not** suited for tall content modals, as there is a potential for the header (or top part of the modal) to become inaccessible.  They are better suited to one or two line content (i.e. confirmation dialogs).

The custom CSS in modal.vue can be removed once V4.beta.3 (or later) arrives

### DIFF
--- a/docs/components/modal/README.md
+++ b/docs/components/modal/README.md
@@ -270,10 +270,36 @@ breakpoints to avoid horizontal scrollbars on narrower viewports. Valid optional
 
 <!-- modal-sizes.vue -->
 ```
+
+
+## Vertically centering
+Vertically center your modal in the viewport by setting the `center` prop.
+
+```html
+<div>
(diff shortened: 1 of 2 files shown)


tmorehouse:
Ive also made a proposal to twbs/BSV4 on handling viewport overflowing  vertically centered modals.  So maybe once V4.beta.3 rolls around, we can remove the warning in the docs.

mosinve:
Cool.. So we are a step ahead official BS4? )))

tmorehouse:
Yep, 1/2 step ahead 🙂

Issue #873, opened 2017-08-16
Using Vue.Draggable with a bootstrap-vue table


rymate1234:
Hi, 

I'm currently using the bootstrap-vue table element, and want to use the [Vue.Draggable](https://github.com/SortableJS/Vue.Draggable) library to add drag and drop item moving and reordering.

However, it appears that you need to use a standard vue v-for loop to use the library, which is as far as I can tell abstracted away in order for filtering and pagination to work

Is there a way to do this without manually constructing a table and losing the filtering and pagination?

tmorehouse:
I Don't think there is a way to directly add that library with `b-table`.  You could roll your own b-table, or take the current b-table code and make a new component that has the dragable features.

boriskogan81:
+1

SebastienDenooz:
+1

tmorehouse:
the latest `dev` branch now allows you to specify a primary column key that uniquely identify a table row.  The primary-key column must be unique across all of your data in the items array.  This value (string or number) will be used as the `:key` value on the rendered `<tr>`.  A transition object can be passed to b-table now, for defining list-transitions on the `<tbody>`  (the `<tbody>` becomes the root element of the transition group.

Checkerbeule:
Hey people. I also have a problem with vue.draggable and b-table. I can't make my code work.
Here ist a code snippet:

<draggable v-model="listA" draggable=".tr">
  <b-table hover :items="listA" primary-key="id" tbody-transition-props="transProps"></b-table     </draggable>

what is the problem here?
Im looking forward for your help!

tmorehouse:
@Checkerbeule Vue-Dragable expects the transition element to be a direct child of the draggable component.

Try using <b-table-simple>, which allows you to structure the table directly.

Checkerbeule:
@tmorehouse thanks a lot!
It ist working with `<b-table-simple>` .

Tywele:
@Checkerbeule Could you post a fiddle of your solution with `<b-table-simple>`? I'm having some trouble wrapping my head around it.

euoia:
In case anyone else comes across this thread, I made a codepen showing `b-table-simple` working with `Vue.draggable`.
https://codepen.io/picklebone/project/editor/ZnGkaa#

PR #1247, opened 2017-10-26
feat(table): Add option to use drag-and-drop reordering with Vue.Draggable #873

sschadwick:
Adding drag and drop functionality to table. Passing in the `:draggable='true'` property does not currently work with `top-row` or `row-details`, as it will skew the internal `<draggable>` counter and move incorrect items.

### DIFF
--- a/docs/components/table/README.md
+++ b/docs/components/table/README.md
@@ -1205,4 +1205,13 @@ export default {
 <!-- table-complete-1.vue -->
 ```
 
+## Drag and Drop Reordering with Vue.Draggable
+To incorporate reordering rows within a table, pass the `draggable` property in as `true`. The variable passed into `:items` will be updated with the newly mutated array as you reorder the array of items.
+
+Note: This does not currently work with `top-row` or `row-details`!
+
+```html
+<b-table :draggable='true' :items='items'></b-table>
+```
+
 ## Component Reference
--- a/lib/components/table.vue
+++ b/lib/components/table.vue
@@ -58,12 +58,65 @@
                 </th>
             </tr>
         </tfoot>
-        <tbody>
+        <draggable :list='items' v-if='draggable' element='tbody'>
+            <template v-for="(item,index) in computedItems">
+                <tr :key="index"
+                    :class="rowClasses(item)"
+                    @click="rowClicked($event,item,index)"
+                    @dblclick="rowDblClicked($event,item,index)"
+                    @mouseenter="rowHovered($event,item,index)">
+                    <template v-for="field in computedFields">
+                        <td v-if="$scopedSlots[field.key]"
+                            :class="tdClasses(field, item)"
+                            :key="field.key"
+                            v-bind="field.tdAttr">
+                            <slot :name="field.key"
+                                  :value="getFormattedValue(item, field)"
+                                  :unformatted="item[field.key]"
+                                  :item="item"
+                                  :index="index"
+                            ></slot>
+                        </td>
+                        <td v-else
+                            :class="tdClasses(field, item)"
+                            :key="field.key"
+                            v-html="getFormattedValue(item, field)"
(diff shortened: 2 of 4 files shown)


tmorehouse:
This adds an extra dependancy to the Bootstrap-Vue library (and weight to the final compiled javascript).

It might be better suited to be created as a `components-extra` (which extends the `b-table` component) that does not get compiled into the main library

tmorehouse:
There are a few caveats with this PR:

One is that items is not always an array.  It can be an items-provider function (there is no check for this in your code).

Items should technically not be mutated (which is why we create  shallow copy of items before paginating, sorting, filtering in the `computedItems` computed prop).  Displayed data (the visible rows) are available via the `v-model` interface.

b-table also slices the items array when paginating.  This PR doesn't take that into account, so the items list you are passing to `dragable` is not always the same as the list of items that is passed to the v-for loop (which is shallow sliced copy of the original items after pagination, sorting and filtering are applied. (i.e. it is a new array instance). Not to mention that the props `items` could be a function reference.

One would have to disable any pagination, sorting and filtering, and disabled the user of providerFunctions internally to prevent issues (and bug reports), not to mention changing how `computedItems` is calculated.

B-Table is already one of the more complex/large components in Bootstrap-Vue, and this would just increase that complexity (dealing with all the _what-ifs_).

Which is why I'll recommend that this not be applied as a patch to `b-table`, and recommend that it be created as a `components-extra` add-on.

tmorehouse:
You could easily wrap `b-table` in a `dragable` component, based on this from the vue-dragable docs:

image

You can specify a CSS selector to target a specific thing that is being drag sorted:

```html
<dragable v-model="myItems" :options="{draggable:'tbody > tr'}">
  <b-table :items="myItems" />
</dragable>
```

The key here is telling dragable what you want _dragable_ via the `dragable` option, which accepts a CSS selector.

This is without the need to embed dragable within the `<b-table>` component.  This targets the rendered `<tr>`'s within `<tbody>` (the CSS selector `tbody > tr`).  If your table isn't empty, and doesn't specify any top or bottom static rows, not user pagination, filtering or storing on b-table (since you are now incontrol, and don't specify ny of these on teh `<b-table>` markup. 

This would work without the need of a PR.

tmorehouse:
Also to mention that Vue-Draggable is not currently SSR friendly (ses issue https://github.com/SortableJS/Vue.Draggable/issues/242), which would make b-table not SSR friendly (which it is today).

sschadwick:
Thank you for the quick response and the well thought-out response. You make several great points that I overlooked while writing the PR. I wasn't able to get the `:options="{draggable:''}"` prop working and ended up building a close approximation of the table in my code and using the Draggable element as `tbody`.
```html
<draggable v-model='obj.items' element='tbody'>
  <tr class='row' v-for='item in obj.items' :key='item.id'>
  ...
  </tr>
</draggable>
```

Upon further thought, I think there are too many obstacles and the use case is too specific to move into `extra-components`. I do however look forward to contributing in the future, and would like to thank you for such an awesome project.

tmorehouse:
I still think vue-draggable could be tweaked a bit to be able to handle an "alternate" root element (via CSS query selector), which would allow it to be wrapped around a "deeper" component, such as `<b-table>`.

I might open a PR against it in the next week or so :-)

Kostecki:
> I still think vue-draggable could be tweaked a bit to be able to handle an "alternate" root element (via CSS query selector), which would allow it to be wrapped around a "deeper" component, such as `<b-table>`.
> 
> I might open a PR against it in the next week or so :-)

Do you ever do in regards to this? :-) I'm trying to get vue.draggable to work with b-table and your quick solution of using the draggable option doesn't seem to work. I don't see why though?

Any other suggestions for how it could be done without rewriting a bunch of things?

PR #1257, opened 2017-10-27
[wip] new structure

pi0:
This PR changes lib directory structure in a more modular one. One step towards modular packages and better ESM support.

Status:
- [x] Marge tests/lib/docs into src
- [x] Update builds
- [x] Update docs
- [x] Get search working
- [x] Update tests (optional)
- [ ] Tree shakeable esm builds (optional)

I think optional labeled parts can be non-blocking for 1.0.0 release (and merging this branch twice for faster release)

The new structure can be reviewed from [branch](https://github.com/bootstrap-vue/bootstrap-vue/tree/feat/new-structure/src). Much better than github diff :))


mosinve:
Hm...tests wont fix... 
```
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { loadFixture, testVM } from "../../utils/helpers";
                                                                                             ^^^^^^
    SyntaxError: Unexpected token import
```
Will try to deal with'em a bit later

pi0:
@mosinve I think this is because the lack of `.babelrc` and jest won't understand that :)) Thnks for your help, right now i'm focused on fixing docs. Will come back to help fixing tests too.

docs/utils/index.js
```diff
@@ -90,3 +69,40 @@ export function importAll(r) {
...
+
+// Return an element's offset wrt document element
+// https://j11y.io/jquery/#v=git&fn=jQuery.fn.offset
+export function offsetTop(el) {
+    if (!el.getClientRects().length) {
+        return 0;
+    }
+    const bcr = el.getBoundingClientRect();
+    const win = el.ownerDocument.defaultView;
+    return bcr.top + win.pageYOffset;
```
tmorehouse: Where does `win` get defined?

pi0: Don't know just refactored old docs :))) What should it be?

tmorehouse: window :-)

Must have been in there somewhere originally (or missed) when I refactored it.  Although it appears to be working...

tmorehouse:
Should `container` be moved into `layout` along side `col` and `row`?

pi0:
It is also a dep of the carousel. But seems a good idea.

tmorehouse:
Hmmm I didn't think carousel used b-container...

package.json
```diff
@@ -43,7 +43,7 @@
     "url": "git+https://github.com/bootstrap-vue/bootstrap-vue.git"
   },
   "files": [
-    "lib",
+    "src",
     "dist",
     "nuxt"
```
tmorehouse: I think `nuxt` has to change to `docs`

pi0: This nuxt refers to nuxt module not docs. I will deprecate `@nuxtjs/bootstrap-vue` in flavor of this internal so we can offer an always update integration module :)

pi0:
Seems we are in a stable status now. Really thanks for your contribution @tmorhouse @mosinve ❤😌

package.json
tmorehouse: Ah, ok...

PR #1267, opened 2017-10-31
feat: Make Components & Directives available as Vue plugins

tmorehouse:
Extending @pi0's `es` build enhancement (#1257), this PR adds an `index.js` file to each component/directive directory, allowing each (along with sub components and related directives) to be imported as Vue Plugins as needed.

One can easily import `b-card` and all its related sub component as so:
```js
import Card from 'bootstrap-vue/es/components/card';
Vue.use(Card);
```
Or import `b-modal` (including directive `v-b-modal`) as so:
```js
import Modal from 'bootstrap-vue/es/components/modal';
Vue.use(Modal);
```

Users can still also import single components as so:
```js
import bCard from 'bootstrap-vue/es/components/card/card'; // .js functional component
import bModal from 'bootstrap-vue/es/components/modal/modal.vue';
import vbModal from 'bootstrap-vue/es/directives/modal/modal';  // .js file
Vue.component('b-card', bCard);
Vue.component('b-modal', bModal);
Vue.directive('b-modal', vbModal);
```

### DIFF
--- a/src/components/breadcrumb/index.js
+++ b/src/components/breadcrumb/index.js
@@ -13,8 +13,8 @@ const components = {
 
 const VuePlugin = {
   install(Vue) {
-    if (!registerComponent(Vue, 'breadcrumb')) {
-      for (var component in components) {
+    for (var component in components) {
+      if (!registerComponent(Vue, component)) {
         Vue.component(component, components[component]);
       }
     }
--- a/src/components/alert/index.js
+++ b/src/components/alert/index.js
@@ -9,8 +9,8 @@ const components = {
 
 const VuePlugin = {
   install(Vue) {
-    if (!registerComponent(Vue, 'alert')) {
-      for (var component in components) {
+    for (var component in components) {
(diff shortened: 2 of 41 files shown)


pi0:
@tmorehouse Fill free merging when it was stable enough :)

tmorehouse:
Just adding in a plugin helper util to determine if a plugin has already been loaded

src/components/breadcrumb/index.js
```diff
@@ -1,6 +1,7 @@
 import bBreadcrumb from './breadcrumb';
 import bBreadcrumbItem from './breadcrumb-item';
 import bBreadcrumbLink from './breadcrumb-link';
+import registerComponent from '../../utils';
```
pi0: Shoudn't it be `{ register component }` ?

tmorehouse: Yep, all fixed now ;-)

PR #1278, opened 2017-11-02
feat(progress): Add configurable transition time support + Rounded ends

tmorehouse:
For some strange reason Bootstrap V4.beta.2 removed the default transition (`width 0.6s ease`) from class `.progress-bar`, so the progress bar no longer animates when the value changes. There is no mention in the BSV4 PR https://github.com/twbs/bootstrap/pull/22703 as to why transition was removed.

This PR adds in configurable transition support via a new prop `transition-time` (value given in seconds), and applies the transition as an inline style.

If `transition-time` is not set, then no transition between values will occur (unless CSS is used to provide transition support).

I've asked in the BS PR as to why this might have been removed. So we may want to hold off on this PR until we hear an answer from the TWBS team

### DIFF
--- a/src/components/progress/README.md
+++ b/src/components/progress/README.md
@@ -34,12 +34,15 @@ export default {
 
 <!-- progress-1.vue -->
 ```
+
+
 ## Value
 Set the maximum value with the `max` prop (default is `100`), and the current value via the
 `value` prop (default `0`).
 
 When creating multiple bars in a sinple process, place hte value prop on the individual
 `<b-progress-bar>` sub components (see the **Multiple Bars** section below for more details)
+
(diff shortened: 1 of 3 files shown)


tmorehouse:
It appears that TWBS will be adding the progress-bar transition back in in V4.0.0-beta.3.

We could either leave this in as is, or modify it slightly to only set the `transition-duration` CSS property in the inline `style` value.

tmorehouse:
Putting this PR on hold until details are finalized at TWBS.

tmorehouse:
Closing as the template file has changed, and pending on BSV4.beta.3 changes

PR #1279, opened 2017-11-02
feat: make all components available as .js files

tmorehouse:
This will make it easier to import individual components without needing to know if it should have a `.vue` extension or not.


tmorehouse:
And if we convert additional components to function components, or rewrite some of the `.vue` compoennts as render function based compoents, it won't be a breking change for users who import individual components.

PR #1292, opened 2017-11-06
feat(pagination): New prop for adding link to the last page in pagination

annandr12:
This PR provides property for showing link to the last page in pagination (like a button with number not like arrow to the first or last). Link will be show only if last ellipsis are shown (looks sensibly for me). It is necessary when need to show how many pages we have.

### DIFF
--- a/src/components/pagination/pagination.vue
+++ b/src/components/pagination/pagination.vue
@@ -74,6 +74,25 @@
             <span class="page-link" v-html="ellipsisText"></span>
         </li>
 
+        <!-- Last Page Link -->
+        <li v-if="showLastPage && showLastDots" role="none presentation">
+          <a :class="pageLinkClasses(numberOfPages)"
+             :disabled="disabled"
+             :aria-disabled="disabled ? 'true' : null"
+             :aria-label="labelPage + ' ' + numberOfPages"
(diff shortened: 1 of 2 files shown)


tmorehouse:
Thanks for the PR...

A few considerations:
- Pagination-nav should also provide the same user interface.  By adding the new prop to the mixin, you have added the prop also to pagination-nav, but not the feature.  This will cause user confusion as pagination-nav  will show the prop in reference, it will not work.
- The number of buttons shown (excluding first/prev/next/last), i.e. limit calculations,  should take this extra button into account when restricting the total number of buttons, other wise the "shape" of pagination will jump wider only when the last ellipsis is shown (i.e. the buttons on the right will get jump to the right). This would add a bit more complexity to the calculations of the page buttons to display.
- Should there also be an option to display the first page button (i.e. always show button "1" when the first ellipsis is shown.

Also, currently, with this PR it is possible to see something like this:
```
 |<<| < | ... | 7 | 8 | 9 | ... | 10 | > | >>|
```
(assuming a maximum number of pages of 10, current page is 8, limit is 5), which doesn't make visual sense having the elipsis between 9 and 10.

Would it make more sense to change the text of the "GotoLastButton" to be the last page number?

tmorehouse:
Due to recent changes in the usage of render functions, this PR is now out of date.

PR #1298, opened 2017-11-07
Fix wrong relative path to package.json

BruceHem:


tmorehouse:
Hmm, this might not be pointing to the wrong file, but to a missing file (Nuxt modules usually have a local package.json file that describes the module)

@pi0 should this point to Bootstrap-Vue's package.json or to its own package.json?

BruceHem:
Please help publish an npm patch for this, It'd be helpful for me.

PR #1302, opened 2017-11-07
[SSR/Nuxt] Change the way bootstrap-vue is imported

BruceHem:
This way it will work with nuxt, without error.


pi0:
The problem with some components is probably not a nuxt only bug. If we even decide to use esm bunble as a temporary fix, I think it would be better to update fields in `package.json` instead :)

pi0:
Thanks @BruceHem. Things should be working with 1.0.1 :)

PR #1304, opened 2017-11-07
fix(form-file) missing formSizeMixin

### LINKED ISSUE #1303: form-file "sizeFormClass" is not defined
[Vue warn]: Property or method "sizeFormClass" is not defined on the instance but referenced during render.

Looks like missed `formSizeMixin`

https://github.com/bootstrap-vue/bootstrap-vue/blob/6cfcd545ccd362e8cc3a042be5e9f953d5f6a807/src/components/form-file/form-file.vue#L133

https://github.com/bootstrap-vue/bootstrap-vue/blob/6cfcd545ccd362e8cc3a042be5e9f953d5f6a807/src/components/form-file/form-file.vue#L137

mosinve:
fixes #1303

### DIFF
--- a/src/components/form-file/form-file.vue
+++ b/src/components/form-file/form-file.vue
@@ -130,11 +130,11 @@
 </style>
 
 <script>
-    import { idMixin, formStateMixin, formCustomMixin, formMixin } from '../../mixins';
+    import { idMixin, formStateMixin, formCustomMixin, formMixin, formSizeMixin } from '../../mixins';
     import { from as arrayFrom } from '../../utils/array';
 
     export default {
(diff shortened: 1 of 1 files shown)


tmorehouse:
I thought that form-file didn't support the size class based on official Bootstrap Docs.  Since it doesn't use form-component-{sm|lg}

And the inner hidden file input sizes strangely causing some areas of the custom file input to be not clickable.

Which is why I left out the size mixin, and didn't document the size attribute.

Might need to revert this

tmorehouse:
Also, the mixin doesn't automatically apply the size classes...

pi0:
@tmorehouse Opps. Sorry for fast merge. Would you please do that? We may have 1.0.2 then :)

PR #1305, opened 2017-11-07
fix(form-file): control size not supported in BS V4

tmorehouse:
Bootstrap currently does not support the form-control-size on form-file input.

Removing both the mixin and the and the class reference.


pi0:
LGTM. Please merge your self if is working :)

PR #1313, opened 2017-11-08
perf(navbar-toggle): convert template to render function

tmorehouse:
Also removed no-longer used prop `position`


tmorehouse:
When compiling to the ES modules, the .vue template get copied over as is (no transpilation).  and we have had reports of SSR bail rrors for users importing the `es` module version of Bootstrap-Vue (eitehr as a whole, or individual components).  This is due to issues with vue-loader configurations for various users (in many cases, the imported .vue templates were rendering a single extra space during SSR and when hydrating on the client side Vue would bail because of the extra space.

Converting the .vue files to render functions (although not as legible as templates) avoids this SSR hydration bail.

The only other option is to have some vue-compiler generate the components as `.js` files in the `es` modules directory (instead of copying over the raw .vue files)

mosinve:
have you seen this https://vue-loader.vuejs.org/en/options.html#optimizessr ?

mosinve:
I think the issue not in vue-compiler, but in rollup-plugin-vue,  as vue-loader uses vue-compiler and supports SSR, but i'm not sure Rollup is.

tmorehouse:
These would be similar to the render functions used in our functional components.  Although functional components now support using single file components (using the latest vue-loader version).

PR #1317, opened 2017-11-08
fix(input-group): correct input-group right addon via prop

williammck:
Found this issue on >=1.0.0, appeared in commit bd4c3c3.


tmorehouse:
Thanks!

PR #1327, opened 2017-11-10
fix(modal): modal-open body class lost when switching between modals

### LINKED ISSUE #1325: Opening another Modal from a Modal breaks body scrolling
### Steps to reproduce

1. Create two `<b-modal>`.
2. Open first modal.
3. While the first modal is opened, trigger opening the second modal.

### Actual result

1. First modal hides
2. Second modal shows
3. Background body becomes scrollable

### Expected result

1. First modal hides
2. Second modal shows
3. Background body stays fixed

### Example

See this jsfiddle: https://jsfiddle.net/semenov/ngdr0gtk/

1. Click "Show Modal 1"
2. Ensure that the long background doesn't scroll
3. Click "Show Modal 2"
4. Observe that the long background becomes scrollable

### Environment

* Chrome 62 under macOS (latest)
* Vue 2.5.3 (latest)
* bootstrap-vue 1.0.2 (latest)
* bootstrap 4.0.0-beta.2 (latest)

### Workaround

The simple (but ugly) workaround is to manually hide the currently opened modal, and open the second modal with `setTimeout`. Note that the timeout must be reasonably large, in the sense that 0 (or `$nextTick`) or even 100 is not enough. Possibly, it has something to do with the animation.

IlyaSemenov:
This fixes #1325, preventing `modal-open` class lost on document body when a new modal opens before the previous one's transition is over.

This sets a new custom class `modal-closing` on body during transition, I am not sure if that's appropriate. It could probably be better to use a flag variable on `this.$root`.

### DIFF
--- a/src/components/modal/modal.vue
+++ b/src/components/modal/modal.vue
@@ -358,18 +358,17 @@
                     relatedTarget: null
                 });
                 this.emitEvent(showEvt);
-                // Show if not canceled
                 if (showEvt.defaultPrevented || this.is_visible) {
+                    // Don't show if canceled
                     return;
                 }
-                this.is_hidden = false;
-                this.$nextTick(() => {
-                    // We do this in nextTick to ensure hte modal is in DOM first before we show it
-                    this.is_visible = true;
-                    this.$emit('change', true);
-                    // Observe changes in modal content and adjust if necessary
-                    this._observer = observeDom(this.$refs.content, this.adjustDialog.bind(this), OBSERVER_CONFIG);
-                });
+                if (hasClass(document.body, 'modal-open')) {
+                    // If another modal is already open, wait for it to close
+                    this.$root.$once('bv::modal::hidden', this.doShow)
+                } else {
+                    // Show the modal
+                    this.doShow();
+                }
             },
             hide(trigger) {
                 if (!this.is_visible) {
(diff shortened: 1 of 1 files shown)


tmorehouse:
We'll need to test this out to ensure that the body padding and fixed/static/sticky element padding/margins are not doubled up when the new modal is opened. 

It might be better to trigger a listen (`this.$root.$once()`) for the event `bv::modal::hidden`, which is fired after a modal has completely hidden, before opening the new modal, which will ensure all adjustments to navbar padding/margins and body padding have been returned to normal.  We use a similar technique in b-carousel to wait for a transition to complete before jumping to another slide.

IlyaSemenov:
Agreed, this makes sense.

tmorehouse:
Just did a quick test using the following code:

```html
<div>
  <b-btn v-b-modal.modal1>Launch demo modal</b-btn>
  <b-modal id="modal1" title="Bootstrap-Vue">
    <p class="my-4">Hello from modal!</p>
    <b-btn v-b-modal.modal1a>Open second modal</b-btn>
  </b-modal>
  <b-modal id="modal1a" title="Bootstrap-Vue">
    <p class="my-4">Hello from modal #2!</p>
  </b-modal>
</div>
```
And it appears to work well.  Note that there is a small delay of the new modal open as it waits for the first modal transition to end.

src/components/modal/modal.vue
```diff
@@ -438,21 +448,17 @@
             onBeforeLeave() {
                 this.is_transitioning = true;
                 this.setResizeEvent(false);
-                addClass(document.body, 'b-modal-closing');
-            },
+             },
```
IlyaSemenov: Extra space character sneaked here by mistake.

tmorehouse: Fixed with commit https://github.com/bootstrap-vue/bootstrap-vue/commit/d5a0487686d485b8999afd2dc4e62418519d4a32

PR #1362, opened 2017-11-18
feat(tabs): vertical tabs + new props for adding classes to inner elements

tmorehouse:
Adds support for vertical tabs via the `vertical` prop

Also adds the following new props:
- `container-class` added to the `.tab-container` div
- `nav-class` added to the `.nav` tab controls
- `nav-wrapper-class` added to div wrapper around nav tab controls

### DIFF
--- a/src/components/tabs/README.md
+++ b/src/components/tabs/README.md
@@ -64,6 +64,92 @@ the pill style variant.
 <!-- tabs-pills.vue -->
 ```
 
+## Bottom placement of tab controls
+
+Move the tab controls to the bottom by setting the prop `end`
+
+```html
+<b-card no-body>
+    <b-tabs pills card end>
+        <b-tab title="Tab 1" active>
+            Tab Contents
+        </b-tab>
+        <b-tab title="Tab 2">
+            Tab Contents 2
+        </b-tab>
+    </b-tabs>
+</b-card>
+
+<!-- tabs-bottom.vue -->
+```
+
+**Note:** the `bottom` prop has been deprecated in favor of the `end` prop.
+
+## Vertical tabs
+
+Have the tab controls placed on the lefthand side by setting the `vertical` prop to `true`.
+
+```html
+<b-card no-body>
+    <b-tabs pills card virtical>
+        <b-tab title="Tab 1" active>
+            Tab Contents
+        </b-tab>
+        <b-tab title="Tab 2">
+            Tab Contents 2
+        </b-tab>
+        <b-tab title="Tab 3">
+            Tab Contents 3
+        </b-tab>
+    </b-tabs>
+</b-card>
+
+<!-- tabs-vertical.vue -->
+```
+
+Move the tab controls to the right hand side by setting the `end` prop:
+
+```html
+<b-card no-body>
+    <b-tabs pills card virtical end>
+        <b-tab title="Tab 1" active>
+            Tab Contents
+        </b-tab>
+        <b-tab title="Tab 2">
+            Tab Contents 2
+        </b-tab>
+        <b-tab title="Tab 3">
+            Tab Contents 3
+        </b-tab>
+    </b-tabs>
+</b-card>
+
+<!-- tabs-vertical-end.vue -->
+```
+
+The width of the vertical tab controls will expand to fit the width of the tab title.
+To control the width, set a [width utility class](/docs/reference/size-props#sizing-utility-classes)
+via the prop `nav-wrapper-class`. You can use values such as `w-25` (25% width), `w-50` (50% width),
+or max-width vlaues such as `mw-25` or `mw-50`.
+
+```html
+<b-card no-body>
+    <b-tabs pills card virtical nav-wrapper-class="w-50">
+        <b-tab title="Tab 1" active>
+            Tab Contents
+        </b-tab>
+        <b-tab title="Tab 2">
+            Tab Contents 2
+        </b-tab>
+        <b-tab title="Tab 3">
+            Tab Contents 3
+        </b-tab>
+    </b-tabs>
+</b-card>
+
+<!-- tabs-vertical-width.vue -->
+```
+
 
 ## Fade animation
 
--- a/src/components/tabs/README.md
+++ b/src/components/tabs/README.md
@@ -91,7 +91,7 @@ Have the tab controls placed on the lefthand side by setting the `vertical` prop
 
 ```html
 <b-card no-body>
-    <b-tabs pills card virtical>
+    <b-tabs pills card vertical>
         <b-tab title="Tab 1" active>
             Tab Contents
         </b-tab>
@@ -111,7 +111,7 @@ Move the tab controls to the right hand side by setting the `end` prop:
 
 ```html
 <b-card no-body>
-    <b-tabs pills card virtical end>
+    <b-tabs pills card vertical end>
         <b-tab title="Tab 1" active>
             Tab Contents
         </b-tab>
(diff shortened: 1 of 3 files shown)


src/components/tabs/README.md
```diff
@@ -92,7 +183,7 @@ In some situations, you may want to add classes to the `<li>` (nav-item) and/or
 the `title-item-class` prop (for the `<li>` element) or `title-link-class` prop (for the
 `<a>` element).  Value's can be passed as a string or array of strings.
 
-Note: THe `ative` class is automatically applied to the `<a>` element. You may need to accomodate
+Note: THe `ative` class is automatically applied to the active tabs `<a>` element. You may need to accomodate
```
Morgul: Two typos: `THe` and `ative`.

tmorehouse: Thanks! 👍

Morgul: No problem! Just happened to notice while looking through the changes. 😄

pimlie:
Some remarks (some you already mentioned yourself in #558):
- the vertical tabs are only displayed correctly when using pills, 'normal' tabs are not symmetrical so you need the css from #558 to have correct borders etc
- the pill'ed bottom tabs are missing a border (probably need to switch class card-header to card-footer)
- I think adding `no-gutters` class should be optional, a user might want to (try to) align the columns with a previous row and now we force them to use `no-gutters` on that row as well
- In my previous PR you mentioned that for assistive technologies it would be best that in the html structure the tab navs always comes first, thats partially why I used `order-12` classes to move the tab navs for position bottom and right. In this PR thats not the case, is that on purpose or an oversight?

Although I am grateful for your assistance with this feature, I feel like we jumped ship a bit as it seems the questions I asked here https://github.com/bootstrap-vue/bootstrap-vue/pull/558#issuecomment-345396592 are still relevant. To me this PR only confuses the situation a bit more, because I still believe that we should either support bottom / vertical tabs fully including all the required css or if you dont want that for `b-tabs` we make the move to a separate `b-tabs-extra` component.

tmorehouse:
We're trying to work with the limitations of the existing Bootstrap V4 CSS and using existing styling clases to make it work.

Even in native Bootstrap vertical navs, the do not style the tabs appropriately.  HTey show the regular tab style (as tehy re just using `flex-column` class to move them into vertical mode.

Styling issues should probably be opened up in the twbs/bootstrap repo.

I still agree that the tab controls should be visually moved to the end, rather than physycally.  We can address that in this PR>

I'll see if using the `card-footer` class will add the top border when the controls are moved to the bottom.

As for the border when placed on the sides, bootstrap doesn't have any CSS classes for adding the side border (just removing them) unfortunately.

tmorehouse:
`no-gutters` removes the padding from the columns and the negative margins from the b-row.  the row itself will be aligned with any other row before or after it (even if they don't have the no-gutters set).  The columns won't necessarily align (but their edge boundaries will).

Without the no-gutters option, there is overly excessive space between the tabs and the content when in vertical mode (basically double what is expected).

One can always create a custom class which is applied to the `b-tabs` component, and use that class to target the styling of the inner elements within b-tabs (i.e. changing the Bootstrap default styling of the nav-item in the tab controls).

pimlie:
Bootstrap has removed support for vertical tabs since v3.0 (see their issue 6342) so I dont think opening styling issues about that in the bootstrap repo will be useful. They specifically stated that: _You can still use the JavaScript plugin and custom CSS to make left/right tabs, but we will no longer include those in the core._

That leaves us at the main question again, how far do we want to stray away with `bootstrap-vue` in this case? Bootstrap doesn't support any other tabs then with the navs on top. They dont provide css for tabs on the left, right or even bottom and I am quite sure they never will because that probably doesnt fit in their `mobile-first` strategy. Again, I dont think we should take the 'easy' route by only partially supporting vertical tabs due to not providing the correct css for all use-cases that the `b-tabs` component provides.

With regards to `no-gutters`, I am sure you will understand that I ran into that problem as well with my PR. But I still made it optional as I don't think we should oppose styling options like that on users. Although you are right by saying you can always add a custom class and change the inner elements with that, that's much more difficult then just a `no-gutters="false"` prop.

tmorehouse:
Visually, without the no-gutters class, the layout sucks (especially on narrower screens due to the excessive double padding that occurs from card components).  When not used with card integration, the no-gutters class is not applied.

I think what is here in this PR adds enough functionality within the limits of Bootstrap V4's CSS, and provides basic support and accessible markup.  Visually, if people want to change how the tabs mode looks in vertical mode, they can provide custom CSS to address it, just as they would have to if using vertical tabs style navs in Bootstrap V4 native. The visual styles of the vertical (and bottom placement) tab style navs, is documented as not being "stylish", however, the pills style works quite well.

There is enough functionality within the component that masks some of the gory tweaks that needs to be done in order to make the controls vertical which normally couldn't be done easily (due to the elements being buried inside a component).

tmorehouse:
Hmm after playing with ordering, it doesn't work well for keyboard users as the visual tab sequence is confusing (if there are elements with tab stops inside the content).

For visually impaired users, while ordering would make better semantic sense (nav controls always on top), we do have ARIA attributes that to relate the content to the elements (navs) that do control the content.  So I think reverting back to physical ordering of the navs/content might be best (rather than using flex ordering)

pimlie:
Although I understand your view about having `enough gory tweaks` I have to respectfully disagree with your conclusion. 

All in all we are providing just half of the solution and are shifting the workload to the users who individually have to repeat the same process again and again. Isn't there a mid-way like including the required css as an extra component or maybe even a separate but official package? Then at least we provide the full solution and maintenance on vertical tabs only has to be done once for the whole user-base.

Maybe this is nitpicking but I think we should strive towards that it works 'well' and not settle for 'quite well' :smile:

tmorehouse:
The real issue is Bootstrap's CSS.  We do want to avoid any massive amounts of custom CSS in Bootstrap-Vue, as it can easily conflict with user's custom CSS, or bootstrap V4 custom CSS

Visually I do not find the tab version that bad when in vertical mode.  It doesn't look good in non-vertical bottom mode, but that is just a caveat of the base Bootstrap V4 CSS.
OK:
image
Not as OK, but not ugly:
image
Definitely OK:
image
Definitely OK as well:
image

The big issue with custom CSS, when not used to just tweak positioning/margins, temporary fixes for bugs in Bootstrap V4's beta CSS (i.e. form-file), or providing non-bootstrap styles for things like the sorting arrows in b-table, but actually controls border styles and color, is that it ends up being hard coded in the bootstrap-vue.css distribution file.  If someone is using a custom themed Bootstrap V4 (which changes the default border colors of say, nav-tabs, then our hard coded CSS overrides some of theirs ending up with wonky looking  components, or things like nav-tabs not following their custom theaming 😮 

Even if one uses SCSS to define the custom component styles (i.e. making new style for the b-tabs nav items), importing the variables from the Bootstrap V4 SCSS, then end result when Bootstrap-Vue gets built/compiled and distributed is that the SCSS variables are compiled into absolute/static values in the final bootstrap-vue.css dist file,

Importing that bootstrap-vue.css file into someones project may end up with components not matching the styling of the rest of their project.  This is why it is best to avoid as much custom CSS as possible.

pimlie:
> This is why it is best to avoid as much custom CSS as possible

Sure, but I still dont agree with your conclusion in this case as the bulk of the users will probably not use theming anyway. And even if they would there is no difference for them to just not include the `bootstrap-vue(.extra?).css`. I think the problem cases you are describing as an argument against adding the custom CSS actually are advanced edge-cases for which we can also easily provide a solution by adding all custom css in scss file ourselves. The fact that some custom 3rd party bootstrap theme doesnt include styling for vertical tabs is evidently and really cant be used as an argument against including CSS here.

The factual issue at hand is that if we have made the decision to support a feature that is not supported by bootstrap we should support it all the way so that at least the basics are covered. Otherwise you are just saying _Well you can use that but it was too difficult or too much work for us to implement fully and because we only see problems and no solutions you have to do the rest yourself_

Sorry, I really appreciate your hard work but I just feel very strongly about half-assing stuff like this. The argument against adding custom CSS should be made _before_ adding a feature and not _after_ adding a feature :disappointed: 

So eh, my last proposal before I will stop this discussion and mock and curse in silence :wink:  What about taking an example from nuxt and create a bootstrap-vue community repo / organisation where stuff like this can be added? That way you don't have to deal with the css and it will become much easier to list/share custom/extra components that should work nicely with `bootstrap-vue` but are supported by the community instead.

tmorehouse:
http://getbootstrap.com/docs/4.0/components/navs/#vertical  Vertical navs are supported by Bootstrap. they just don't provide any special styling for the tabs versions.

pimlie:
That was indeed maybe a bit harsh statement of me, sorry about that. So eh, case closed what I am concerned.

PR #1365, opened 2017-11-19
Improve docs for alert

Tekki:


pi0:
Thanks!

Tekki:
I'll take a look at the examples for the other components too.

src/components/alert/README.md
```diff
@@ -44,17 +44,19 @@ Alerts are available for any length of text, as well as an optional dismiss butt
 
 <script>
 export default {
-  data: {
-    dismissSecs: 10,
-    dismissCountDown: 0,
-    showDismissibleAlert: false
+  data () {
```
tmorehouse: Technically, the examples are run as app instance (each example is a `new Vue(...)`), rather than component instances, so `data` as a function is not required.

Tekki:
@tmorehouse  In my opinion, it should be possible to copy the examples into a .vue file and this file should compile without errors. At the moment this is not possible, not only because of 'data'. But this is your repo and you decide how the docs should look like. So please tell me if I should continue to edit or not.

tmorehouse:
When run as an app, `data` can be an object, rather than a function (which is needed when the example is converted to a custom component, i.e. in a Single File Component - SFC - `.vue` file): https://vuejs.org/v2/api/#Options-Data

Although in application mode, data can either be a function or an object. so it doesn't really matter if the data is converted to a function.

Tekki:
I'll prepare a new pull request and you can discuss it with @pi0.

tmorehouse:
Sure! and thanks!

PR #1366, opened 2017-11-19
feat(tooltip+popover): ability to programmatically show and hide tooltip and popover

tmorehouse:
For the component versions, adds a new syncable Boolean prop `show` which allows the user to control the visibility state of the tooltip or popover.

For both component version and directive versions, one can use the `bv::hide::{tooltip|popover}` $root event to close a particular tooltip or popover by passing the trigger element's ID as the first argument. if no ID is passed, then all tooltips or popovers are closed (existing functionality)

For both component version and directive versions, one can use the `bv::show::{tooltip|popover}` $root event to open a particular tooltip or popover by passing the trigger element's ID as the first argument.

### DIFF
--- a/src/components/popover/popover.js
+++ b/src/components/popover/popover.js
@@ -32,7 +32,7 @@ export default {
     },
     placement: {
       type: String,
-      default: 'auto'
+      default: 'right'
     }
   },
   methods: {
--- a/src/components/popover/popover.js
+++ b/src/components/popover/popover.js
@@ -32,7 +32,7 @@ export default {
     },
     placement: {
       type: String,
-      default: 'right'
+      default: 'auto'
     }
   },
   methods: {
--- a/src/mixins/toolpop.js
+++ b/src/mixins/toolpop.js
@@ -61,14 +61,11 @@ export default {
   },
   watch: {
     show (show, old) {
-      if (show === old || !this._toolpop) {
+      if (show === old) {
         return
       }
-      if (show) {
-        this._toolpop.show()
-      } else {
-        this._toolpop.hide()
-      }
+
+      show ? this.onOpen() : this.onClose()
     }
   },
   created () {
@@ -84,13 +81,15 @@ export default {
       // Instantiate ToolTip/PopOver on target
       // createToolpop method must exist in main component
       if (this.createToolpop()) {
+        // Listen to open signals from others
+        this.$on('open', this.onOpen)
         // Listen to close signals from others
         this.$on('close', this.onClose)
         // Observe content Child changes so we can notify popper of possible size change
         this.setObservers(true)
         // Set intially open state
         if (this.show) {
-          this._toolpop.show()
+          this.onOpen()
         }
       }
     })
@@ -169,6 +168,11 @@ export default {
         cfg.html = true
       }
       return cfg
+    },
+    onOpen () {
+      if (this._toolpop) {
+        this._toolpop.show()
+      }
     },
     onClose (callback) {
       if (this._toolpop) {
--- a/src/mixins/toolpop.js
+++ b/src/mixins/toolpop.js
@@ -53,6 +53,22 @@ export default {
       // String ID of container, if null body is used (default)
       type: String,
       default: null
+    },
+    show: {
+      type: Boolean,
+      default: false
+    }
+  },
+  watch: {
+    show (show, old) {
+      if (show === old || !this._toolpop) {
+        return
+      }
+      if (show) {
+        this._toolpop.show()
+      } else {
+        this._toolpop.hide()
+      }
     }
   },
   created () {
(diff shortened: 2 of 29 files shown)


tmorehouse:
@SirLamer when you get a chance could you test out this PR with your usage scenario?

SirLamer:
Haha well I just finished work on implementing this and was going about testing it before submitting a PR, though events-only to not disturb your current design approach. I prefer the synced prop approach anyway (when I did it before, synced props was not a Vue feature).

I'll review this but it is very late where I am right now so I'll have to kick the tires tomorrow.

SirLamer:
Actually I had a look-over and merged some of my work with yours instead of sleeping, OOPS. I'll give it another pass tomorrow. In particular I want to implement the "advanced reactive" example using the new 'show' prop since I think it's cleaner.

SirLamer:
OH, also, I caught a couple of other documentation errors and fixed those, also I very arrogantly changed the default 'placement' to 'auto' since I think that makes sense, but also because I noticed the documented default did not match the actual default which compelled me to take a look at it in the first place. I can change it back to 'top' or 'right' if you'd prefer those, though I think that would be presumptuous.

tmorehouse:
The default placements were following bootstraps default placements (top for tooltip, and right for popover).  Both will "flip" automatically if tooltip or popover is to close to the viewport edge.

src/mixins/toolpop.js
```diff
@@ -61,14 +61,11 @@ export default {
...
+      if (show === old) {
         return
       }
-      if (show) {
-        this._toolpop.show()
-      } else {
-        this._toolpop.hide()
-      }
+
+      show ? this.onOpen() : this.onClose()
```
tmorehouse: 👍

src/components/popover/popover.js
```diff
@@ -32,7 +32,7 @@ export default {
     },
     placement: {
       type: String,
-      default: 'right'
+      default: 'auto'
```
tmorehouse: Bootstrap V4 uses a default of 'right' for their popovers.

tmorehouse: Although I was never overly fond of it being 'right'

SirLamer: It's fine to match Bootstrap V4, I fussed with it after noticing the docs didn't match the actual as-defined default. I'll change it to match V4.

SirLamer:
Okay I have completed my adjustments, have at it.

tmorehouse:
@SirLamer looks great!

I think we'll wait to merge this just before the next release goes out (as the docs site gets updated automagically from changes in the dev branch).

SirLamer:
Then, as an aside, perhaps the master branch should be used for the release build so that you can merge PRs to the dev build as they are completed, otherwise you are just going to face a collision mess come release.

SirLamer:
By the way, one issue I did not resolve is the behavior of the show-on-load popover state, since this has the same "fails to react to built-in triggers" problem as when the popover is made to programmatically display after load. In effect the user has to click the button twice to make it close. Per the issue discussion on this I didn't take it as part of the scope of this PR to resolve this general issue of event triggers causing a display state change, so I left it as it is for now. The subsequent demo I included illustrates the problem and a potential fix which can be affected externally to Popover.

tmorehouse:
Yeh, we need to change master to be based on the BS V4.beta dev branch, rather than the old alpha branch.  I think @pi0 is working on that.

tmorehouse:
Yup, re the initial show state... Bootstrap V4 native has the same issue. 

One way maybe to address it, is if the tooltip/popover has only a single "trigger" specified, then that trigger is set as the "active" trigger via the show/hide methods in the base class.  But then it still has wonky effect if multiple triggers are specified.

SirLamer:
Maybe it's something better fixed in Popper.js, instead of here.

tmorehouse:
In the original Bootstrap code they set the 'manual' trigger in the active triggers tracking object, so that a tooltip/popover couldn't be closed by any other trigger, except programmatically.  But that is no longer present in their current codebase.

pi0:
Master rebased to dev and CI now deploys from master instead of dev (22e432d972a67638ec9111675333c934e6b0ff1b). So we are safe to merge (And possibly next release :D)

tmorehouse:
@SirLamer at least the caveat regarding manual trigger is mentioned in the docs as a known behavior.

tmorehouse:
So we are good fer merging?

SirLamer:
As far as I am concerned we're good.

Re the manual trigger issue, this PR change does not affect or harm anyone not using programmatic control, so I am okay with the current caveat condition and maybe a better solution will be arrived upon later.

tmorehouse:
What if we introduce a new trigger called `none` or `manual` (which could be used on both directive and component)?

This could disable the default triggers (regular event handlers), which means the popover can only be opened programmatically?

if none/manual is specified with any other triggers, manual/none it would trump the other triggers and take precedence.

SirLamer:
Providing an empty string is the same as "none", this works.

Also, I was experimenting a bit and found that the "blur" trigger works very well with programmatic control. It makes the trigger close when either the button or the popover loses focus after being opened with "show.sync".

tmorehouse:
I've created PR #1387 which adds a `disabled` prop to the component versions of tooltip and popover Take a look when you get a chance.

Issue #1065, opened 2017-09-16
[feat] Disabling Tooltips


brightertools:
Hi, 

Is it possible to disable all tooltips, for instance if I have tooltips on toolbar buttons I don't want users on mobile devices to have to click the button twice?

If I detect the screen is smaller than a certain size, how would I disable the tooltips?

Thanks

tmorehouse:
Currently it is not possible to disable tooltips or popovers.  It is possible to emit an event to close all open tooltips or popovers on a page though.

Are you using the directive or component version of tooltips?

The original bootstrap code has an option to disable tooltips/popovers (on an individual bases), which we have yet to implement.

It is difficult to detect the change in breakpoint programmatically, but you could implement your own browser detection code and set a global flag that instructs your app to not instantiate the component version (i.e. using `v-if` directive on the component version).

brightertools:
Ok, thanks 👍

tmorehouse:
PR #1387 partially addresses this issue by providing a `disabled` prop on the `<b-popover>` and `<b-tooltip>` components.

You would still need to figure out the display size to determine if the tooltips should be disabled.

Note this is only available in the component version of tooltip/popover, not the directive version.

tmorehouse:
v1.3.0 has been released, and has options for programmatically disabling, opening, and closing tooltips/popovers.

PR #1387, opened 2017-11-21
feat(tooltip+popover): programmatically disable/enable tooltip or popover

tmorehouse:
Adds a new syncable prop `disabled` to disabled/enable tooltip/popove componentsr. Defaults to `false`

Also introduces new root events for enabling/disabling all/specific tooltips/popovers (components and directives)

Addresses issue #1065

### DIFF
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -217,7 +217,7 @@ when disabled is set to `false`, the tooltip will close.
       <b-btn id="tooltipButton-disable" variant="primary">I have a tooltip</b-btn>
     </div>
     <div class="p-2">
-      <b-btn class="px-1" @click="disabled = !disabled">
+      <b-btn @click="disabled = !disabled">
         {{ disabled ? 'Enable' : 'Disable' }} Tooltip
       </b-btn>
 
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -270,7 +270,7 @@ prop modifier.
         {{ disabled ? 'Enable' : 'Disable' }} Tooltip by prop
       </b-btn>
             
-      <b-btn @click="disableByRoot">
+      <b-btn @click="disableByRef">
         {{ disabled ? 'Enable' : 'Disable' }} Tooltip by $ref event
       </b-btn>
 
@@ -289,7 +289,7 @@ prop modifier.
       }
     },
     methods: {
-      disableByRoot () {
+      disableByRef () {
         if (this.disabled) {
           this.$refs.tooltip.$emit('enable')
         } else {
--- a/src/components/tooltip/README.md
+++ b/src/components/tooltip/README.md
@@ -221,7 +221,7 @@ when disabled is set to `false`, the tooltip will close.
         {{ disabled ? 'Enable' : 'Disable' }} Tooltip
       </b-btn>
 
-      <b-tooltip :disabled"disabled" target="tooltipButton-disable">
+      <b-tooltip :disabled="disabled" target="tooltipButton-disable">
         Hello <strong>World!</strong>
       </b-tooltip>
     </div>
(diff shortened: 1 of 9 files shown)


SirLamer:
Maybe disable should freeze it in state, open or close? It should just disable triggers and programmatic control should still work.

tmorehouse:
Hmmmm.... that might be an option... but the triggers are passed into the tooltip/popover class instance, so are not directly available in the component versions... unless we dip into the `$activeTriggers` property on `this._toolpop` and add a new value (i.e. `manual: true`)

This PR was mainly to address issue #1065

SirLamer:
Well that's cool, keeping it simple is good. The affect of forcing it open can be done with "show.sync" and an empty "triggers" string

tmorehouse:
From the bootstrap docs:

>`.tooltip('disable')`
> Removes the ability for an element’s tooltip to be shown. The tooltip will only be able to be shown if it is re-enabled.

SirLamer:
ok well maybe our prop should be "disable" instead of "disabled"? The problem is HTML spec uses "disabled" so...

tmorehouse:
We use `disabled` in a lot of other components (i.e. dropdowns, form-inputs, pagination, etc)

Since the attribute (prop) is on a custom component it shouldn't conflict with native attributes.

SirLamer:
I meant that it's best for the word spelling to match common use. I agree there is no collision problem.

Yeah "disabled" is fine

SirLamer:
btw I don't have a chance to run use tests with this today, I had a look and it looks good but I haven't done functional testing.

tmorehouse:
Might be able to add a 'disable()` and `enabled()` methods in the class instance, which would then be called to disable/enable the tooltip/popover.

tmorehouse:
Quickly looking at the bootstrap V4 code, it looks like if the tooltip/popover is visible when disable, it remains visible.

I'll re-implement by adding `disable` and `enable` methods on the base class instance.

tmorehouse:
Added in disable/enable events on the components, as well as $root `bv:{disable|enable}::{tooltip|popover|` listeners to allow enabling/disabling of individual or all tooltips or popovers.

Now just need to re-document the two.

SirLamer:
Yeah that sounds good, I like that approach only because it allows for flexible use, but not the point of bloating the JS package to handle complex logic.

tmorehouse:
The good thing is that most of this stuff is in shared base class or mixin, and not a whole lot more code either

tmorehouse:
@SirLamer I think everything should be all hunky-dory now.

tmorehouse:
Might still be one issue... looking into it now.

src/components/tooltip/README.md
```diff
@@ -245,13 +249,22 @@ prop modifier.
       return {
         disabled: false
       }
+    },
+    methods: {
+      disableByRoot() {
+        if (this.disabled){
+          this.$root.$emit('bv::enable::popover')
```
tmorehouse: these emit events need the trigger button's ID string (the id of the button that has the tooltip), otherwise it enables or disables **all** tooltips

Also the event should be `bv::{enable|disable}::tooltip` for tooltips 😉

src/components/tooltip/README.md
```diff
@@ -245,13 +249,22 @@ prop modifier.
...
+        }else{
+          this.$root.$emit('bv::disable::popover')
```
tmorehouse: ditto for here 😜

src/components/tooltip/README.md
mosinve: Yep, true

src/components/tooltip/README.md
```diff
@@ -255,9 +291,9 @@ prop modifier.
     methods: {
       disableByRoot () {
         if (this.disabled) {
-          this.$root.$emit('bv::enable::tooltip', 'tooltipButton-disable')
+          this.$refs.tooltip.$emit('enable')
         } else {
-          this.$root.$emit('bv::disable::tooltip', 'tooltipButton-disable')
+          this.$refs.tooltip.$emit('disable')
```
mosinve: should Popover docs be changed this way too?

mosinve:
I don't see any issues by now.

tmorehouse:
We can merge into dev, and adjust later if needed.

PR #1389, opened 2017-11-21
fix(button): allow custom size classes to be passed to the size prop

tmorehouse:
Removed the validator for the size prop, so that people can use custom `.btn-{size}` classes


tmorehouse:
I've created PR #1387 which adds a 'disabled' prop to the component versions of tooltip and popover Take a look when you get a chance.

PR #1399, opened 2017-11-23
fix(table): use stable sort algorithm to prevent SSR issues

tmorehouse:
Different JavaScript environments implement native array sorting differently, and in some cases are implemented with an unstable/inconsistent routine (Chrome and V8 are examples where native sort is inconsistent/unstable).

This can produce SSR client mismatch errors in different environments when  `sortBy` is preset on initial server render of `<b-table>`, where the server native sort uses one method while the client native sort uses a different algorithm.

This PR introduces a new utility function `stableSort`, of which `<b-table>` will use instead of native sort.

`stableSort` uses the native sort, and will fallback to sorting by index when the entries compare equal. This introduces a performance penalty over native sort, but prevents SSR DOM mismatch errors.

### DIFF
--- a/src/utils/stable-sort.js
+++ b/src/utils/stable-sort.js
@@ -1,0 +1,24 @@
+/*
+ * Consitant and stable sort function across JavsaScript platforms
+ *
+ * Inconsistant sorts can cause SSR problems between client and server
+ * such as in <b-table> if sortBy is applied to the data on server side render.
+ * Chrome and V8 native sorts are inconsistant/unstable
+ *
+ * This function uses native sort with fallback to index compare when the a and b
+ * compare returns 0
+ *
+ * Algorithm bsaed on:
+ * https://stackoverflow.com/questions/1427608/fast-stable-sorting-algorithm-implementation-in-javascript/45422645#45422645
+ *
+ * @param {array} array to sort
+ * @param {function} sortcompare function
+ * @return {array}
+ */
+
+export default function stableSort (array, compareFn) {
+  return array
+    .map((a, index) => [ index, a ])
+    .sort(function (a, b) { return this(a[1], b[1]) || a[0] - b[0] }.bind(compareFn))
+    .map(e => e[1])
+}
--- a/src/utils/stable-sort.js
+++ b/src/utils/stable-sort.js
@@ -17,6 +17,10 @@
  */
 
 export default function stableSort (array, compareFn) {
+  // Using `.bind(compareFn)` on the wrapped anonymous function improves
+  // performance by avoiding the function call setup. We don't use an arrow
+  // function here as it binds `this` to the `stableSort` context rather than
+  // the `compareFn` context, which wouldn't give us the performance increase.
   return array
     .map((a, index) => [ index, a ])
     .sort(function (a, b) { return this(a[1], b[1]) || a[0] - b[0] }.bind(compareFn))
--- a/src/components/table/table.vue
+++ b/src/components/table/table.vue
@@ -62,7 +62,7 @@
 </style>
 
 <script>
-  import { warn, looseEqual, KeyCodes } from '../../utils'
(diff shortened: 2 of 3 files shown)


src/utils/stable-sort.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {array} array to sort
+ * @param {function} sortcompare function
+ * @return {array}
+ */
+
+export default function stableSort (array, compareFn) {
+  return array
+    .map((a, index) => [ index, a ])
+    .sort(function (a, b) { return this(a[1], b[1]) || a[0] - b[0] }.bind(compareFn))
```
mosinve: This works the same way:
`.sort( (a, b) => compareFn(a[1], b[1]) || (a[0] - b[0]) )`

tmorehouse: Binding the compareFn as the `this` context speeds up performance about 5%

mosinve: did you test it? o_O

mosinve: with arrow funcs stableSort execution times between 0,21 - 1 ms.
with bind(): 0,32 - 1.4 ms

mosinve: In chrome

mosinve: and overall sort-click behavior average execution time less with arrow funcs

mosinve: `.sort((a, b) => compareFn(a[1], b[1]) || (a[0] - b[0]))`
image

`.sort(function (a, b) { return this(a[1], b[1]) || a[0] - b[0] }.bind(compareFn))`
image

mosinve: `.sort(function (a, b) { return this(a[1], b[1]) || a[0] - b[0] }.bind(compareFn))`
bind

`.sort((a, b) => compareFn(a[1], b[1]) || (a[0] - b[0]))`
arrow

tmorehouse: Based on the original source: https://stackoverflow.com/questions/1427608/fast-stable-sorting-algorithm-implementation-in-javascript/45422645#45422645
>Using .bind(compareFunction) on the wrapped anonymous function within stableSort() boosted that relative performance from about 38% by avoiding an unneeded scoped reference to compareFunction on each call by assigning it to the context instead.

mosinve: It compares with native `Array.sort()`, not the arrow functions implementation. 
And besides, after babel arrow funcs wil get transpiled to bind.... 
It's all about readability

tmorehouse: But the babel transpile will not bind the `compareFn` function as `this`, babel (transpiling the arrow function) will set `this` to the `this` context of the location it was defined, which is the outer `stableSort` function context.

This code explicitly sets `this` as the `compareFn`, so saving the function call setup on each iteration of the inner loop.

Since this function will be called every time the table needs updating (when there is a sort change), speed makes the difference with large datasets.

mosinve: https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

tmorehouse: > Note that `this` in a method shorthand or function refers to the object literal itself whereas **this in an arrow function refers to the scope _outside_ the object literal.**

In this case, we do not want `this` to refer to the `stableSort` context (the outer function), we want `this` to refer to the `compareFn` (so calling `this(a, b)` actually calls `compareFn`, without the overhead of calling `compareFn` each iteration fo the `.sort(..)` routine)

tmorehouse: I have added  comment above the code to explain the reasoning.

tmorehouse: Note that a stables sort is *much* less efficient than using QuickSort (which is what V8 and Chrome use if the array size is > 10), so we are taking a larger performance hit by implementing stable sort in order to avoid SSR errors.  So being able to squeeze some speed improvements is needed.

PR #1401, opened 2017-11-24
fix(carousel): uncaught typeerror on empty slides.

bharathvaj1995:
Add extra validation for slides Uncaught TypeError: Cannot set property 'tabIndex' of undefined in pause() function

### DIFF
--- a/src/components/carousel/carousel.js
+++ b/src/components/carousel/carousel.js
@@ -286,7 +286,7 @@ export default {
       if (this.isCycling) {
         clearInterval(this.intervalId)
         this.intervalId = null
-        if (this.index < this.slides.length) {
+        if (this.slides[this.index]) {
           // Make current slide focusable for screen readers
           this.slides[this.index].tabIndex = 0
         }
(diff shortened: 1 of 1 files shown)


tmorehouse:
Could you provide a code sample which causes your issue?

bharathvaj1995:
` <template v-for="slide in slides">
        <b-carousel-slide :caption="slide.title" :key="slide.id"
                  :text="slide.tagline"
                  :img-src="slide.cover_image"
        ></b-carousel-slide>
      </template>
<script>
export default {
    data() {
        return {
            slides: []
        };
    },
    created() {
        this.loadSliders();
    },
    methods: {
        loadSliders() {
           // code to fetch slides from the server
            });
        }
    }
};
</script>
`

tmorehouse:
You should probably be using the `mounted()` hook instead of the `created ()` hook in this case, as the carousel has not been mounted in the DOM at `created()`

bharathvaj1995:
It gives me the same error. 
err

Issue #1405, opened 2017-11-25
Applying a class to table or tbody


tristanbes:
Hello,

After reading the doc and trying on my application I have two questions:

1/ How to apply a class on the table and **only** use that class ? because the `class` props will add the custom class to the "default" bootstrap4 and I don't want that. Ex: 
```html
<table class="my-custom-class c-table table b-table table-striped table-hover></table>
// instead of 
<table class="my-custom-class"></table>
```

2/ How to apply a class on the `tbody` node ? I don't see any options to achieve that.

Thank you

tmorehouse:
Currently there isn't an option to add a specific class to the tbody, although you can target the tbody using standard CSS selectors. (i.e. `table.my-table > tbody { ... }`.

`<b-table>` will always add the Bootstrap `table` class (and the bootstrap-vue custom class `b-table`, mainly for targeting the sort styling).  THe other classes (such as `table-hover`, `table-striped` are controlled via the respective props.

tristanbes:
thanks for the quick reply. so there's no way we have total control of the classes ? that's sad :(. The integration is already done, so having to adapt the already existing css code because of (any) library is not a great idea.

The table looks like

```
<table class="c-table">
        <thead class="c-table__header">
            <tr class="c-table__row">
                <th class="c-table__cell">Header</th>
            </tr>
        </thead>
    <tbody class="c-table__body">
        <tr class="c-table__row">
            <td class="c-table__cell">
                Example cell
            </td>
    </tbody>
</table>
```

The only thing I managed to do is the class on the cell itself :(. Is there any hacks involving not touching the css/selector itself.

Thank you ;-)

tmorehouse:
As it generates a bootstrap V4 table, it will add the bootstrap v4 `table` class.

But if you are defining these custom table/row/cell classes in your app, then it shouldn't be too difficult to create selectors that override bootstrap's CSS when a specific class is applied to the outer `b-table`

tmorehouse:
As a side note, you can apply additional custom classes to the outer table element (the root element of `<b-table>`.  Any attributes that are not props (plus class and style attributes) will be directly placed on the `<table>` element (although the bootstrap class `table` will always be applied),

You can also use the `fields` definition array of objects to apply custom classes to the data row cells (`<td>`), and to the header/footer `<th>` cells.

We just don't have an option for applying any classes to the `<thead>`, `<tbody>`, or `<tfoot>` elements.

[...]

PR #1407, opened 2017-11-25
feat(table): add responsive stacked table option

tmorehouse:
This adds a new prop `stacked`, which allows the table to render in a visually stacked format based on breakpoint. (custom CSS assumes standard bootstrap breakpoint values).

If no value is passed to `stacked`, the table will always be rendered as stacked.

If `stacked` is set to a breakpoint of `sm`, `md`, `lg`, or `xl`, then the table will only appear stacked at view ports below the specified breakpoint.

Example stacked table:
image

When visually stacked, the table thead and tfoot will not be shown.  Cells will be prefixed with the fields' label (not the scoped HEAD slot). Because of this, it is not possible to click the "headings" to sort the columns.  Users however, can provide an external control (such as b-form-select) to control the sorting and sort direction (via b-table's `sort-by` and `sort-desc` syncable props).

Also optimized the thead and tfoot rendering via the user of a factory function, which reduces the code-base slightly.

Our custom CSS has increased to 8.5kbytes, due to the CSS required for each stacking breakpoint unfortunately,


tmorehouse:
Added new props to allow user to add classes to `thead`, `tfoot`, `tbody`, as well as the `tr` elements.

Addresses issue #1405

tmorehouse:
Move script contents into external file so that code transpiles in `/es` buld dir

PR #1412, opened 2017-11-28
feat(form-group): new prop for label-class, deprecate prop feedback in favor of invalid-feedback

tmorehouse:
Adds a new prop `label-class` which allows the user to apply additional classes to the label (i.e. responsive padding or alignment classes)


tmorehouse:
Also deprecated the prop `feedback` in favor of `invalid-feedback` (to align with the `valid-feedback` prop.

PR #1414, opened 2017-11-29
chore: update changelog headings

tmorehouse:


tmorehouse:
Should I squash and merge, regular merge, or rebase merge?

tmorehouse:
standard-version-writer has an option for using custom templates to generate the change-log, but not sure if there is an easy way get them set.

pi0:
For merges, i think rebase would be the best option for keeps both master and dev in sync. 

Also will try to investigate more about `standard-version-writer`. Thanks for the tip :)

tmorehouse:
stndard-version-writer (used under the hood) uses handlebars templates (4 of them) to layout the changelog.

Would just need to tweak how it set the heading levels/names, like including the version number of the sections `fxes`, `features`, `performance` etc.

But I couldn't find any CLI options on standard-version cmd to set the templates.

Issue #1434, opened 2017-12-06
Popover position not changed


josespinal:
Hi,

Recently I updated after a couple of weeks not touching this project. Now the popovers show not on the side (left, right) of the target element but on top instead.

Also placement has changed, had to set "left" when previously I had "right" to show the popover on the right side.

Thanks.

Previously:
![image](https://user-images.githubusercontent.com/10059/33669350-9d051502-da78-11e7-9c3a-2f1c512a411c.png)

Now:
![image](https://user-images.githubusercontent.com/10059/33669286-68000d94-da78-11e7-896d-983a006f1829.png)

**Here are my options:**
```html
<b-popover :target="'popover-' + product.id"
            triggers="click"
            placement="left"
            container="order-list"
            ref="popover"
            @show="onShow(index, $event)">
</b-popover>
```
- Operating system: Mac OS X
- Browser: Chrome
- Version of Bootstrap-Vue: Latest

tmorehouse:
The default placement for popovers is `right`, but note that popovers will "flip" to the opposite side if there is not enough room in the container for the requested placement (i.e. if it is too close to the edge of the viewport)

Placement is also affected by where the trigger element is located, and the positioning of the container that the trigger element is in (relative, absolute, etc, as well as any css `overflow` settings)

tmorehouse:
Could you provide a sample fo the additional markup around your popover/trigger elements?

josespinal:
Made a CodePen replicating some. You can see it full page here: [https://codepen.io/josespinal/full/EbzdXM/](https://codepen.io/josespinal/full/EbzdXM/)

I've noticed something, if I remove scroll/auto from 'overflow-y' it is positioned as it did before:
You can check it out here (line 53 of CSS):
[https://codepen.io/josespinal/full/mqYzxO](https://codepen.io/josespinal/full/mqYzxO)

.order-list-wrapper {
   overflow-y: auto;
}

**Version used before was 1.0.0, now it's 1.3.0**

tmorehouse:
Popper.js has issues with elements placed inside elements with overflow settings (the popper version has changed since version 1.0.0)

Maybe not using the container option might work better. Or position it below the table.

josespinal:
Not using the container option, didn't make any change. Not onto trying positioning below the container but I'm a little stuck on referencing the target from outside the v-for loop.

tmorehouse:
[...]

Issue #1163, opened 2017-10-04
Big <b-dropdown>s are unusable when used inside an `overflow:scroll` container, due to being constrained inside it


ronjouch:
This is a followup to https://github.com/twbs/bootstrap/issues/24251 .

## Problem

When using a [bootstrap-vue (1.0.0-beta.9) `<b-dropdown>`](https://bootstrap-vue.js.org/docs/components/dropdown) in an element that is a child of an `overflow:scroll` element, the dropdown is "stuck" inside the container, making it unusable if the dropdown contents are too big horizontally/vertically (see demo at [JSFiddle/ronj/t6z6Lnfb](https://jsfiddle.net/ronj/t6z6Lnfb/)):

![overflow-scroll_bootstrap-dropdown-fs8](https://user-images.githubusercontent.com/522085/31194832-54d24bda-a916-11e7-9ad3-85541dd35f62.png)

Contrarily, a vanilla HTML `<select>` will merrily bleed out of the `overflow:scroll` element as necessary:
![overflow-scroll_regular-select-fs8](https://user-images.githubusercontent.com/522085/31194823-50451b9c-a916-11e7-85be-41e113fc0895.png)

## Native Bootstrap workaround & need help for bootstrap-vue

Over at https://github.com/twbs/bootstrap/issues/24251#issuecomment-334271729 , @danijelGombac suggests to reparent the dropdown to `body`:

```js
$(document).on('show.bs.dropdown', '.scrolled-heighted-box', function(e) {
    var dropdown = $(e.target).find('.dropdown-menu');
    dropdown.appendTo('body');
    $(this).on('hidden.bs.dropdown', function () {
        dropdown.appendTo(e.target);
    })
});
```

... and it works great in pure non-Vue Bootstrap! Now, back to my `bootstrap-vue` project I'm able to successfully listen to the equivalent and documented `$root / bv::dropdown::shown`  event, but then I'm lost in Vue forum posts mentioning `$el` / `$mount()` / `$appendTo`.

→ 1. Can anyone help me port this workaround to `bootstrap-vue`? If not, forgetting this re-parenting trick, any other idea to fix the initial problem?

→ 2. If a solution to 1. above is found, it would be awesome if it was encapsulated behind a `<b-dropdown>` boolean property (name proposal: `reparent-to-body`?) to do that easily. I'll be happy to submit a PR with docs.

## Use case

[...]

PR #1439, opened 2017-12-07
feat(tooltip+popover): add boundary element config option (position constraint)

tmorehouse:
Adds new prop `boundary` to component versions - can be either `scrollParent` (Popper default), `window`, `viewport`, or an element reference

Adds modifiers `window` and `viewport` to directive versions (if not present defaults to scrollParent)

Addresses #1434 

A similar issue (#1163 and https://github.com/twbs/bootstrap/issues/24251) applies to dropdowns, which will be addressed in a separate PR #1440  PR https://github.com/twbs/bootstrap/pull/24979 addresses the issue in Bootstrap V4 official.


tmjoen:
Does this address the problem with popovers not being centered anymore?

There was a regression somewhere between the alphas and the current version which messes with the centering of the popover.

This is from `v4-alpha.6`

image

And this is from `1.3.0`

image

tmorehouse:
I believe alpha 6 used the old popover code (pre- Popper.js).  There is/was an issue with how the centering calculation occurred on the Bootstrap V4.beta Popper code, and there was a PR/Issue at twbs/bootstrap that mentioned this https://github.com/twbs/bootstrap/issues/23793

This PR doesn't address that issue, it addresses the issue when the trigger element is placed inside an container element that has `overflow: scroll` set, which prevents the tooltip/popover from being placed correctly (due to being constrained to the inside of the scrollable element.

PR #1444, opened 2017-12-09
fix(list-group): disabled button items

tmorehouse:
When `tag` was set to `'button'` and the `disabled` prop was also set, the rendered button was not being disabled (although it had the disabled class and appeared disabled).  This fix ensures the disabled attribute is set when rendered as a button and the disabled prop is true.

Adds new prop `button`, as a shortcut, to force render as a `button` list-group-item-action

Also extends the documentation with additional examples

### DIFF
--- a/src/components/list-group/README.md
+++ b/src/components/list-group/README.md
@@ -1,27 +1,197 @@
 # List group
 
 > List groups are a flexible and powerful component for displaying a series of content.
-List group items can be modified and extended to support just about any content within.
-They can also be used as navigation with the right modifier class.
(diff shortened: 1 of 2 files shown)


tmorehouse:
I'm wondering if maybe we should add in a boolean `button` prop to force the list-item to be rendered as a button?

alexsasharegan:
@tmorehouse , that seems like a good idea. For links, is the component rendering based on the presence of an `href` or `to` prop?

tmorehouse:
Yeah, It renders a `b-link` if either `to` or `href` is set

PR #1493, opened 2017-12-25
Allow a callback for the tooltip target attribute.

michaelzangl:
I had a problem using the tooltip component on a column: When referencing the target by ID, the tooltp always appeared at the first column.

An alternative is to reference the target using $refs. I tried to implement this in a universal way by allowing a callback being passed as target attribute.

That way, the following code works:
```html
<td ref="contactColumn">
  column content
  <b-tooltip :target="() => $refs.contactColumn">
    complex tooltip content
  </b-tooltip>
</td>
```


tmorehouse:
Tooltips will accept a reference  an element, or an id, but either method the element must exist in the document before it can find the element to attach the event listeners to.

michaelzangl:
Yes, that was exactly what my problem was: The references were not parsed by VUE before the attributes were evaluated. Using the callback solved it for me.

If you have any other working/clean way of writing those, I would appreciate it:
```html
<b-tooltip :target="() => $el">
```
```html
<b-tooltip :target="() => $refs.parentElement">
```

(Background: I need complex tooltips that contain some html / other components and I need the tooltip on the root element of the component, which is a td in many cases)

PR #1498, opened 2017-12-28
fix(dropdown): fixed aria-labbeledby for non-split dropdowns

larionov:


src/components/dropdown/dropdown.js
```diff
@@ -62,7 +62,7 @@ export default {
         class: t.menuClasses,
         attrs: {
           role: t.role,
-          'aria-labelledby': t.safeId(split ? '_BV_toggle_' : '_BV_button_')
+          'aria-labelledby': t.safeId(this.split ? '_BV_toggle_' : '_BV_button_')
```
tmorehouse: This should be `t.split`, and not `this.split` for minification reasons.

larionov: fixed in #1500

PR #1499, opened 2017-12-28
fix(id): fixed check for _uid in id generator

larionov:
On the first run _uid === 0 and this value fails the check not generating an id for the component.

### DIFF
--- a/src/mixins/id.js
+++ b/src/mixins/id.js
@@ -16,7 +16,7 @@ export default {
     }
   },
   mounted () {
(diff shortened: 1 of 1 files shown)


tmorehouse:
Usually, the root vue instance has a uid of 0, and your components would have 1 and up, but this pr is a good additional check

larionov:
The problem occured when I tried to extend the component and then run it in
my unit tests. In cases when the component was first in order there was an
error because id wasn't generating correctly.

PR #1505, opened 2018-01-03
update input group addon changes from bootstrap beta 3

toadkicker:
This addresses changes introduced in https://github.com/twbs/bootstrap/pull/25020

* drops the `.input-group-addon` class and replaces it with `.input-group-prepend` and `.input-group-append` depending on the value of the slot
* deprecates the `.input-group-btn` class and uses the classes above
* defaults component to `input-group-append` if no slot value provided
* updates specs

### DIFF
--- a/src/components/input-group/input-group-addon.js
+++ b/src/components/input-group/input-group-addon.js
@@ -18,7 +18,7 @@ export default {
     return h(
       props.tag,
       mergeData(data, {
-        staticClass: 'input-group-addon',
+        staticClass: `input-group-${data.slot === 'left' ? 'prepend' : 'append'}`,
         attrs: { id: props.id }
       }),
       children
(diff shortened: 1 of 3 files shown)


toadkicker:
I don't know if this should just become two separate components instead of using a slot/prop definition, e.g `<input-group-prepend>` vs current method. If that is preferred over this solution I'm happy to provide it.

pi0:
Thanks. I think this change was *necessary* to provide this update with minimum code change impact. But adding support for `<input-group-append|prepend>` would be much better for long-term. So PR welcomes for those components.

toadkicker:
I got a chance to check this out more today. The new style needs to have a child element of `div.input-group-text` in order for the styles to get picked up. This change does not insert the additional div around the text. I'll figure out how to do that next.

PR #1506, opened 2018-01-03
update checkbox changes from bootstrap beta 3

toadkicker:
(this commit does not change the html structure however)


pi0:
Thanks. I've rebased PR into beta-3 branch so we can finalize everything before release.

toadkicker:
Ok I will most likely put more PR's up to support this as its a blocker in my project. Would you prefer I target beta-3 branch instead of dev?

pi0:
@toadkicker Yes, please :)

PR #1523, opened 2018-01-10
Fix #1522: Import button-close in alert with a name matching tag.

### LINKED ISSUE #1522: Alert cannot load b-button-close when imported as an individual component
Hi,

When importing only b-alert, if the dismissed prop is set like this: 

```

<template>
  <b-alert :show="true" :dismissible="true">Message</b-alert>
</template>
<script>
  import bAlert from "bootstrap-vue/es/components/alert/alert";
  export default {
    name: 'Dummy',
    components: {
        bAlert
    }
}
```

An error is throwed :

> [Vue warn]: Unknown custom element: <b-button-close> - did you register the component correctly? For recursive components, make sure to provide the "name" option.

This is because the button si imported as bBtnClose (from button/button) and used with b-button-close.

mehdirande:
bButtonClose component was imported in alert with the shortcut alias
(bBtnClose) but used with the full tag (b-button-close). When importing
b-alert as an individual component b-button-close alias was not set and
vue failed to import the b-button-close component.
Importing bButtonClose component with the fullname fix #1522 .

### DIFF
--- a/src/components/alert/alert.js
+++ b/src/components/alert/alert.js
@@ -1,7 +1,7 @@
-import bBtnClose from '../button/button-close'
+import bButtonClose from '../button/button-close'
 
 export default {
-  components: {bBtnClose},
+  components: {bButtonClose},
   render (h) {
     if (!this.localShow) {
       // If not showing, render placeholder


pi0:
Thanks.

PR #1538, opened 2018-01-16
Feat/table multi sort

rubybrowncoat:
Breaking change to table sorting to support multi-column sorting.

The sortBy property changes from a string to an array of objects describing the current sorted columns. Move the sortCompare function to the array of field objects to provide column specific sortCompare functions, and use the default sortCompare when the field doesn't provide its own.

This is a feature I found a need for in the current project I'm working on. I tried to go around the table internal sorting mechanisms at first, but it wasn't great and it took a lot away from the greatness that is bootstrap-vue. In the end I made this modification and maybe there is interest in this specific functionality?

Hope we can make it even better.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "lodash.startcase": "^4.4.0",
     "opencollective": "^1.0.3",
     "popper.js": "^1.12.9",
+    "thenby": "^1.2.3",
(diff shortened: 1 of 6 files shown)


pi0:
Closing due inactivity.

Issue #1550, opened 2018-01-22
Feature request - Add classes to auto-generated elements


ghenry:
Hi all,

As per my chat in the slack channel:

Hi all, I'm trying to add an additional class to a template button slot on a dropdown - <template slot="button-content">

This gets generated: 
`
<a href="#" id="profileDropdown__BV_button_" aria-haspopup="true" aria-expanded="true" class="nav-link dropdown-toggle"><img src="blah"></a>`

but I want to do this:

`<a href="#" id="profileDropdown__BV_button_" aria-haspopup="true" aria-expanded="true" class="nav-link nav-link-profile dropdown-toggle"><img src="></a>`

ie add nav-link-profile to that a href

I think a mixin or a way to get at these all these automatic class elements would be perfect. Video attached:

[class_toggle.zip](https://github.com/bootstrap-vue/bootstrap-vue/files/1651630/class_toggle.zip)

ghenry:
Hi all,

I'm done my best to put this in the right place, add docs and a test for it. 
I feel this should be more generic though as there should be a prop to do this on every auto-generated child element etc. I've done what I need, now onto #1553 

Thoughts?

Thanks,
Gavin.

PR #1555, opened 2018-01-23
feat(): add a extra-toggle-classes prop to nav-item-dropdown.js with …

ghenry:
…tests and docs. Refs #1550

### DIFF
--- a/src/components/nav/README.md
+++ b/src/components/nav/README.md
@@ -121,7 +121,7 @@ Use `<b-nav-item-dropdown>` to place dropdown items within your nav.
 <b-nav pills>
   <b-nav-item active>Active</b-nav-item>
   <b-nav-item>Link</b-nav-item>
-  <b-nav-item-dropdown id="nav7_ddown" text="Dropdown" right>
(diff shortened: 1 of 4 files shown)


pi0:
Hey. Thanks for this PR. What's your idea if we simplify prop name `extra-toggle-classes` into `toggle-class`?

ghenry:
Hi!

Sounds good, but it sounds more like an override rather than an additional
class. Ideally I'd like to have a common prop available for every type of
component like this. I only spent last night and this morning looking
around the project design to do enough to get this to work.

I know I'll need more of these migrating my static bs4 theme to bs vue.

Thanks.

pi0:
Seems reasonable for naming. Thanks again for PR. Will be available soon in `2.0.0-rc`.

pi0:
Update: it is available with both `v1.5.0` and `v2.0.0-rc.0` now :)

ghenry:
I have a doc correction:

> that by default has the classes 

should be

> that by default have the classes 

best way to fix? Another PR?

pi0:
Ah fixed. Thanks for notifing :)

ghenry:
Have now upgraded and working perfectly. Thanks for building such a great
project.

Issue #1548, opened 2018-01-20
Components import unused dependencies


psi-4ward:
Example for form-checkbox:

`import bCheckbox from 'bootstrap-vue/src/components/form-checkbox/form-checkbox';`

results in addition of **~49kb minified** cause it includes popper.js, dropdown.js, pagination.js and other deps which are not needed for the custom-form-checkbox

Trainmaster:
@psi-4ward Did you try `import bCheckbox from 'bootstrap-vue/es/components/form-checkbox/form-checkbox'`

psi-4ward:
Yes I did. Same result.

![screenshot_20180125_130743](https://user-images.githubusercontent.com/1191572/35387609-bc4e05ac-01d0-11e8-96e2-1c244c5a5f8b.png)

We should change the imports:
```js
// import { idMixin, formRadioCheckMixin, formMixin, formSizeMixin, formStateMixin, formCustomMixin } from '../../mixins';
// import { looseEqual } from '../../utils';

import { isArray } from '../../utils/array';
import idMixin from '../../mixins/id';
import formRadioCheckMixin from '../../mixins/form-radio-check';
import formMixin from '../../mixins/form';
import formSizeMixin from '../../mixins/form-size';
import formStateMixin from '../../mixins/form-state';
import formCustomMixin from '../../mixins/form-custom';
import looseEqual from '../../utils/loose-equal';
```

to get this result

![screenshot_20180125_131409](https://user-images.githubusercontent.com/1191572/35387834-a220eb6c-01d1-11e8-948c-bc168a9547de.png)

psi-4ward:
I don't know why tree-shaking does not work here.

psi-4ward:
yay NICE!

PR #1557, opened 2018-01-24
Call removeEventListener on the right element

### LINKED ISSUE #1391: Possible memory leak with dropdown and vue-router
Hello guys, I have found a memory leak when using the drop-down component and switching between routes.
Here is a reproduction link https://jsfiddle.net/tochoromero/1tup51he/2/

If you use the link to move back and forth and take a look at the memory usage you will see that the VueComponents are not being garbage collected.

I initialize opened an issue against the Vue repo, but they suggested to open it here instead, here is the link to the other issue https://github.com/vuejs/vue/issues/7101.

I hope you guys have an idea of what is going on and can provide an isolated reproduction example without having to add bootstrap-vue.

tochoromero:
When removing the click listener, we were calling `removeEventListener` on the wrong element, and this was causing a memory leak.
fixes #1391

### DIFF
--- a/src/mixins/clickout.js
+++ b/src/mixins/clickout.js
@@ -6,7 +6,7 @@ export default {
   },
   beforeDestroy () {
     if (typeof document !== 'undefined') {
-      document.removeEventListener('click', this._clickOutListener)
+      document.documentElement.removeEventListener('click', this._clickOutListener)
     }
   },
   methods: {


pi0:
Thanks ❤️

tochoromero:
You are welcome, any idea when you are planning a new release?

pi0:
Do you need any releases for old 1.x beta branch or 2.x?

tochoromero:
I personally don´t need the old 1.x branch to be patched, but if there are
people who cannot update to the latest version they will greatly benefit
from the fix...

pi0:
bootstrap-vue@1.5.1 (beta) just released with this patch. Will release rc.2 too after addressing #1548 .

tochoromero:
Thank you :)

pi0:
v2.0.0-rc.1 released too :)

tochoromero:
Awesome, thanks.

PR #1587, opened 2018-02-04
fix(Table) allow empty labels

### LINKED ISSUE #1582: The table uses key for the label when assigned an empty string
In the table component, the label is assigned as follows.
```javascript
f.label = f.label || startCase(f.key)
```
[(Permalink to this line in the table component)][1]

But evaluating f.label this way means that the empty string will be falsy and f.key will be used instead.

An empty label is common in actionable columns where a header may not be desirable. It would be reasonable to assume that a user would provide an empty string in this case, but doing so provides a confusing result.

The documentation notes that the label, "Defaults to the field's key (in humanized format) if not provided." But that is deceptive because a standard user would assume providing an empty string is providing a value.
[(Link to this section in the documentation)][2]

I suggest modifying this assignment to allow empty string, and making it clear in the documentation that any other falsy value will result in the key being used.

[1]: https://github.com/bootstrap-vue/bootstrap-vue/blob/e0e40069b2e75514f07edcc6927bd358d34f9f0c/src/components/table/table.js#L771
[2]: https://bootstrap-vue.js.org/docs/components/table/#field-definition-reference

mosinve:
fixes #1582

### DIFF
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -264,7 +264,7 @@ The following field properties are recognized:
 | Property | Type | Description
 | ---------| ---- | -----------
 | `key` | String | The key for selecting data from the record in the items array. Required when passing the props `fields` an array of objects.
-| `label` | String | Appears in the columns table header (and footer if `foot-clone` is set). Defaults to the field's key (in humanized format) if not provided.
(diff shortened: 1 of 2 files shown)


tmorehouse:
Note this can break ARIA compliance when a column doesn't have a header or an aria label for blind users, as they cannot see the visual context of the column, they need a label.

The field's key should be used as the aria-label on the `<th :aria-label="field.key">`, and the td's in hte column will need an aria-label when in stacked mode (since the thead is hidden).

PR #1589, opened 2018-02-04
Fix table html injection

### LINKED ISSUE #1554: Table silently allows HTML injection
Hi,

as I just noticed it's possible to inject HTML-Tags and therefore also CSS into Vue-based web applications without noticing it.

Here is an example: https://jsfiddle.net/sfc51gdx/4/

Users who start building web applications probably start by using `Variable: {{myVariable}}` to output data and then start thinking, that VueJS is allways safely escaping HTML and that they don't need to think any longer about it. That's fine because the VueJS-Escaping seems to work pretty good unless you explicitly allow HTML-Injection with a slightly different Vue syntax.
But Bootstrap-Vue behaves completely reversed and allows HTML-injection by default and the programmer probably won't ever notice this security issue.

Of course there should be any kind of server-side input validation for any data entered by the user, but still this is a possible security issue.

I'd suggest to escape any content in tables (and probably more VueJS-components?) by default and allow  HTML-parsing by adding a parameter or something like that.

mosinve:
fixes #1554

### DIFF
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -256,7 +256,7 @@ export default {
 ```
 
 >**Note:** _if a `key` property is defined in the field definition, it will take
-precidence over the key used to define the field._
+precedence over the key used to define the field._
 
 ### Field definition reference
 The following field properties are recognized:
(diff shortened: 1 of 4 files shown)


pi0:
Thanks vitaly :)

PR #1621, opened 2018-02-17
Accept an HTMLElement as a valid target for tooltips and popovers

elghobaty:
While it works, passing an `HTMLelement` (e.g. an `HTMLButtonElement`) to a tooltip/popover raises the following warning:
```[Vue warn]: Invalid prop: type check failed for prop "target". Expected String, Object, got HTMLButtonElement. ```

### DIFF
--- a/src/mixins/toolpop.js
+++ b/src/mixins/toolpop.js
@@ -35,7 +35,7 @@ export default {
   props: {
     target: {
       // String ID of element, or element/component reference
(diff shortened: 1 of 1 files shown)


pi0:
This breaks SSR as we don't have `HTMLElement` in node :)

mosinve:
this whole PR or this usecase?

PR #1640, opened 2018-02-28
Always pass selectSize to selects

### LINKED ISSUE #1639: Form Select select-size does not apply in single mode
Operating System: Windows
Browser: Chrome 64, Edge
Bootstrap-Vue Version: 2.0.0-rc1

I need a select list style, not a dropdown, in single mode. The documentation leads me to believe that using the select-size prop is the way to do this. It has no effect regardless of value.

Example : 
`  <b-form-select v-model="selected" :selected-size="20" :options="options" class="mb-3"></b-form-select>`

This problem is also view able in the example on the documentation page as not rendering correctly.

robcresswell:
The size attribute is not specific to multiple-select fields: it can be
used on single selects on both Bootstrap and raw HTML. We could probably
pass the value straight through instead of renaming it "selectSize",
but I'd like to avoid making a breaking API change at this stage.

Fixes #1639

### DIFF
--- a/src/components/form-select/form-select.js
+++ b/src/components/form-select/form-select.js
@@ -42,7 +42,7 @@ export default {
           id: t.safeId(),
           name: t.name,
           multiple: t.multiple || null,
-          size: t.isMultiple ? t.selectSize : null,
+          size: t.selectSize,
           disabled: t.disabled,
           required: t.required,
           'aria-required': t.required ? 'true' : null,


jslewis90:
The docs list this property as not being required. And yet, after this change, the arrows on the right hand of single selects are no longer properly displayed without explicitly declaring their select-size value of 1.

Therefore, can this either be reverted or have the property use 1 as its default value?

PR #1654, opened 2018-03-07
Manually add rc2 release notes

robcresswell:
convential-changelog seems to be failing to handle our commit logs, probably due to bad formatting on our part. Manually adding some important release notes. I've left out all the minor "fix typo" commits, though I can go back and add them if there is value in that.


pi0:
This is great! Merci :)

Issue #1658, opened 2018-03-07
Missing arrows in Form Select in v2.0.0-rc.2


x42y:
Version: 2.0.0-rc.2

As seen here: https://bootstrap-vue.js.org/docs/components/form-select/ the arrows in the select dropdowns are missing in rc 2.

robcresswell:
Ack, I think this is something I've changed. I'll take a look later today.

robcresswell:
Yeah, this is my fault, I've found the problem, fixing now.

ChrisCalculus:
Hi @robcresswell, when will this be available in master? Thanks.

robcresswell:
@ChrisCalculus When I get time to finish it? In the next couple of days I'd imagine 😄

robcresswell:
That should fix a couple of issues with select. Sorry for breaking it, was fixing something else and derped.

ChrisCalculus:
Great @robcresswel, thanks for your time and solving it.

Morgul:
I'm currently getting his by this as well, it's blocking upgrading to rc.2 for me.

robcresswell:
@Morgul: did you read the thread?

Morgul:
@robcresswell Yes, I did, I'm aware of the open merge request. But it's often useful to know how many people are effected by an issue when planning releases, etc, and my comment was just an attempt to provide that. :)

iruizeguilaz:
I have updated the version and getiing thiis solved, but someone else has come acrross a similar problem with the table columns in sorting??? I do not get the arrows there....

Jodge:
I am currently experinceing the same problem on `"vue": "^2.5.2"`

PR #1673, opened 2018-03-12
fix(select): Fix issues with form-select

robcresswell:
The form-input component currently has some arbitary restrictions around
the combinations of multiple, custom-select and selectSize that will be
applied. Previously I tried to address this by passing through the
selectSize, but this caused an issue (hiding the arrows) in custom
selects when the size was set to 0.

This patch cleans up the internal computed state of form-select to be
more relaxed about what user input is rendered in the final component

Fixes: #1658

### DIFF
--- a/src/components/form-select/form-select.js
+++ b/src/components/form-select/form-select.js
@@ -42,7 +42,7 @@ export default {
           id: t.safeId(),
           name: t.name,
           multiple: t.multiple || null,
-          size: t.selectSize,
+          size: t.computedSelectSize,
           disabled: t.disabled,
           required: t.required,
           'aria-required': t.required ? 'true' : null,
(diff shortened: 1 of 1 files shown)


pi0:
Sorry for the delay. Thanks :)

tmorehouse:
It should be noted that firefox has (or had) issues when select size is set to `1`.

pi0:
OMG @tmorehouse you are back 😍😍😍😍😍😍😍😍

robcresswell:
hahaha @pi0 I had the same reaction! (But quietly in my head)

pi0:
Let's have a chat on slack tonight, Friends 😸

tmorehouse:
I'll see if I can get my slack working... We are now using slack at work, and was able to only log into my work slack account.

mosinve:
I'm in

Issue #1649, opened 2018-03-06
Ability to use custom dropdown menu classes


RomeroMsk:
Feature request: add a `menuClass` prop that will be merged with the default classes in the `menuClasses()`.

RomeroMsk:
Let's consider this as two feature requests:

1.  Add a `menuClass` prop that will be merged with the default classes in the `menuClasses()`.
2.  Add a `menu-content` slot that will be rendered instead of `menu` (not inside, but instead).

The first one is easy, I can make a PR for this. The second one is not that important for me and can be put off for a while.

RomeroMsk:
@pi0, @robcresswell, is there anything I should change/improve in the corresponding PR to be sure it will be merged before the next release?

RomeroMsk:
Tested, thanks.

PR #1683, opened 2018-03-15
Added `menuClass` prop to `bDropdown`

RomeroMsk:
Solves #1649


pi0:
Sorry for the delay. Thanks.

PR #1704, opened 2018-03-27
Clean unnecessary measures

valexiev:
The props already have default values that cover the cases, so it is unnecessary to cover them again.
Also I have ordered the properties _value, text, disabled_ to be in the same order everywhere in the file.


pi0:
Thanks.

PR #1723, opened 2018-04-07
feat(tabs/noNavStyle): added related prop and check

syffs:
`b-tabs` is very convenient to handle content, but its style is currently limited to `nav-tabs` and `nav-pills`. 

`no-nav-style` gets rid of the mandatory `nav-${navStyle}` allowing more customization without the overhead of deflecting `nav-tabs` and `nav-pills` effects. Here's a sample:

I'd have introduced a `nav-style` or `nav-variant` prop instead, but it'd have lacked constitency with `pills` prop to remain backward compatible.

cheers

### DIFF
--- a/src/components/tabs/tabs.js
+++ b/src/components/tabs/tabs.js
@@ -102,8 +102,8 @@ export default {
       {
         class: [
           'nav',
-          `nav-${this.navStyle}`,
           {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks!

PR #1731, opened 2018-04-13
fixed 12 cols remove for custom $grid-columns

dvgamer:
for bootstrap customize.

### DIFF
--- a/src/components/form-group/form-group.js
+++ b/src/components/form-group/form-group.js
@@ -241,7 +241,7 @@ export default {
     },
     inputLayoutClasses () {
       return [
-        this.horizontal ? `col-${this.breakpoint}-${12 - Number(this.labelCols)}` : null
+        this.horizontal ? `col-${this.breakpoint}` : null
       ]
     },
     hasLabel () {


pi0:
@dvgamer I think setting `labelCols` to `0` should do the trick! Would you please explain more or provide a reproduction example for this problem? 

Another thing: I think this would be a BREAKING CHANGE because this PR removes any usage of `labelCols` which is 3 by default now.

dvgamer:
@pi0 how to fixed that. i can't setting `inputCols`.

image

pi0:
It is `labelCols` (`label-cols`)

dvgamer:
Yes, `label-cols="10"`, but input auto set 2 because `12 - this.labelCols` = 2 is wrong.
i want set input col is `col-md-26`

image

emanuelmutschlechner:
We're not gonna change this for your specific use case. But feel free to extend `b-form-group` component and modify to your needs.
```js
import bFormGroup from "bootstrap-vue/es/components/form-group/form-group.js";
import warn from "bootstrap-vue/src/utils/warn";

export default {
  mixins: [bFormGroup],
  props: {
    labelCols: {
      type: [Number, String],
      default: 3,
      validator(value) {
        if (Number(value) >= 1 && Number(value) <= 35) {
          return true;
        }
        warn("b-form-group: label-cols must be a value between 1 and 35");
        return false;
      }
    }
  },
  computed: {
    inputLayoutClasses() {
      return [
        this.horizontal
          ? `col-${this.breakpoint}-${36 - Number(this.labelCols)}`
          : null
      ];
    }
  }
};
```

pi0:
BTW if this change is necessary we are open accepting PRs :)

Issue #1735, opened 2018-04-15
Typo in tabs key event


emanuelmutschlechner:
There is a typo in `src/components/tabs/tabs.js` for key code `DOWN`:
```js
} else if (key === KeyCodes.DWON || key === KeyCodes.RIGHT) {
```

PR #1736, opened 2018-04-15
fix(tabs): typo in tabs (#1735)

emanuelmutschlechner:
Fix

### DIFF
--- a/src/components/tabs/tabs.js
+++ b/src/components/tabs/tabs.js
@@ -15,7 +15,8 @@ const bTabButtonHelper = {
     active: { type: Boolean, default: false },
     disabled: { type: Boolean, default: false },
     linkClass: { default: null },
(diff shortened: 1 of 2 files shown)


src/components/tabs/tabs.js
```diff
@@ -237,6 +242,10 @@ export default {
     navWrapperClass: {
       type: [String, Array, Object],
       default: null
+    },
+    keyNav: {
+      type: Boolean,
+      default: false
```
pi0: Disabling `keyNav` by default will be a BREAKING CHANGE. Also, accessibility enhancements (like keyboard navigation) are enabled in many other areas so is there any reason to disable it by default?

emanuelmutschlechner: That's true. I pushed a fix for this

Issue #1741, opened 2018-04-16
TypeScript definitions in 2.0.0-rc.7


FantasticFiasco:
It seems that there is a problem with the TypeScript definitions in the 2.0.0-rc.7 release:

```
ERROR in .../searchlight/node_modules/bootstrap-vue/src/index.d.ts
[tsl] ERROR in .../searchlight/node_modules/bootstrap-vue/src/index.d.ts(9,41)
      TS1005: '{' expected.
```

You can see the complete build log [here](https://travis-ci.org/FantasticFiasco/searchlight/jobs/367075194).

pi0:
@FantasticFiasco Please open a PR for fix.

pi0:
Fixed in RC.8

PR #1742, opened 2018-04-16
fix(typescript): Fixed the typescript export

Nulifier:
Changed the default export from the TypeScript definition files to use Vue's PluginObject interface. This fixes issue #1741


pi0:
Hey! It is fixed in #1744 so closing this one. BTW thanks for your contribution :)

PR #1745, opened 2018-04-17
Update README.md

mons54:


pi0:
BootstrapVue should be 100% SSR compatible. If any bug or unwanted behavior happens when using `ssr: true` please open an issue with reproduction code.

Issue #1751, opened 2018-04-18
:min :max property on b-input type="number" is ignored when 0 is passed


tarreislam:
Hello.

If you set `:min` or `:max` is set to 0 it will have no effect on `type="number"`

Example:
 `<b-form-input type="number" :min="0" :max="23"></b-form-input>`  will allow a max of 23 and ignore :min=""

However if you pass `:min` or `:max` as strings, it will work as intended but then Vue will throw `[Vue warn]: Invalid prop: type check failed for prop "min". Expected Number, got String`

im @ `2.0.0-rc.8`

emanuelmutschlechner:
Created a pull request for it #1752

emanuelmutschlechner:
@Diaskhan this was fixed and can be closed

PR #1752, opened 2018-04-18
fix(form-input): remove checks to allow setting min, max and step to 0 (#1751)

emanuelmutschlechner:


emanuelmutschlechner:
@pi0 why do we define these non functional, pure html markup attributes as props? Vue attribute inheritance takes care of it. 

Feel free to merge this pull request or revert #1734

emanuelmutschlechner:
See #1767

Issue #1512, opened 2018-01-05
Support "fade" class/property of Alerts


hmexx:
Self-explanatory. Bootstrap 4 allows you to add a "fade" class to alerts to make them go away in a more attractive fashion (fade-out). Can bootstrap-vue be made to support this?

Great work btw.

Thx

AdsonCicilioti:
Yep!
Wrapping with a `transition` Vue tag i can animate only the enter. When Dismissing Timeout ends or click on close button the alert is Destroyed, dont allow any animatiom with transition effects... 
Nice work on htis feature!!

PR #1785, opened 2018-04-28
feat(alert): Add fade prop

emanuelmutschlechner:
Closes https://github.com/bootstrap-vue/bootstrap-vue/issues/1512

### DIFF
--- a/src/components/alert/alert.css
+++ b/src/components/alert/alert.css
@@ -1,0 +1,6 @@
+.fade-enter-active, .fade-leave-active {
+    transition: opacity .15s linear;
+}
+.fade-enter, .fade-leave-to {
+    opacity: 0;
+}
--- a/src/components/alert/alert.js
+++ b/src/components/alert/alert.js
@@ -1,5 +1,7 @@
 import bButtonClose from '../button/button-close'
 
+import './alert.css'
+
 export default {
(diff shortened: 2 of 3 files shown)


src/components/alert/alert.js
```diff
@@ -16,10 +18,15 @@ export default {
         [ this.$slots.dismiss ]
       )
     }
-    return h(
+    let alert = h(
```
mosinve: I'd use  `const` here

emanuelmutschlechner:
@pi0 Sure, let me know if this is ok.

src/components/alert/alert.css
```diff
@@ -0,0 +1,6 @@
+.fade-enter-active, .fade-leave-active {
+    transition: opacity .15s linear;
+}
+.fade-enter, .fade-leave-to {
+    opacity: 0;
+}
```
tmorehouse: I wonder if we should target these classes to elements containing the class `.alert`, so that it doesn't interfere with users who may have their own fade transition with different settings?

Maybe:

```css
.alert.fade-enter-active,
.alert.fade-leave-active {
     transition: opacity .15s linear;
 }
.alert.fade-enter,
.alert.fade-leave-to {
     opacity: 0;
 }
```

PR #1788, opened 2018-04-29
feat(table): Add the sort-direction prop

### LINKED ISSUE #1783: Table: Controlling sort direction
This is a feature request.

I would like to be able to control the sort order direction, ideally something like [this](https://datatables.net/examples/advanced_init/sort_direction_control.html).

The library hardcodes none -> asc -> desc.

(I spent many hours trying to force it none -> desc -> asc, without having arrows point in the wrong direction, but never got it to work properly).

emanuelmutschlechner:
Closes #1783.

Improvements for docs are welcome


emanuelmutschlechner:
Any ideas for tests?

mpranjic:
Wow, this was fast.

How about this setting can be provided per-column instead of per-table ?

pi0:
Thanks :)

PR #1812, opened 2018-05-10
fix(tab): fix the delay in tab transition

### LINKED ISSUE #1806: fade  transition of tabs component is too slow
compare the animation speed between bootstrap and bootstrap-vue:
https://bootstrap-vue.js.org/docs/components/tabs/
https://getbootstrap.com/docs/4.1/components/navs/#nav-tab

both using `transition: opacity .15s linear;` but bootstrap-vue takes about 0.3 seconds instead of 0.15. 

seems it wait for fade out finish and then start fade in, the blank time makes user feel that content is not ready. I tried changing the time to 0.07s fadeout + 0.07s fadein, total time is 0.14s, but the blank time in the middle still feels weird.

if it could do a crossfade, and start  fade in fade out transition at same time, it would look much better

rinick:
after this change, bootstrap-vue tab transition has the same behavior as original bootstrap.
hide of the previous tab happens instantly, and the new tab have a 0.15s animation. resolves #1806

### DIFF
--- a/src/components/tabs/tab.js
+++ b/src/components/tabs/tab.js
@@ -28,8 +28,7 @@ export default {
         props: { mode: 'out-in' },
         on: {
           beforeEnter: this.beforeEnter,
-          afterEnter: this.afterEnter,
-          afterLeave: this.afterLeave
+          beforeLeave: this.beforeLeave
         }
       },
       [content]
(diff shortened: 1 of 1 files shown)


pi0:
Thanks :)

PR #1815, opened 2018-05-11
fix(dropdown): use mousedown to close dropdown

### LINKED ISSUE #1814: dropdown conflict with component that use stopPropagation
1) paste following code in https://bootstrap-vue.js.org/play/
2) click dropdown
3) click tab
drop down is not closed because tab eats the event with stopPropagation()
```
<b-tabs>
  <b-tab title="first" active>
    <br>I'm the first fading tab
  </b-tab>
  <b-tab title="second">
    <br>I'm the second tab content
  </b-tab>
</b-tabs>
<b-dropdown text="Dropdown Button">
  <b-dropdown-item>First Action</b-dropdown-item>
  <b-dropdown-item>Second Action</b-dropdown-item>
</b-dropdown>
```

I think it should close on mouseDown or mouseUp event, isntead of on click event

### LINKED ISSUE #1817: clickout.js need some optimization
right now it adds a global event listener on mount, so every time there is a click event the listeners will be called on every dropdown in the dom, even when they are display:none

to improve the performance, the listener should only be added when the dropdown is shown. and removed when dropdown is hidden

rinick:
dropdown wont close when event gets stopPropagation()
since click event stopPropagation() is used in several componenets in bootstrap-vue, it's better to switch to mousedown event to close popup.
resolves #1814 

besides that, clickout listener is now only added on demand, resolves #1817


rinick:
code coverage dropped, but I would argue this is normal.  since the test `should open only one dropdown at a time` is commented out, drop down is never open in the test, so dropdown close should not be called either. right now dropdown close is called on any click event even when dropdown is not opened , that's why the unit test hit the coverage before, which is actually a bug, reported here: https://github.com/bootstrap-vue/bootstrap-vue/issues/1817

tmorehouse:
Note multiple component use the `clickoutListener`, have you ensured that this change does not impact other components?

tmorehouse:
There are also a few conflicts with the test fixture HTML (and indirectly with the fixture test script).

Could you also adjust your test HTML and JavaScript?

tmorehouse:
And would you be able to merge the latest `dev` into your repo?  There are a few conflicts from recent changes in dropdown test fixtures.

tmorehouse:
Closing as PR #2159 fixes the issues wrt clickout listener.

PR #1825, opened 2018-05-13
replace lorempixel.com with picsum.photos, resolves #1823

### LINKED ISSUE #1823: Carousel site
https://bootstrap-vue.js.org/docs/components/carousel

It's hard to read, then image size always changing..

rinick:
lorempixel.com is down, replace them with placeimg.com


emanuelmutschlechner:
placeimg.com does not have the same categories, and serves random images instead. 

And the second and more important thing I noticed is, it serves random images all the time. I would go with picsum.photos and pick specific images from their [list](https://picsum.photos/images).

PR #1837, opened 2018-05-16
fix(modal) optimize model.resetScrollbar , resolves #1831 

### LINKED ISSUE #1831: Destorying modal is slow due to scrollbar reset
I've got a component with quite a few nested components. It's some kind of a form builder and one form contains a lot of elements and each elements as well as each row has a few options. Those options I'm hiding by default by using a bootstrap-vue modal component.

That works well and reasonably fast, but when I unload the page, things get insanely slow.
The problem starts in `resetScrollbar` which is called a lot and eventually causes `querySelectorAll` to consume a lot of CPU time. You can see it in the profile below:

image

Is there some way to get around that? Would it be possible to manually call that method if required and disable it by default? I know that I can customize the component and add a switch, but I'm looking for a solution that doesn't require me overriding code which won't get updated in the future. I'm also wondering if it wouldn't be enough to reset the scrollbar if we close one modal instead of doing so when it gets destroyed.

rinick:
modal resetScrollbar should revert the change only when the change was done in the first place

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -752,6 +752,8 @@ export default {
         const computedStyle = window.getComputedStyle
         const body = document.body
         const scrollbarWidth = this.scrollbarWidth
+        this._marginChangedForScroll = []
+        this._paddingChangedForScroll = []
         // Adjust fixed content padding
         selectAll(Selector.FIXED_CONTENT).forEach(el => {
           const actualPadding = el.style.paddingRight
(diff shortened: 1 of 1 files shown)


tmorehouse:
Hmmm, we are looking at making modal support multiple modals opened at a time, and basically
calling `setScrollbar` and `resetSCrollbar` only if the modal is the only one opened at the current time.

We may have to change how this is done, since the modal that was first opened may not be the modal that is closed last, and may not have a reference of the elements `adjusted`.

I think we will have to change from storing the remembered lists of adjustments on the modal instance (`this._xxx`) to the actual body element (`document.body._xxx`)

I'll work on a PR that can handle the new requirements, while still keeping the destroy process somewhat fast.

PR #1841, opened 2018-05-17
fix(readonly): allow readonly to be forced to false when plaintext is enabled

syffs:
plaintext could be convenient to customize `b-input` styling (for example when you need to wrap it into a div), but readonly is currently forced to true, when plaintext is enabled.

this PR introduces an indeterminate state for readonly, and hence allows it to be forced to false even with plaintext enabled.

### DIFF
--- a/src/components/form-input/form-input.js
+++ b/src/components/form-input/form-input.js
@@ -39,7 +39,7 @@ export default {
         type: this.localType,
         disabled: this.disabled,
         required: this.required,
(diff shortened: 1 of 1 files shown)


tmorehouse:
Actually plaintext is supposed to be a readonly input per Bootstrap V4 docs (it is meant to display a form control that is read only, but without the grey background., making it looks more like text.

It would be better to define override styles to override form-control using an extra class on the input (as per Bootstrap V4 custom styles ability)

PR #1845, opened 2018-05-19
fix(form-input): force update formatted value

farnabaz:
When fromatter function removes last char from value, text input does not update and character still visible in text input
We could update value of text input directly using its reference

Most of this changes done by @pi0 so all credits goes to him

### DIFF
--- a/src/components/form-input/form-input.js
+++ b/src/components/form-input/form-input.js
@@ -32,7 +32,6 @@ export default {
     return h('input', {
       ref: 'input',
       class: this.inputClass,
-      domProps: { value: this.localValue },
       attrs: {
         id: this.safeId(),
         name: this.name,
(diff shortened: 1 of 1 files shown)


ericdrobinson:
This commit broke form input elements when used in Adobe CEP 8 contexts. It results in constantly triggering [this bug](https://forums.adobe.com/thread/2422941).

Adobe CEP (Common Extensibility Platform) is the extension system that Adobe uses for current Creative Cloud applications. See [this documentation](https://github.com/Adobe-CEP/CEP-Resources/blob/master/CEP_8.x/Documentation/CEP%208.0%20HTML%20Extension%20Cookbook.md) for an overview.

**Adobe** is aware of the bug on their end. That said, **BootstrapVue** did not exhibit any related issues until this change came in. My hunch is that something around [here](https://github.com/bootstrap-vue/bootstrap-vue/pull/1845/commits/a74f755d3b0ffef22d42d4e9ed1a7af43883d602#diff-9f3870033653e73a44b1414e01d6d177R147) is the cause.

pi0:
@ericdrobinson Thanks for your report. I would be happy to release a fix ASAP. But we need more information and probably a reproduction example. Would you please open an issue with more details?

ericdrobinson:
@pi0 See 上

PawelDmochowskiDRAST:
This change introduces triggering of input-event on model changes which is against the docs and is a breaking change since it is no longer possible to distinguish between a change via user interaction and change of underlying model.

tmorehouse:
New version of form-input uses a custom event for updating the v-model, while preserving the input and change  events (and not firing extras). PR #2100 addresses most of the issues, and a new PR will be making a few adjustments.

PR #1847, opened 2018-05-22
fix(modal): fixes an issue that the dialog would not close.

### LINKED ISSUE #1826: Repeating modal's show() and hide() asynchronously, modal is not closed at the last hide().
OS: Windows 10 Pro (1709)
Browser: Firefox 60.0, Chrome 66.0.3359.170
Bootstrap-Vue: 2.0.0-rc.9

When you run the following script and press the 'start' button, modal will remain open.

```
<template>
  <div class="test">
    <b-button @click="start">start</b-button>
    <b-modal v-model="busy">{{api}}: calling...</b-modal>
  </div>
</template>

<script>
export default {
  name: 'Test',
  data () {
    return {
      API_DELAY: 100,
      WAIT: 1,
      api: null,
      busy: false
    }
  },
  methods: {
    callApi (api) {
      this.api = api
      const begin = () => {
        console.log(this.api + ': begin')
        this.busy = true
      }
      const end = () => {
        console.log(api + ': end')
        this.busy = false
      }
      begin()
      return new Promise((resolve, reject) => {
        setTimeout(resolve, this.API_DELAY)
      })
        .then(end)
    },
    wait () {
      return new Promise((resolve, reject) => {
        setTimeout(resolve, this.WAIT)
      })
    },
    api1 () {
      return this.callApi('API1')
    },
    api2 () {
      return this.callApi('API2')
    },
    start () {
      console.log('start')
      this.api1()
        .then(this.wait)
        .then(this.api2)
    }
  }
}
</script>
```

Setting `API_DELAY` to 500 and `WAIT` to 100 will result in expected behavior. That is, modal opens and "API1: calling ..." is displayed, modal closes, modal opens and "API 2: calling ..." is displayed, and modal closes.

Setting `API_DELAY` to short period, and `WAIT` to very short period, will result the modal which displays "API 2: calling ..." remains open.

The reason is that when `hide()`, `show()`, `hide()` is executed very quickly, the second `hide()` is ignored.

First, the second `show()` begins before the first `hide()`'s transition processing is over, but since the `modal-open` class added in the first `show()` still remains, execution of `doShow()` is delayed until the `hidden` event comes. Therefore, `is_visible` will remain `false` at this point.

When the second `hide()` starts at this time, `is_visible` is false so `hide()` skip processing and do nothing.

Eventually the first `hide()`'s transition process is completed. Modal is closed, a `hidden` event occurs, the delayed `doShow()` is executed and modal opens. However, since the second `hide()` has already ended, modal will remain open.

I got the expected behavior with the following patch.

https://github.com/4dn-oss/bootstrap-vue/commit/6cb97774b6fe0760393cd797b5d6490de2b95755

4dn-oss:
This maybe fixes #1826.

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -248,6 +248,7 @@ export default {
       is_visible: false,
       is_transitioning: false,
       is_show: false,
+      is_show_delayed: false,
       is_block: false,
       scrollbarWidth: 0,
       isBodyOverflowing: false,
@@ -501,6 +502,7 @@ export default {
       }
       if (hasClass(document.body, 'modal-open')) {
         // If another modal is already open, wait for it to close
+        this.is_show_delayed = true
         this.$root.$once('bv::modal::hidden', this.doShow)
       } else {
         // Show the modal
@@ -509,8 +511,18 @@ export default {
     },
     hide (trigger) {
       if (!this.is_visible) {
-        return
+        if (this.is_show_delayed) {
+          // If another modal is started to hide but not hidden yet,
+          // and yet another modal is about to open, wait for it to
+          // open.
+          this.$root.$once('bv::modal::shown', this.doHide)
+        } else {
+          return
+        }
       }
+      this.doHide(trigger)
+    },
+    doHide (trigger) {
       const hideEvt = new BvEvent('hide', {
         cancelable: true,
         vueTarget: this,
@@ -534,7 +546,10 @@ export default {
       }
       this.emitEvent(hideEvt)
       // Hide if not canceled
-      if (hideEvt.defaultPrevented || !this.is_visible) {
+      if (hideEvt.defaultPrevented) {
+        return
+      }
+      if (!this.is_visible) {
         return
       }
       // stop observing for content changes
@@ -547,6 +562,7 @@ export default {
     },
     // Private method to finish showing modal
     doShow () {
+      this.is_show_delayed = false
       // Plce modal in DOM if lazy
       this.is_hidden = false
       this.$nextTick(() => {


src/components/modal/modal.js
```diff
@@ -509,8 +511,18 @@ export default {
     },
     hide (trigger) {
       if (!this.is_visible) {
-        return
+        if (this.is_show_delayed) {
+          // If another modal is started to hide but not hidden yet,
+          // and yet another modal is about to open, wait for it to
+          // open.
+          this.$root.$once('bv::modal::shown', this.doHide)
```
tmorehouse: This code causes the `trigger` information (if any present) to be lost.

src/components/modal/modal.js
```diff
@@ -534,7 +546,10 @@ export default {
       }
       this.emitEvent(hideEvt)
       // Hide if not canceled
-      if (hideEvt.defaultPrevented || !this.is_visible) {
+      if (hideEvt.defaultPrevented) {
+        return
+      }
+      if (!this.is_visible) {
         return
```
tmorehouse: Why break this test into two separate if statements?

src/components/modal/modal.js
4dn-oss: Thanks. I will change It as follows:
```suggestion
          this.$root.$once('bv::modal::shown', () => {
            this.doHide(trigger)
          })
```

src/components/modal/modal.js
4dn-oss: Sorry. I changed it so that it is convenient for debugging and I forgot to revert it. I will revert it.

PR #1857, opened 2018-05-28
Fix typos in Popover docs

ThibaultVlacich:
Here's some typos I found when reading this doc, there may be more, but this is what I found.


mosinve:
Thanks 👍

PR #1868, opened 2018-06-07
feat (table): Updated table component to support sorting on nested object properties.

bobzurad:
I noticed that you can't sort a table if the key is a nested property. This fixes that. See the example in the components/table/readme.md that the address.city column can now be sorted.

### DIFF
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -40,12 +40,29 @@ function recToString (obj) {
 }
 
 function defaultSortCompare (a, b, sortBy) {
-  if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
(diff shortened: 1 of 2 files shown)


src/components/table/table.js
```diff
@@ -40,14 +40,31 @@ function recToString (obj) {
...
 }
 
+function getNestedValue (obj, path) {
+  path = path.split('.')
+  var value = obj
+  for (var i = 0; i < path.length; i++) {
+    value = value[path[i]]
+  }
+  return value
+}
```
tmorehouse: Since lodash.get was already imported into `b-table` and used when computing the field values, PR #2149 eliminates the need for this custom function by switching over to use `_.get`

tmorehouse: Also, `_.get` handles properties that are not found (i.e. `undefined`) allowing for a fallback default value.

PR #1874, opened 2018-06-08
feat(table): Add row-unhovered event

gaaf:
Some users might want to know not only when the mouse is hovering a row,
but also when the mouse leaves the row.

For example for dynamically showing (and hiding) some row content based
on mouse hovering.

### DIFF
--- a/src/components/table/package.json
+++ b/src/components/table/package.json
@@ -60,6 +60,24 @@
                 ]
             },
             {
+                "event": "row-unhovered",
+                "description": "Emitted when a row is unhovered.",
+                "args": [
+                    {
+                        "arg": "item",
+                        "description": "Item data of the row being unhovered."
+                    },
+                    {
+                        "arg": "index",
+                        "description": "Index of the row being unhovered."
+                    },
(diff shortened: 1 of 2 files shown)


mosinve:
There's a `::hover` css pseudo-class, and b-table has a `tbody-tr-class` prop. Combining both, in most cases could help, i think

mosinve:
b-table is already over-customizable.

gaaf:
@mosinve Thanks for the hint. That indeed works for showing/hiding stuff in the same row (column with action buttons in my case). But this won't solve the general case.

This PR is to make the API more coherent as it is asymmetric now. There is an event for handling mouseenter (row-hovered), but there is no way to handle mouseleave (row-unhovered).

tmorehouse:
@gaaf could you add in a test for the unhovered event?

tmorehouse:
For testing, I am not sure how to test (i.e. simulate) a mouse over/out event JSDOM.  So I guess we skip that (as we currently do not test for the current `row-hovered` event).

Issue #225, opened 2017-04-12
Export table data functionality


coderabsolute:
Am requesting to add the functionality to the table component to export to pdf, excel, csv if possible....

pi0:
@coderabsolute Sorry but we can't extend table functionality anymore. It is meant to keep to base bootstrap. You may use special libraries for that like https://github.com/ratiw/vue-table

pi0:
Related to #224 :)

SirLamer:
Milenon, if I may suggest, export to spreadsheet is best handled server-side and so is external to the function of this client-side package. It can be a big job and you have to assume that your user is connecting on their free crappy cell phone.

For my project, one trick I use is that I maintain two separate table templates for each table - one meant for rendering, one meant for spreadsheet export. Though inefficient from a design perspective, I chose this way because there are often subtle differences in what you want to present in each case. I am using this Package in Laravel, using the Blade templating engine:
https://github.com/PHPOffice/PHPExcel

So, you can just shovel a HTML table in to this package and it will produce a spreadsheet, which makes templating easy but isn't the most efficient programatically. It also allows you to just feed it a PHP array or other techniques if you want to get fancy. For me, I wanted quick changes which look similar to the web app rendering, so I am using HTML tables to spreadsheet.

coderabsolute:
@pi0 Noted with thanks.
@SirLamer Thanks for the explanation.

PR #1893, opened 2018-06-15
feat(table): Split computedItems into multiple functions

gaaf:
To make it easier to extend the functionality of b-table without feature creep in the b-table itself, a bit more/easier access to the various stages of computedItems is needed.

For example, to implement #225 externally, access is needed to the filtered and sorted, but not yet paginated items. With this change this becomes possible by doing something like this:
    const table = this.$refs.the-table
    let items = table.hasProvider ? table.localItems : table.items
    items = table.sortItems(table.filterItems(items))

This PR incorporates no functional changes and adds no features.

### DIFF
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -787,7 +787,7 @@ export default {
       // Array copy for sorting, filtering, etc.
       items = items.slice()
       // Apply local filter
-      items = this.filterItems(items)
+      this.filteredItems = items = this.filterItems(items)
       // Apply local sort
       items = this.sortItems(items)
       // Apply local pagination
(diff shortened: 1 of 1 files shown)


tmorehouse:
This change actually broke efficient reactivity, since the computed prop getters were moved from a computed prop into non reactive methods.. slowing down rendering on larger tables.

External functions/addons should not be dipping into unpublished methods or computed props. Event interfaces or oneway-syncable props should be used instead.

PR #2149 fixes the broken reactivity by making sure the props/data that need to be reacted on are inside computed props (i.e to register that they are dependent upon them.  Moving to methods broke that).

PR #1914, opened 2018-06-28
If you're closing the popover, the action is "hide" not "show". Chang…

phil-w:
Teeny documentation typo fix. Not sure if this is how I make it.


phil-w:
I think that's what I need to do. Please shout if that's not the process.

PR #1929, opened 2018-07-08
feat(button): Make button tag configurable

jackmu95:
This PR makes the `b-button`s tag confirgurable (e.g. to a `div`).

This can be very handy if you want the buttons styling but want to place another clickable element inside the button which is normally prevented by the [interactive content model](https://www.w3.org/TR/html5/dom.html#interactive-content).


jackmu95:
@mosinve I've added the test.

PR #1947, opened 2018-07-17
feat(carousel): added carousel options

shinrox:
This PR enable some options on carousel component

## Focus
Whether to focus or not on carousels without an interval (auto play)

## Wrap
Defines if carousel cycles continuously or have hard stops

## Pause
Whether the carousel should stop on hover or not

### DIFF
--- a/src/components/carousel/README.md
+++ b/src/components/carousel/README.md
@@ -12,6 +12,7 @@ controls and indicators. Please be aware that nested carousels are **not** suppo
                 controls
                 indicators
                 focus
+                wrap
                 background="#ababab"
                 :interval="4000"
                 img-width="1024"
@@ -122,6 +123,9 @@ Both indicators and controls can be set at the same time or independently.
 The carousel will focus the next slide for screen readers if not in play mode (doesn't have an interval)
 
 To avoid this, set the prop `focus` to false
+
+# Wrap
+Set the prop `wrap` to set how the carousel should cycle. Continuously or with hard stops.
 
 
 ## V-model support
--- a/src/components/carousel/README.md
+++ b/src/components/carousel/README.md
@@ -13,6 +13,7 @@ controls and indicators. Please be aware that nested carousels are **not** suppo
                 indicators
                 focus
                 wrap
+                pause="hover"
                 background="#ababab"
                 :interval="4000"
                 img-width="1024"
@@ -52,14 +53,12 @@ controls and indicators. Please be aware that nested carousels are **not** suppo
           ut lacus vel interdum.
         </p>
       </b-carousel-slide>
-
     </b-carousel>
 
     <p class="mt-4">
       Slide #: {{ slide }}<br>
       Sliding: {{ sliding }}
     </p>
-
   </div>
 </template>
 
@@ -127,6 +126,8 @@ To avoid this, set the prop `focus` to false
 # Wrap
 Set the prop `wrap` to set how the carousel should cycle. Continuously or with hard stops.
 
+# Pause
+Set the prop `pause` to avoid scroll to pause on `hover`
 
 ## V-model support
 Programmaticaly control which slide is showing via `v-model` (which binds to the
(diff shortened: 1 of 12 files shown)


tmorehouse:
Could you change the new props to `noFocus`, `noWrap` and `noHoverPause`, and default them to `false`?

This would better follow other components (when disabling the defaults), and works better when using just the presence of the prop to disable the default.

shinrox:
Instead of changing what about adding those you've mentioned?

Except for the focus (cause it doesn't exists in bootstrap) the other ones are following the bootstrap default, should I really change them? Not following the framework that this plugin is built on is a bit confusing, isn't it?

https://getbootstrap.com/docs/4.1/components/carousel/#options

tmorehouse:
@shinrox Its more of a Vue thing.

Look at it this way when adding attributes... such as `wrap` which defaults to `true`:

```html
/* : required to convert string to boolean */
<b-carousel :wrap="false" ... >

 /* actually evaluates to a `truthy` value, so in fact wrap is true, which is default */
<b-carousel wrap="false" ... >

 /* ends up being the same as default (Boolean true) */
<b-carousel wrap ... >
```

Where as the `no-wrap` option is more intuitive to disable the default behaviour:
```html
/* not attribute value required to disable default */
<b-carousel no-wrap  ... >

 /* actually evaluates to a `truthy` value */
<b-carousel no-wrap="true" ... >

 /* evaluates to a `true` value because of v-bind: */
<b-carousel :no-wrap="true" ... >
```

It makes the attribute value processing much simpler (and the components smaller)

Internally, with no attribute/props
- focus functionality default is true (required for ARIA compliance)
- wrap functionality default is true
- pause functionality default is true (required for ARIA compliance)

Hence it makes more sense to use the `no-` form for the props (as is the norm with all other Bootstrap Vue components where `true` is the default)

shinrox:
Cool, thanks for the explanation, I'll try to update this PR by the end of this week.

src/components/carousel/README.md
```diff
@@ -11,6 +11,9 @@ controls and indicators. Please be aware that nested carousels are **not** suppo
                 style="text-shadow: 1px 1px 2px #333;"
                 controls
                 indicators
+                focus
+                wrap
```
tmorehouse: using just `wrap` is the same thing as not using `wrap` at all... this makes for confusing props and docs.

src/components/carousel/README.md
```diff
@@ -117,6 +118,16 @@ Set the prop `indicators` to show the slide indicator buttons.
 
 Both indicators and controls can be set at the same time or independently.
 
+# Focus
+The carousel will focus the next slide for screen readers if not in play mode (doesn't have an interval)
+
+To avoid this, set the prop `focus` to false
```
tmorehouse: this is not recommended as it breaks ARIA and accessibility. 

It actually focuses the _current_ slide not the next one that isn't shown.

src/components/carousel/README.md
```diff
@@ -11,6 +11,9 @@ controls and indicators. Please be aware that nested carousels are **not** suppo
                 style="text-shadow: 1px 1px 2px #333;"
                 controls
                 indicators
+                focus
+                wrap
+                pause="hover"
```
tmorehouse: why is this a string, when the only option is really  on or off (true or false)? And it is either on hover or not.

src/components/carousel/carousel.js
```diff
@@ -413,11 +433,11 @@ export default {
         setAttr(nextSlide, 'aria-hidden', 'false')
         currentSlide.tabIndex = -1
         nextSlide.tabIndex = -1
-        if (!this.isCycling) {
+        if (!this.isCycling && this.focus) {
           // Focus the next slide for screen readers if not in play mode
           nextSlide.tabIndex = 0
           this.$nextTick(() => {
-            nextSlide.focus()
```
tmorehouse: normally the focusing of the slide only happens when the slide is paused due to a mouseover.

tmorehouse:
Any movement on this?

tmorehouse:
Closing as stale

qkdreyer:
@shinrox Are you willing to finish this PR ?

PR #1994, opened 2018-08-09
feat(table): pass sortDesc to user provided sortCompare routine

brianmcburney:
Rationale: often, we'd prefer that undefined/null values are sorted last regardless of sort order. The user provided sortCompare routine needs to know the sort order in order to achieve this.


tmorehouse:
Could you also update the documentation to reflect the change?

tmorehouse:
Added documentation and added a few code tweaks

PR #1996, opened 2018-08-15
feat(dropdown): add support for contextmenu

lnufnu:
Added prop `isRightClick` to dropdown component to support`contextmenu` click event on both split dropdowns and default dropdown buttons.

### DIFF
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -279,10 +279,10 @@ export default {
         // We only toggle on Click, Enter, Space, and Arrow Down
         return
       }
-      if (this.isRightClick && (type == 'click' || type === 'keydown')) {
+      if (this.isRightClick && (type === 'click' || type === 'keydown')) {
         return
       }
-      if (!this.isRightClick && type == 'contextmenu') {
+      if (!this.isRightClick && type === 'contextmenu') {
         return
       }
       if (this.disabled) {


tmorehouse:
The prop name might be better if it was called `isContextMenu` (or just `isContext`), as some left handed users switch the mouse button order (i.e. context becomes the left button, select the right), and Apple computers use a click sequence to create a context click.

tmorehouse:
Another issue we must look into, is that this option no longer makes the menu accessible to keyboard only users (ARIA compliance).  When the trigger element is focused it is impossible to open the menu by pressing enter or space.  Since the trigger is a button, it should work like a button (when focused, pressing enter or space should open it)

Context menus are typically used for shortcuts to other options already on the page, and don't appear in the flow for screen readers or keyboard users.  Maybe, when this option is enabled, the entire dropdown component should have the attribute `aria-hidden="true"`, and all other ARIA attributes should be removed (`aria-haspopup` and `aria-expanded`).

The documentation for this option should be moved upwards and out of the accessibility section (since it is not accessible), and then a note in the accessibility section should mention best practices about using context menu, and having other available on screen options for the items in the context menu (i.e. having each menu option available elsewhere as a link or button, as context menu should be used as a convenience, and not the only way to access the items in the menu)

The way this works at the moment, I cannot approve this PR, as it breaks all ARIA accessibility.

It might be better made as a separate component (built upon the dropdown mixins), that has a named `slot` for the menu items, and a default slot for an item (such as a form) that the context menu applies to. i.e.

```html
<b-context-dropdown>
  <slot name="items">
    <b-dropdown-item>First Action</b-dropdown-item>
    <b-dropdown-item>Second Action</b-dropdown-item>
    <b-dropdown-item>Third Action</b-dropdown-item>
  </slot>
  <b-alert show>
    This is an item that has a context menu
  </b-alert>
</b-contextmenu>
```

And the dropdown would open at the current pointer position on right click

src/mixins/dropdown.js
```diff
@@ -275,10 +275,16 @@ export default {
       evt = evt || {}
       const type = evt.type
       const key = evt.keyCode
-      if (type !== 'click' && !(type === 'keydown' && (key === KeyCodes.ENTER || key === KeyCodes.SPACE || key === KeyCodes.DOWN))) {
+      if (type !== 'click' && type !== 'contextmenu' && !(type === 'keydown' && (key === KeyCodes.ENTER || key === KeyCodes.SPACE || key === KeyCodes.DOWN))) {
         // We only toggle on Click, Enter, Space, and Arrow Down
         return
       }
+      if (this.isRightClick && (type === 'click' || type === 'keydown')) {
```
tmorehouse: This makes the menu completely inaccessible to keyboard only users

PR #2008, opened 2018-08-30
fix(form-file): issue with "accept" values (fixes #1526, reverts #2028)

### LINKED ISSUE #1526: b-form-file issue with "accept" values - in "multiple" mode
[Link to JSFiddle](https://jsfiddle.net/5Lcspc1m/1/)

I found what to me looks like a strange behaviour of `b-form-file` component. I set an instance of `b-form-file` to support selecting multiple files, and specified `accept` MIME types, exactly like it's shown in the docs (just skipped the `image/gif` type):

```
<b-file v-model="files" multiple accept="image/jpeg, image/png"></b-file>
```

But when dropping any files (single or multiple) on the input doesn't update the model - the component simply ignores all the files, even valid ones. When picking files manually, instead of dropping them, the non-accepted files are disabled in the browser file select dialog (in Chrome), so that's OK. But then again - choosing any JPG or PNG file doesn't update the model, and the event emitted by the component has an empty payload.

`b-form-file` still behaves as expected when not in "multiple" mode.

OS: macOS High Sierra
Browsers: Chrome 63, Firefox 57
Bootstrap: 4.beta-2
BootstrapVue: 1.4.0

impowski:
Fixes #1526 

Reverts #2028

### DIFF
--- a/src/components/form-file/form-file.js
+++ b/src/components/form-file/form-file.js
@@ -183,18 +183,18 @@ export default {
     setFiles (files) {
       if (!files) {
         this.selectedFile = null
-        return
-      }
-      if (!this.multiple) {
+      } else if (this.multiple) {
+        // Convert files to array
+        const filesArray = []
+        for (let i = 0; i < files.length; i++) {
+          filesArray.push(files[i])
+        }
+        // Return file(s) as array
+        this.selectedFile = filesArray
+      } else {
+        // Return single file object
         this.selectedFile = files[0]
-        return
       }
-      // Convert files to array
-      const filesArray = []
-      for (let i = 0; i < files.length; i++) {
-        filesArray.push(files[i])
-      }
-      this.selectedFile = filesArray
     },
     dragover (evt) {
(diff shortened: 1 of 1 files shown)


toadkicker:
I propose we use the mime-types package and simplify this a lot for developers. It would be really nice to just say `accepts="images"`, `accepts="video"`, etc. and not worry about the specifics of all the mime types that means. We would also prevent a possible content spoofing attack because it can trust that the JS API is comparing the same symbols.

impowski:
@toadkicker there is already `images/*` simple enough I guess. I see what you saying about spoofing I'll see how it can be done.

tmorehouse:
Basically, we should fall back to letting the browser handle the accept formats.

Older browsers had issues with which format to use (i.e. extension vs IANA media types).

Until we make a matching algorithm to test the file type (parsing the accept string into RegEx's and then testing against the `file.type` for IANA or `file.name` for extensions), we should just let the browser do hte validation

Issue #2018, opened 2018-09-06
Document why v-b-modal directive is recommended


ClemMakesApps:
I'm currently working on integrating bootstrap vue's modal into my application and wanted to learn more about why `v-b-modal` directive is the recommended approach. If I can get some insight here, I'd be happy to update the documentation with that information.

tmorehouse:
One of the main reasons is that the `v-b-modal` directive automatically adds/updates the `aria-*` attributes (for accessibility) to reflect that the trigger controls a "popup/dialog"  and what state the modal is in (on the trigger element).

ClemMakesApps:
Sorry, I'm still having trouble following. Can you show me an example?

tmorehouse:
You can see the `aria-*` attributes that re added/controled on the trigger element that has the `v-b-modal` directive applied, when you inspect the element.

ClemMakesApps:
That doesn't seem to be the case in the documentation website. Unless I am missing something?

![2018-09-06 10 49 00](https://user-images.githubusercontent.com/3363898/45169133-8ea8d080-b1c2-11e8-8bc1-b0d8f4f31837.gif)

tmorehouse:
Ah, right, not aria-attributes, but it does add the appropriate `role` (button) to non button triggers (i.e. link or DIV), and also automatically returns focus to the trigger element when the modal closes.

Notice when the modal closes that the trigger button has regained focus.  When not using the v-b-modal directive, you have to specify which element to return focus to (for accessibility reasons)

tmorehouse:
See https://bootstrap-vue.js.org/docs/components/modal/#auto-return-focus in the section https://bootstrap-vue.js.org/docs/components/modal/#returning-focus-to-the-triggering-element

PR #2020, opened 2018-09-07
docs: clarify the benefits of using v-b-modal

ClemMakesApps:
I changed the wording to make it more explicit as to why `v-b-modal` is recommended over other approaches. I hope this change makes it clearer to understand.

https://github.com/bootstrap-vue/bootstrap-vue/issues/2018

### DIFF
--- a/src/components/modal/README.md
+++ b/src/components/modal/README.md
@@ -41,7 +41,7 @@ not be present, nor can you use the `modal-title` slot.
 
 There are several methods that you can employ to toggle the visibility of `<b-modal>`.
 
-### Using `v-b-modal` directive (recommended)
+### Using `v-b-modal` directive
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for this PR. Your changes seem good but I think we should not add the word `recommended`. For example on complex scenarios, `v-model` is better.

ClemMakesApps:
@pi0 line 44 mentions that it is recommended. Perhaps that wording should change as well?

pi0:
Nice catch @ClemMakesApps. I think it would be better to remove that too.

ClemMakesApps:
@pi0 thanks, I've made the appropriate changes. Please take a look :)

tmorehouse:
Looks good

PR #2024, opened 2018-09-13
popover: watch content changing

chiaweilee:
to fix issue like this: 

```vue
<b-popover :content="$t('Hello I18n')"/>
```

or

```vue
<b-popover :content="someComputed"/>
```

### DIFF
--- a/src/components/popover/popover.js
+++ b/src/components/popover/popover.js
@@ -39,6 +39,14 @@ export default {
       default: 'right'
     }
   },
+  watch: {
+    content: {
+      handler: function (to) {
+        // watch content changing, +v
+        this._toolpop.$config.content = to
+      }
+    }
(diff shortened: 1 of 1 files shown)


tmorehouse:
Note that this will not dynamically update the content, as the config is evaluated when the popover in opened.

The popover gets destroyed when it is closed, and re-built when it is opened.

chiaweilee:
> Note that this will not dynamically update the content, as the popover config is evaluated when the popover in opened (not after it is opened)
> 
> The popover gets destroyed when it is closed, and re-built when it is opened, at which point the `content` prop is re-evaluated.
> 
> If you want dynamic content, when using the component version of popover, use the default slot to place dynamic content.

thanks for your help, it's helpful.

PR #2050, opened 2018-09-27
fix(modal): clear modal paddingLeft and paddingRight if no Scrollbar(s) in adjustDialog().

huadong:
modal.style.paddingLeft and modal.style.paddingRight should be set to 0, if no scrollbar exist.

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -739,9 +739,13 @@ export default {
         modal.scrollHeight > document.documentElement.clientHeight
       if (!this.isBodyOverflowing && isModalOverflowing) {
         modal.style.paddingLeft = `${this.scrollbarWidth}px`
+      } else {
+        modal.style.paddingLeft = ''
       }
       if (this.isBodyOverflowing && !isModalOverflowing) {
         modal.style.paddingRight = `${this.scrollbarWidth}px`
+      } else {
+        modal.style.paddingRight = ''
       }
     },
     resetAdjustments () {
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -608,7 +608,7 @@ export default {
     },
     onAfterLeave () {
       this.is_block = false
-      this.resetAdjustments()
+      this.resetDialogAdjustments()
       this.resetScrollbar()
       this.is_transitioning = false
       removeClass(document.body, 'modal-open')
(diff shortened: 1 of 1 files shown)


tmorehouse:
Have you run into an issue where this is needed?

huadong:
When the modal is opened, if something which is under the modal is hidden or resized, it will cause the scrollbar be changed, but the modal's position won't be adjusted.

tmorehouse:
This would ignore the adjustments made to the other (possible) elements on the page (such as fixed or sticky navbars).

huadong:
Thank you. I will try change the code: 
```js
      if (!this.isBodyOverflowing && isModalOverflowing) {
        modal.style.paddingLeft = `${this.scrollbarWidth}px`
      } else {
        modal.style.paddingLeft = ''
      }
      if (this.isBodyOverflowing && !isModalOverflowing) {
        modal.style.paddingRight = `${this.scrollbarWidth}px`
      } else {
        modal.style.paddingRight = ''
      }
```

to

```js
      this.checkScrollbar()
      if (!this.isBodyOverflowing && isModalOverflowing) {
        modal.style.paddingLeft = `${this.scrollbarWidth}px`
      }
      if (this.isBodyOverflowing && !isModalOverflowing) {
        modal.style.paddingRight = `${this.scrollbarWidth}px`
      }
```

tmorehouse:
We may want to compare what native bootstrap V4 is now doing.

tmorehouse:
Note `this.scrollbarWidth` will always have a value equal to the browsers native scrollbar width (even if one is not showing)

huadong:
May I reset the adjustments?
reset the adjustments and check scrollbar again.

```js
      this.resetAdjustments()
      this.checkScrollbar()
      if (!this.isBodyOverflowing && isModalOverflowing) {
        modal.style.paddingLeft = `${this.scrollbarWidth}px`
      }
      if (this.isBodyOverflowing && !isModalOverflowing) {
        modal.style.paddingRight = `${this.scrollbarWidth}px`
      }
```
reproduction:
1、open a modal whose height is overflowing.
2、adjust the window's height, until it's height is bigger than modal's. 

the issue snapshot:


image

huadong:
I add a function onResize() instead of modifying adjustDialog().
It works fine for me, but I am not sure what will happen if there are more than one modal at the same time.

BTW: `this.scrollbarWidth` is detected only once, chrome and firefox will change scrollbar's width when switch to `Responsive Design Mode`. But it's not an issue for production environment.

tmorehouse:
`b-modal` only supports a single modal open at a time, which is the same as native Bootstrap V4 modals).

PR #2064, opened 2018-10-07
feat(table): Support contextmenu event binding for table rows

peterjosling:
Allows for right click as well as existing click + double click.

### DIFF
--- a/src/components/table/package.json
+++ b/src/components/table/package.json
@@ -42,6 +42,24 @@
                 ]
             },
             {
+                "event": "row-contextmenu",
+                "description": "Emitted when a row is right clicked.",
+                "args": [
+                    {
+                        "arg": "item",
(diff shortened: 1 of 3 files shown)


Joebayld:
Would love to see this get merged!

tmorehouse:
I think the event name should be `row-contextclicked` rather than `row-rightclicked`, as some left handed users have configured their computer to swap the mouse button positions (i.e. select on right, context on left), as well as some computers only have a single button (i.e. apple) and use a click sequence to indicate context.

peterjosling:
Good thinking. I've renamed it to `row-contextmenu` instead of `row-contextclicked`, as that matches the native browser event.

TIL: the browser event is `contextmenu` instead of right click because it also fires for the context key.

PR #2091, opened 2018-10-22
Show v-model usage for b-alert

PatBriPerso:


tmorehouse:
Rather than changing an existing example (which shows how to use the `show` attribute), it might be better to create a _new_ example using the `v-model` method.

PatBriPerso:
Thanks @tmorehouse. 
Sorry to not having time to do it my-self.

PR #2102, opened 2018-10-25
fix(collapse/toggle): "collapsed" class cleared when component updated. fixes #1798

### LINKED ISSUE #1798: Collapse: collapsed class not persistent
See this [fiddle](https://jsfiddle.net/sLgj23zu/1/)

the v-b-toggle directive sets the "collapsed" class of the button, that works well, click on the button and its title will change accordingly.

but... close the collapse, the title then correctly read "Open Collapse", then change the value in the input field... The "collapsed" class disappears, and the button now read "Close Collapse".

lianee:
fixes #1798

### DIFF
--- a/src/components/collapse/collapse.js
+++ b/src/components/collapse/collapse.js
@@ -187,6 +187,9 @@ export default {
     }
     this.emitState()
   },
+  updated() {
+    this.$root.$emit(EVENT_STATE, this.id, this.show)
+  },
   beforeDestroy () {
     if (this.isNav && typeof document !== 'undefined') {
       window.removeEventListener('resize', this.handleResize, false)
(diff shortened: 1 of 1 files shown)


src/components/collapse/collapse.js
```diff
@@ -187,6 +187,9 @@ export default {
     }
     this.emitState()
   },
+  updated () {
```
tmorehouse: This actually runs when the component's vdom has updated, and doesn't monitor for changes outside of the component.

What might be a better approach to this, is to update the `v-b-toggle` directive to use the `observeDom` utility to monitor changes to the trigger element's attributes (shallow observer).

I will look into this.

tmorehouse: Although just testing in the playground, this appears to work for most cases.

PR #2116, opened 2018-10-29
fix(tooltip): Add directive to component plugin

jackmu95:
This commit adds the missing tooltip directive to the plugin.


tmorehouse:
take a look here for how directives are registered:

https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/directives/tooltip/index.js

PR #2174, opened 2018-11-14
fix(fom-input): revert changes from PR #1841

tmorehouse:
reverts changes from #1841, as it breaks the readonly plaintext requirements (plus documentatijn was not properly updated to reflect changes)

### DIFF
--- a/src/components/form-input/form-input.js
+++ b/src/components/form-input/form-input.js
@@ -48,7 +48,7 @@ export default {
         type: self.localType,
         disabled: self.disabled,
         required: self.required,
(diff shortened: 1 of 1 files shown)


syffs:
I understand. Now this was a very convenient for rendering some custom styling (material) like this:
dynamic-label

How would you implement it now?

tmorehouse:
One can still use utility classes such as `class="border-0" to remove the border off of `<b-input>`, or just use a regular `<input>` with one's own custom classes.`

We may introduce a `plain` prop (similar to file, checkboxes, and radios), that renders without any styles.

PR #2251, opened 2018-12-07
fix(dropdown): Item click event

jackmu95:
## Description of Pull Request:
#2140 added the `onFocusOut` handler to the `focusout` event of the `.dropdown-menu`. This caused an issue on some browsers (e.g. Safari) that the `click` event on the `.dropdown-item` wasn't fired anymore.

This PR adds a minimal timeout before setting `visible` to `false` which allows all browsers to handle the `click` event before hiding the menu.

## PR checklist:

**What kind of change does this PR introduce?** (check at least one)
- [x] Bugfix
- [ ] Feature
- [ ] Enhancement to an existing feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)
- [ ] Yes
- [x] No

If yes, please describe the impact:

**The PR fulfills these requirements:**
- [x] It's submitted to the `dev` branch, **not** the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (i.e. `fixes #xxxx[,#xxxx]`, where "xxxx" is the issue number)
- [x] The PR should address only one issue or feature. If adding multiple features or fixing a bug and adding a new feature, break them into separate PRs if at all possible.
- [x] PR titles should following the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention (i.e. "fix(alert): not alerting during SSR render", "docs(badge): Updated pill examples, fix typos", "chore: fix typo in docs", etc). **This is very important, as the `CHANGELOG` is generated from these messages.**

**If new features/enhancement/fixes are added or changed:**
- [ ] Includes documentation updates (including updating the component's `package.json` for slot and event changes)
- [ ] New/updated tests are included and passing (if required)
- [x] Existing test suites are passing
- [x] The changes have not impacted the functionality of other components or directives 
- [ ] ARIA Accessibility has been taken into consideration (does it affect screen reader users or keyboard only users? clickable items should be in the tab index, etc)

**If adding a new feature, or changing the functionality of an existing feature, the PR's description above includes:**
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

PR #2252, opened 2018-12-07
fix(dropdown): Menu focusout close handling

jackmu95:
This is a follow-up PR to #2251 and addresses the real issue for `.dropdown-menu`'s `focusout` handling.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] When resolving a specific issue, it's referenced in the PR's title (i.e. `fixes #xxxx[,#xxxx]`, where "xxxx" is the issue number)

### DIFF
--- a/src/components/dropdown/dropdown.spec.js
+++ b/src/components/dropdown/dropdown.spec.js
@@ -96,20 +96,6 @@ describe('dropdown', async () => {
     expect(Array.from(dd_6.$refs.menu.children).find(node => node.innerHTML === 'button')).toBeElement('button')
   })
 
-  it('dd-item-button should emit click event', async () => {
-    const { app: { $refs } } = window
-    const { dd_6 } = $refs // eslint-disable-line camelcase
-
-    const spy = jest.fn()
-
-    dd_6.$parent.$root.$on('clicked::link', spy)
-
-    const buttonItem = Array.from(dd_6.$refs.menu.children).find(node => node.innerHTML === 'button')
-    buttonItem.click()
-
-    expect(spy).toHaveBeenCalled()
-  })
-
   it('dd-divider should render', async () => {
     const { app: { $refs } } = window
     const { dd_6 } = $refs // eslint-disable-line camelcase
--- a/src/components/dropdown/dropdown.spec.js
+++ b/src/components/dropdown/dropdown.spec.js
@@ -96,6 +96,20 @@ describe('dropdown', async () => {
     expect(Array.from(dd_6.$refs.menu.children).find(node => node.innerHTML === 'button')).toBeElement('button')
   })
 
+  it('dd-item-button should emit click event', async () => {
+    const { app: { $refs } } = window
+    const { dd_6 } = $refs // eslint-disable-line camelcase
+
+    const spy = jest.fn()
+
+    dd_6.$parent.$root.$on('clicked::link', spy)
+
+    const buttonItem = Array.from(dd_6.$refs.menu.children).find(node => node.innerHTML === 'button')
+    buttonItem.click()
+
+    expect(spy).toHaveBeenCalled()
+  })
+
   it('dd-divider should render', async () => {
     const { app: { $refs } } = window
     const { dd_6 } = $refs // eslint-disable-line camelcase
--- a/src/components/link/link.js
+++ b/src/components/link/link.js
@@ -155,14 +155,13 @@ function clickHandlerFactory ({ disabled, tag, href, suppliedHandler, parent }) 
       // Kill the event loop attached to this specific EventTarget.
       e.stopImmediatePropagation()
     } else {
-      parent.$root.$emit('clicked::link', e)
-
       if (isRouterLink && e.target.__vue__) {
         e.target.__vue__.$emit('click', e)
       }
       if (typeof suppliedHandler === 'function') {
         suppliedHandler(...arguments)
       }
+      parent.$root.$emit('clicked::link', e)
     }
 
     if ((!isRouterLink && href === '#') || disabled) {
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -347,12 +347,10 @@ export default {
       // Note: Inputs are in a special .dropdown-form container
     },
     onFocusOut (evt) {
-      if (this.$el.contains(evt.relatedTarget)) {
+      if (this.$el.contains(evt.target)) {
         return
       }
-      setTimeout(() => {
-        this.visible = false
-      }, 150)
+      this.visible = false
     },
     onMouseOver (evt) /* istanbul ignore next: not easy to test */ {
       // Removed mouseover focus handler
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -347,10 +347,12 @@ export default {
       // Note: Inputs are in a special .dropdown-form container
     },
     onFocusOut (evt) {
-      if (this.$el.contains(evt.target)) {
+      if (this.$el.contains(evt.relatedTarget)) {
         return
       }
-      this.visible = false
+      setTimeout(() => {
+        this.visible = false
+      }, 150)
     },
     onMouseOver (evt) /* istanbul ignore next: not easy to test */ {
       // Removed mouseover focus handler
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -347,12 +347,10 @@ export default {
       // Note: Inputs are in a special .dropdown-form container
     },
     onFocusOut (evt) {
-      if (this.$el.contains(evt.relatedTarget)) {
+      if (this.$el.contains(evt.target)) {
         return
       }
-      setTimeout(() => {
-        this.visible = false
-      })
+      this.visible = false
     },
     onMouseOver (evt) /* istanbul ignore next: not easy to test */ {
       // Removed mouseover focus handler
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -4,7 +4,7 @@ import { assign } from '../utils/object'
 import KeyCodes from '../utils/key-codes'
 import BvEvent from '../utils/bv-event.class'
 import warn from '../utils/warn'
-import { closest, contains, eventOff, eventOn, getAttr, isVisible, selectAll } from '../utils/dom'
+import { isVisible, closest, selectAll, getAttr, eventOn, eventOff } from '../utils/dom'
 
 // Return an Array of visible items
 function filterVisible (els) {
@@ -238,13 +238,10 @@ export default {
         this.$root.$on('clicked::link', this.rootCloseListener)
         // Use new namespaced events for clicked
         this.$root.$on('bv::link::clicked', this.rootCloseListener)
-        // Hide the dropdown when it loses focus
-        eventOn(document, 'focusin', this.focusHandler, false)
       } else {
         this.$root.$off('bv::dropdown::shown', this.rootCloseListener)
         this.$root.$off('clicked::link', this.rootCloseListener)
         this.$root.$off('bv::link::clicked', this.rootCloseListener)
-        eventOff(document, 'focusin', this.focusHandler, false)
       }
       // touchstart handling fix
       /* istanbul ignore next: not easy to test */
@@ -349,21 +346,14 @@ export default {
       // Tab, if in a text-like input, we should just focus next item in the dropdown
       // Note: Inputs are in a special .dropdown-form container
     },
+    onFocusOut (evt) {
+      if (this.$el.contains(evt.target)) {
+        return
+      }
+      this.visible = false
+    },
     onMouseOver (evt) /* istanbul ignore next: not easy to test */ {
       // Removed mouseover focus handler
-    },
-    // Document focusin listener
-    focusHandler (evt) {
-      // If focus leaves dropdown, hide it
-      const menu = this.$refs.menu
-      if (
-        this.visible &&
-        menu &&
-        document !== evt.target &&
-        !contains(menu, evt.target)
-      ) {
-        this.visible = false
-      }
     },
     focusNext (evt, up) {
       if (!this.visible) {
--- a/src/mixins/dropdown.js
+++ b/src/mixins/dropdown.js
@@ -4,7 +4,7 @@ import { assign } from '../utils/object'
 import KeyCodes from '../utils/key-codes'
 import BvEvent from '../utils/bv-event.class'
 import warn from '../utils/warn'
-import { isVisible, closest, selectAll, getAttr, eventOn, eventOff } from '../utils/dom'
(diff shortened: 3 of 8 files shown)


src/mixins/dropdown.js
```diff
@@ -347,12 +347,10 @@ export default {
       // Note: Inputs are in a special .dropdown-form container
     },
     onFocusOut (evt) {
-      if (this.$el.contains(evt.relatedTarget)) {
+      if (this.$el.contains(evt.target)) {
```
tmorehouse: `evt.relatedTarget` is the correct value to check, as it refers to the element that is _gaining_ focus, not the element that is _loosing_ focus (for a `focusout` event).

We want to see if the element that just received the focus outside of the dropdown menu (as we are listening at the root of the dropdown), and using `evt.target` makes it so that the dropdown never closes on clickout or tab-out.

Perhaps, it might be better for us to instantiate a `focusout` listener, whenever the dropdown is open, on the document element, and then we could test for `evt.target` being inside the dropdown.

jackmu95: @tmorehouse Thanks for pointing out. Do you have a concrete solution for this in mind or should I try to figure it out over the weekend?

Nevertheless, my previous solution isn't working reliable and we need to come up with a better solution.

tmorehouse: You can take a peek at the latest `b-modal` code's `enforceFocus` handler, which is now adding/removing a listener on the `document` whenever the modal is open.

You could basically enable the listener when the dropdown is `shown`, and remove it on `hidden` (or un-cancelled `hide`, and also on `beforeDestroy()`).  It would be a few extra lines of code, but I think it may address the issue.

jackmu95: Thanks for the hint. Will look into it over the weekend 👍

tmorehouse: You could add the `focusout` (or possibly `focusin` instead) registration via `addEvent()`/`removeEvent(...)` to the `whileOpenListen` method (https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/mixins/dropdown.js#L232) method in the dropdown mixin, and remove it off of the dropdown container.

tmorehouse: Hmmm.. Safari Mobile or desktop?

jackmu95: Desktop

tmorehouse: OK.. I know mobile has issues with certain things

tmorehouse:
So just to make sure I'm on the same page,  this issue with adding a `@click` handler on a dropdown item, and it is not being called because the menu is being closed before it gets a chance to run?

tmorehouse:
Hmmm...  dropdown items (and buttons) use `b-link` as their base, which emits the `bv::link::clicked` $root event (which dropdown listens to in order to self-close).

I wonder if this event is being fired before the regular click handler gets called.

Or, if the root listener for that event in dropdown (and its old event) should have the delay added to it instead?

tmorehouse:
Now I'm thinking the `rootCloseListener` method maybe should have the delay added to it...

```js
     rootCloseListener (vm) {
       if (!this.visible) {
         return
       }
       if (vm !== this) {
         setTimeout(() => {
           this.visible = false
         }, 150)
       }
     },
```

tmorehouse:
Or possibly, it might be an issue within b-link in this section:

https://github.com/jackmu95/bootstrap-vue/blob/fix-dropdown-menu-focusout-handling/src/components/link/link.js#L157-L166

```js
    } else {
      parent.$root.$emit('clicked::link', e)

      if (isRouterLink && e.target.__vue__) {
        e.target.__vue__.$emit('click', e)
      }
      if (typeof suppliedHandler === 'function') {
        suppliedHandler(...arguments)
      }
    }
```

Where the root event is emitted before any user events (suppliedHanlder).  So maybe this might fix the issue:

```js
    } else {
      if (isRouterLink && e.target.__vue__) {
        e.target.__vue__.$emit('click', e)
      }
      if (typeof suppliedHandler === 'function') {
        suppliedHandler(...arguments)
      }
      parent.$root.$emit('clicked::link', e)
    }
```

And we _may_ no longer would need to use a `setTimout` delay

jackmu95:
I have no idea why this was closed.

The issue isn't related to `b-link`.

tmorehouse:
I was just thinking it might be the sequence of events fired by `b-link` as it is the root element of `b-dropdown-item`, and dropdown relies on the `clicked::link` event (that `b-link` fires) to trigger the menu close.

jackmu95:
Here is what happen in the different Browsers:

**Chrome**
- Emits the `dropdown-item` `click` event
- Emits the `dropdown-menu` `focusout` event

**Firefox**
- Emits the `dropdown-item` `click` event
- Don't emits the `dropdown-menu` `focusout` event, only after clicking somewhere else

**Safari**
- Don't emits the `dropdown-item` `click` event
- Emits the `dropdown-menu` `focusout` event

tmorehouse:
Interesting...

Could you check for `$root.$on('clicked::link', handler)` in your test sequences?  Curious to see the event emit order.  In the case of clicking an item inside of the dropdown, it is the `$root` event that triggers the close, not  `focusout` (which is mainly used for clickout and tab-out closing of the menu)

tmorehouse:
So that is weird... Safari doesn't emit the `$root.$emit('clicked::link')` event at all... which it should be.

I've been googling a bit, and noticed (an older bug) related to Safari where if a non-interactive element won't emit any kind of click events, unless the element has the style `pointer: cursor;` set on the element.

`b-link` is renders an interactive element (i.e. `a` tag), but maybe safari detects the event `preventDefaut()` (when the href="#", to prevent the page from scrolling) and infers that the element is non-interactive.

Curious if you change `d-dropdown-item` to `b-dropdown-item-button` if the issue persists?

jackmu95:
I think the problem is that the `focusout` event of the `dropdown-menu` is called before the `click` event of `d-dropdown-item` or `b-dropdown-item-button`. This sets the menu to `display: none;` which prevents the `click` event from being fired at all.

tmorehouse:
But in the case of the `focusout` event, the `event.relatedTarget` is the dropdown-item, which shouldn't trigger the menu close (as it is contained within the menu container).

jackmu95:
No, in all the browsers is `event.target` the `dropdown-item` - `event.relatedTarget` is always null.

tmorehouse:
Here is something you could test quickly (regarding the older safari bug/feature):

On the `b-dropdown-item` add the following attribute:  `style="cursor: pointer;"`  and see if it changes the behaviour.

tmorehouse:
But it is the `clicked::link` event that **should** be triggering the menu to close when you click on an item **within** the menu.... not the `focusout` event (as that dropdown-item element is _inside_ the menu and is not moving focus outside of it due to the click, not until the menu is closed and the browser moves the focus to somewhere else in the document, then focus out will trigger).

And your test above shows that `clicked:link` is not being fired at all in Safari, so it has nothing to do with the `focusout` event, but the fact that the menu item is not emitting the `clicked::link` event

jackmu95:
`clicked:link` is working in Safari when the element isn't hidden, which is the case because the `focusout` event is emitted before.

tmorehouse:
Hmmmm...  I'm now thinking a combination of things that may address this...

Switch to the document listener for `focusin` (and checking if evt.target is not contained in the menu container, which should only fire for clickout and focus moving _out_ of the menu), and in `b-link` moving the `$root.$emit('clicked::link')` to after the user handlers, to ensure any user supplied hanlder(s) run before the event is fired off (via $root) to trigger the close of the menu (or leave it where it is, but wrap the emit in a `$nextTick`)

src/mixins/focus-in.js
```diff
@@ -0,0 +1,38 @@
+import { contains, eventOff, eventOn } from '../utils/dom'
+
+export default {
+  data () {
+    return {
+      focusInElement: null,
```
tmorehouse: The only thing about storing an element reference in data is that Vue installs observers on most of it's properties (similar to objects), although we are doing it in a few other components as we need to see if the element is changed.

src/components/dropdown/dropdown.spec.js
```diff
@@ -96,20 +96,6 @@ describe('dropdown', async () => {
     expect(Array.from(dd_6.$refs.menu.children).find(node => node.innerHTML === 'button')).toBeElement('button')
   })
 
-  it('dd-item-button should emit click event', async () => {
```
tmorehouse: Actually the dropdown-item-button should emit a `clicked::link' event (which is fired after any user`click` handlers are run, as the event is merged in last after any user supplied events), just like `b-link` fires the `clicked::link` event to  lose the dropdown.

src/components/link/link.js
```diff
@@ -155,14 +155,13 @@ function clickHandlerFactory ({ disabled, tag, href, suppliedHandler, parent })
...
     } else {
-      parent.$root.$emit('clicked::link', e)
-
       if (isRouterLink && e.target.__vue__) {
         e.target.__vue__.$emit('click', e)
       }
       if (typeof suppliedHandler === 'function') {
         suppliedHandler(...arguments)
       }
+      parent.$root.$emit('clicked::link', e)
```
tmorehouse: :+1:

src/mixins/dropdown.js
```diff
@@ -346,17 +321,31 @@ export default {
...
-      }
-      setTimeout(() => {
-        this.visible = false
-      })
-    },
     onMouseOver (evt) /* istanbul ignore next: not easy to test */ {
       // Removed mouseover focus handler
     },
+    isClickOut () {
+      return true
```
tmorehouse: I think this is why clicking on non-interactive elements in a menu closes the menu.

Shouldn't this be a check to see if the event.target is not contained in menu container (and return true if it is, otherwise return false)?  As the purpose of clickout is to determine if something _outside_ of the menu was clicked (and close it)

tmorehouse: Although, technically, clicking outside of the menu _should_ move focus from the menu item to the clicked element (or the documentElement if not clicking on an interactive element).  So the focus handler _should_ (I would assume) close the menu (unless some browsers don't move focus to the document, possibly Mobile).

jackmu95: What I did here was to use the "click out" listener to replace the "close on dropdown item click" logic. But this caused some issues you have correctly pointed out.

> Although, technically, clicking outside of the menu should move focus from the menu item to the clicked element (or the documentElement if not clicking on an interactive element). So the focus handler should (I would assume) close the menu (unless some browsers don't move focus to the document, possibly Mobile).
Just clicking out of the menu doesn't fire the `focusin` handler when the next element isn't interactive. This is why I added back the "click out" listener.

With the combination of the current listener setup it seems to work very well.

src/mixins/focus-in.js
jackmu95: What about using a ref name instead and resolve `document` when no ref name is defined?

This isn't particularly needed for the current use case but makes this mixin much more flexible for a usage in other parts of the project.

tmorehouse: That might be an option.. or make it a non reactive private property.. unless it needs to be reactive 5o changes in the element ref.

jackmu95:
@tmorehouse I'am happy with this now. Can you please review this again?

src/mixins/dropdown.js
tmorehouse: That must be a safari issue with respect to not firing a focusout/in on clicking the document (I kinda recall reading about that issue somewhere on the web... maybe quirksmode, but I thought it might be safari mobile only)

src/mixins/dropdown.js
```diff
@@ -84,7 +86,9 @@ export default {
     return {
       visible: false,
       inNavbar: null,
-      visibleChangePrevented: false
+      visibleChangePrevented: false,
+      listenForClickOut: false,
+      listenForFocusIn: false
```
tmorehouse: Since these last two data items are defined in the mixins, you could leave them out (or comment them out and leave in noting they are in the mixins). When the mixins are imported and added, the data will be merged with the component data.

src/mixins/focus-in.js
tmorehouse: To make it non reactive, just don't define it in the data.

src/mixins/dropdown.js
jackmu95: They are set to true in the mixin by default and I deactivate them here because we want them only to be active when the dropdown is open.

tmorehouse: Ah, true...

jackmu95:
You are right, will look into that in a moment.

src/mixins/dropdown.js
tmorehouse: Him I wonder if the mixing should default them to false, and then any component that uses the mixins could set the values on created or mounted hook?

tmorehouse:
Looks like an older issue (drop left/eight). Seeing console errors in chrome desktop.

tmorehouse:
Looks like a reference error to the element when popper instance is created (either element, or refs.menu).

I think it is line 172 of dropdown.js where it tests for split or drop left/right.

tmorehouse:
One other oddball thing...  when clicking on an item in the menu, and the menu closes, it doesn't return focus to the dropdown toggle (although pressing `esc` to close the menu _does_ return focus to the trigger element).  Not sure if this is an older issue or not.

When checking to see what the active element is when the menu closes shows that the body has focus, when clicking on a dropdown item.  Which is not good for keyboard only users (they have to tab all the way back from the top of the document to where ever they were before).

jackmu95:
Both issues should be fixed now.

The `dropleft` and `dropright` reference errors where caused by missing placement options in the `AttachmentMap` and the `onEsc()` listener called the `focusToggler()` method after hiding the menu. The `rootCloseListener()` calls this method too now.

jackmu95:
I have also applied your suggestions to the mixins :)

src/mixins/dropdown.js
jackmu95: For me it was the same on Chrome and Firefox.

src/mixins/focus-in.js
```diff
@@ -3,8 +3,7 @@ import { eventOff, eventOn } from '../utils/dom'
 export default {
   data () {
     return {
-      focusInElement: null,
```
tmorehouse: Just thinking, for future reference to future coders, it might be good to create this non-reactive property during the `beforeCreate()` hook:

```js
  beforeCreate () {
    // Create non-reactive property
    this.focusInElement = null
  }
```

tmorehouse: I think once we add this to the mixin, all should be good to merge.

tmorehouse:
Looks good.. any other tweaks you need before merging?

jackmu95:
I'am done here and really happy with what we came up here.

tmorehouse:
Thanks for the great work!

PR #2260, opened 2018-12-10
feat(table): Selectable rows (fixes #1790)

### LINKED ISSUE #1790: Tables: select feature
Feature request.

I'm using Datatables and was thinking about using the Tables component instead, and there is only one thing I miss for that: ability for the user to select items, single, multiple (and the great "os" UI mode of Datatables, allowing to select multiple items using shift or ctrl/cmd keys)

[Datatables example](https://datatables.net/extensions/select/examples/initialisation/multi.html)

It is of course possible to implement it on top of the Table component, but I think it would be a great feature.

lianee:

Allow user to select rows. As this only works for visible rows, this feature rely on item keys.

see #1790

## PR checklist:

- [ ] Bugfix
- [x] Feature
- [ ] ARIA accessibility
- [x] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -69,6 +69,7 @@ these names):
 | `_cellVariants` | Object | Bootstrap contextual state applied to individual cells. Keyed by field (Supported values: `active`, `success`, `info`, `warning`, `danger`). These variants map to classes `table-${variant}` or `bg-${variant}` (when the `dark` prop is set).
 | `_rowVariant` | String | Bootstrap contextual state applied to the entire row (Supported values: `active`, `success`, `info`, `warning`, `danger`). These variants map to classes `table-${variant}` or `bg-${variant}` (when the `dark` prop is set)
 | `_showDetails` | Boolean | Used to trigger the display of the `row-details` scoped slot. See section [Row details support](#row-details-support) below for additional information
+| `_selected` | Boolean | UI triggered. See section [Selectable rows](#row-select-support) below for additional information
 
 **Example: Using variants for table cells**
 ```html
@@ -901,6 +902,52 @@ event will not be emitted when the input, select, textarea is clicked (unless th
 (even when disabled)
 
 
+## Row select
+You can make rows selectable, by using the prop `selectable`.
+
+By default, any row clicked is toggled, any other deselected. the SHIFT key selects a range of
+rows, and CTRL/CMD click will toggle the selected row.
+
+The prop `select-mode` can change the selecting mode:
+ - `multi`: each click will select/deselect the row
+ - `single`: only a single row can be selected at a time
+
+When a table is `selectable` and the user clicks on a row, `<b-table>` will emit the `row-selected`
+event, passing a single argument which is the complete list of selected items.
+**Treat this argument as read-only.**
+
+**Note:** _Paging or sorting will clear the selection._
+
+```html
+<template>
+  <b-table selectable select-mode="os" :items="items" @row-selected="rowSelected"></b-table>
+  {{ selected }}
+</template>
+
+<script>
+export default {
+  data () {
+    return {
+      items: [
+        { isActive: true, age: 40, first_name: 'Dickerson', last_name: 'Macdonald' },
+        { isActive: false, age: 21, first_name: 'Larsen', last_name: 'Shaw' },
+        { isActive: false, age: 89, first_name: 'Geneva', last_name: 'Wilson' },
+        { isActive: true, age: 38, first_name: 'Jami', last_name: 'Carney' }
+      ],
+      selected: []
+    }
+  },
+  methods: {
+    rowSelected(items) {
+      this.selected = items
+    }
+  }
+}
+</script>
+
+<!-- table-selectable.vue -->
+```
+
 ## Row details support
 If you would optionally like to display additional record information (such as
 columns not specified in the fields definition array), you can use the scoped slot
@@ -1454,6 +1501,7 @@ When `b-table` is mounted in the document, it will automatically trigger a provi
 
     <!-- Main table element -->
     <b-table show-empty
+             selectable
              stacked="md"
              :items="items"
              :fields="fields"
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -341,7 +341,10 @@ export default {
               key: `row-${rowIndex}`,
               class: [
                 this.rowClasses(item),
-                { 'b-table-has-details': rowShowDetails }
+                {
+                  'b-table-has-details': rowShowDetails,
+                  [this.selectedClass]: this.selectedRows[rowIndex]
+                }
               ],
               attrs: {
                 'aria-describedby': detailsId,
@@ -575,6 +578,18 @@ export default {
       type: [Boolean, String],
       default: false
     },
+    selectable: {
+      type: Boolean,
+      default: false
+    },
+    selectMode: {
+      type: String,
+      default: ''
+    },
+    selectedClass: {
+      type: String,
+      default: 'table-primary'
+    },
     headVariant: {
       type: String,
       default: ''
@@ -692,7 +707,9 @@ export default {
       // Our local copy of the items. Must be an array
       localItems: isArray(this.items) ? this.items : [],
       // Flag for displaying which empty slot to show, and for some event triggering.
-      isFiltered: false
+      isFiltered: false,
+      selectedRows: [],
+      lastRowClicked: -1
     }
   },
   mounted () {
@@ -733,16 +750,28 @@ export default {
       }
       this.localSortBy = newVal || null
     },
+    perPage (newVal, oldVal) {
+      if (oldVal !== newVal) {
+        this.clearSelected()
+      }
+    },
+    currentPage (newVal, oldVal) {
+      if (oldVal !== newVal) {
+        this.clearSelected()
+      }
+    },
     // Update .sync props
     localSortDesc (newVal, oldVal) {
       // Emit update to sort-desc.sync
       if (newVal !== oldVal) {
+        this.clearSelected()
         this.$emit('update:sortDesc', newVal)
       }
     },
     localSortBy (newVal, oldVal) {
       if (newVal !== oldVal) {
-        this.$emit('update:sortBy', newVal)
+      this.clearSelected()
+      this.$emit('update:sortBy', newVal)
       }
     },
     localBusy (newVal, oldVal) {
@@ -752,6 +781,8 @@ export default {
     },
     // Watch for changes on computedItems and update the v-model
     computedItems (newVal, OldVal) {
+      // Reset for selectable
+      this.lastRowClicked = -1
       this.$emit('input', newVal)
     },
     // Watch for changes to the filter criteria and filtered items vs localItems).
@@ -822,7 +853,8 @@ export default {
         // The following are b-table custom styles
         'b-table-fixed': this.fixed,
         'b-table-stacked': this.stacked === true || this.stacked === '',
-        [`b-table-stacked-${this.stacked}`]: this.stacked !== true && this.stacked
+        [`b-table-stacked-${this.stacked}`]: this.stacked !== true && this.stacked,
+        'b-table-selectable': this.selectable
       }
     },
     headClasses () {
@@ -1186,6 +1218,9 @@ export default {
       // Return the generated function
       return fn
     },
+    clearSelected() {
+      this.selectedRows = []
+    },
     // Event handlers
     rowClicked (e, item, index) {
       if (this.stopIfBusy(e)) {
@@ -1194,6 +1229,36 @@ export default {
       } else if (filterEvent(e)) {
         // clicked on a non-disabled control so ignore
         return
+      }
+      if (this.selectable) {
+        let selected = !this.selectedRows[index]
+        switch (this.selectMode) {
+          case 'single':
+            this.clearSelected()
+            break
+          case 'multi':
+            break
+          default:
+            if (this.lastRowClicked >= 0 && e.shiftKey) { // range
+              for (let idx=Math.min(this.lastRowClicked, index); idx <= Math.max(this.lastRowClicked, index); idx++) {
+                this.selectedRows[idx] = true
+              }
+            } else {
+              if (!(e.ctrlKey || e.metaKey)) { // clear range selection if any
+                this.clearSelected()
+              }
+              this.lastRowClicked = selected ? index : -1;
+            }
+            break
+        }
+        this.$set(this.selectedRows, index, selected)
+        let items = []
+        this.selectedRows.forEach((v, idx) => {
+          if (v) {
+            items.push(this.computedItems[idx])
+          }
+        })
+        this.$emit('row-selected', items)
       }
       this.$emit('row-clicked', item, index, e)
     },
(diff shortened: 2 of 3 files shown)


src/components/table/README.md
```diff
@@ -69,6 +69,7 @@ these names):
 | `_cellVariants` | Object | Bootstrap contextual state applied to individual cells. Keyed by field (Supported values: `active`, `success`, `info`, `warning`, `danger`). These variants map to classes `table-${variant}` or `bg-${variant}` (when the `dark` prop is set).
 | `_rowVariant` | String | Bootstrap contextual state applied to the entire row (Supported values: `active`, `success`, `info`, `warning`, `danger`). These variants map to classes `table-${variant}` or `bg-${variant}` (when the `dark` prop is set)
 | `_showDetails` | Boolean | Used to trigger the display of the `row-details` scoped slot. See section [Row details support](#row-details-support) below for additional information
+| `_selected` | Boolean | UI triggered. See section [Selectable rows](#row-select-support) below for additional information
```
tmorehouse: Is this actually used?  I don't see it referenced in the updated code.

src/components/table/README.md
```diff
@@ -901,6 +902,52 @@ event will not be emitted when the input, select, textarea is clicked (unless th
...
+
+When a table is `selectable` and the user clicks on a row, `<b-table>` will emit the `row-selected`
+event, passing a single argument which is the complete list of selected items.
+**Treat this argument as read-only.**
+
+**Note:** _Paging or sorting will clear the selection._
+
+```html
+<template>
+  <b-table selectable select-mode="os" :items="items" @row-selected="rowSelected"></b-table>
```
tmorehouse: Ans what is `select-mode="os"`?  Shouldn't it be one of `single` or `multi`?

src/components/table/table.js
```diff
@@ -575,6 +578,18 @@ export default {
       type: [Boolean, String],
       default: false
     },
+    selectable: {
+      type: Boolean,
+      default: false
+    },
+    selectMode: {
+      type: String,
+      default: ''
```
tmorehouse: Should this default to mode `single`

src/components/table/README.md
lianee: you're right, this is from a previous version and not used anymore. I need to fix this

src/components/table/README.md
tmorehouse: Yeah, as the mode you are using is not documented as to how it works... and if it is a "default" fallback, then it probably should have an explicit mode name, and any other mode that isn't supported falls back to it.

lianee: I though this text would describe it, but it seems it is not clear enough, I'll see how I can rephrase it:

> By default, any row clicked is toggled, any other deselected. the SHIFT key selects a range of rows, and CTRL/CMD click will toggle the selected row.

src/components/table/table.js
```diff
@@ -575,6 +578,18 @@ export default {
...
+    },
+    selectedClass: {
```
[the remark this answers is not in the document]
lianee: I'm not sure. I personally prefer not to be limited to "variant" classes, and as there are already other props accepting free class names...

src/components/table/table.js
lianee: No idea.
My personal preference is for the "OS" mode, OTOH I think the "multi" mode might be more intuitive, and "single" of limited use, but that really need thinking as it is a default value, not easily changed after

src/components/table/README.md
tmorehouse: Maybe make a default mode of `range`... although I think `single` mode should be the default.

src/components/table/table.js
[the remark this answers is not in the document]
tmorehouse: Those props are for additional "static" classes (or attributes) applied to structute elements within the components.

My preference would be to use the variant nomenclature as it fits with the `_rowVariant`.  And people can _easily_ define their own variants in CSS.

src/components/table/table.js
tmorehouse: Users can easily change it by setting the `select-mode` prop.

src/components/table/README.md
tmorehouse: But the default mode needs a name.

src/components/table/table.js
tmorehouse: I think single mode is more intuitive to the users of the table, which may not know the shift/CMD keyboard controls needed for multi/range

lianee: the beauty of the 'range' mode is that if the user doesn't use SHIFT, CTRL/CMD keys, it behaves like the 'single' mode.
I see the 'single' mode more like a limited mode when one just want a single selection (for editing a record for ex)

src/components/table/table.js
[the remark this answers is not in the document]
lianee: done

src/components/table/README.md
lianee: I rephrased that part, is it better (not yet changed the default mode though)?

src/components/table/table.js
tmorehouse: Ok, just played with multi, and I can see that being the default.  We do need to name the other mode as `range`, and all modes should be listed in the possible values for `select-mode` (noting which one is default)

src/components/table/README.md
tmorehouse: And maybe bind the mode to a select so users can change the mode to see the differences in the example.

We also need to mention that selection only works for the displayed rows, and is cleared when pagination or filtering happens.

src/components/table/table.js
```diff
@@ -343,7 +343,7 @@ export default {
                 this.rowClasses(item),
                 {
                   'b-table-has-details': rowShowDetails,
-                  [this.selectedClass]: this.selectedRows[rowIndex]
+                  ['table-'+this.selectedVariant]: this.selectedRows[rowIndex]
```
tmorehouse: You should check if `this.dark` is true, and if so, switch from `table-{variant}` to `bg-{variant}`

src/components/table/README.md
```diff
@@ -901,6 +901,51 @@ event will not be emitted when the input, select, textarea is clicked (unless th
 (even when disabled)
 
 
+## Row select
```
tmorehouse: This title should be "Row select supoort"

src/components/table/table.js
tmorehouse: Still think this should have a default value explicitly set, rather than an empty string to better convey the default mode.

src/components/table/README.md
lianee: done

src/components/table/README.md
```diff
@@ -901,6 +901,51 @@ event will not be emitted when the input, select, textarea is clicked (unless th
...
+
+When a table is `selectable` and the user clicks on a row, `<b-table>` will emit the `row-selected`
+event, passing a single argument which is the complete list of selected items.
+**Treat this argument as read-only.**
+
+**Note:** _Paging or sorting will clear the selection._
+
+```html
+<template>
+  <b-table selectable select-mode="multi" selectedVariant="success" :items="items" @row-selected="rowSelected"></b-table>
```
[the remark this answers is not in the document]
lianee: done

src/components/table/README.md
```diff
@@ -901,6 +901,51 @@ event will not be emitted when the input, select, textarea is clicked (unless th
 (even when disabled)
 
 
+## Row select
+You can make rows selectable, by using the prop `selectable`.
+
+Users can easily change the selecting mode by setting the `select-mode` prop.
+ - `single`: only a single row can be selected at one time
+ - `multi`: each click will select/deselect the row
```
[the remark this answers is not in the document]
lianee: done

src/components/table/table.js
lianee: done

src/components/table/table.js
lianee: an empty string better explain a "default" mode, and is more linked to the code

tmorehouse:
Also thinking, as part of the render function for the rows, we should add a new class to the selected rows. Maybe `b-row-selected`, and then users can target this class with custom styles if they want any special handling of the row/cell styles.

lianee:
ahem, wasn't it my previous "selectedClass" that I changed to "selectedVariant"?

src/components/table/README.md
```diff
@@ -901,6 +901,65 @@ event will not be emitted when the input, select, textarea is clicked (unless th
...
+</template>
+
+<script>
+export default {
+  data () {
+    return {
+      modes: [
+        { value: 'multi', text: 'multi' },
+        { value: 'single', text: 'single' },
+        { value: 'range', text: 'range' }
```
tmorehouse: You could just pass a simple array of string modes, since both text and value are the same (b-select with magically convert the simple array of strings to the array of objects form)

lianee: yes but I was wondering if you'd prefer to change the label

tmorehouse:
The selectVariant is correct, but I think we should also add a `b-row-selected` class to the `<tr>`, which would give the user an option to do additional custom styling over and above the variant of they wish. They can then write custom css to target `.b-table .b-row-selected` if they so choose.

src/components/table/README.md
tmorehouse: I think the label (option text) should match the actual mode value. Was just thinking that a simple text array would shorten the code.

src/components/table/table.js
tmorehouse: But if people jump to the component reference, and check the props for defaults they see no mode. And most users wouldn't be familiar with the code. They just want to see the options.

lianee:
pretty sure it's redundant, but anyway, I add the b-row selected class

src/components/table/table.js
lianee: done

tmorehouse:
It is similar to what we do with the row details... people can target that class to apply special styling if needed.

tmorehouse:
Also allows people to set the selected variant to an empty sting, if they just want to target only the b-row-selected class via custom css.

lianee:
yes, like I can target ".b-table .table-primary"...
I have to go, I'll check all notes when I'm back

lianee:
or simply set selectedVariant to "myown", and apply ".b-table .table-myown"

tmorehouse:
Gives the users multiple options for styling

tmorehouse:
Made a few minor tweaks and added passing the rowSelected state to the scoped slots for the row (as I figured someone will probably ask for this feature)

tmorehouse:
I think we need to add a bit of aria accessibility to this, so that screen reader users will know that a row is selected, and a way for them to select the row. Doing a bit of research on this now.

tmorehouse:
Still working on how screen reader or keyboard only users would select the rows.

Thinking that each row would need tabindex set, and listen for ckick/space/enter on the row (not just click). But I think this is needed whenever there is a row-clicked or row-selected listener registered.

tmorehouse:
Going to play with this a bit before merging (to test accesibility). Will add a note in documentation that rows will be placed in the document tab sequence when selectable and/or row clicked handles are registered (I know some people will complain that rows will have a focus ring when clicked, but its needed for accesibility)

lianee:
well, it seems that a css exists for taking care that focus ring, unfortunately, it is only supported by firefox at this time: [focus-visible](https://caniuse.com/#search=focus-visible)
Other solutions I found look pretty much overkill

tmorehouse:
And we do similar for sortable header cells.. they have the focus ring when active.

tmorehouse:
Overall, the functionality works well. Just going to add a few notes in the docs, and will merge after.

tmorehouse:
I found one small bug with respect to filtering. If the table was filtered and the filter criteria changed (and table is still in the filtered state), the selection wasn't cleared. I've added in one extra check for that situation.

tmorehouse:
So I think all is good now.

tmorehouse:
Yeah, open a new PR...  I think we can stick with single and multi... is there any way to merge multi and range modes?

tmorehouse:
Would just watching for changes on computedItems be enough to trigger the clearSelected? The array ref is new each time sorting, pagination or filtering run.

tmorehouse:
Although changes to row details would trigger new values, and would clear the selections if people show/hide the details.

lianee:
yes, tried that already and show/hide details trigger the watch, not desired

tmorehouse:
Would almost need to use a loose equals, and filter each row properties, and then compare row by row. Would add a bit of overhead though.

lianee:
what do you mean with "merging multi and range modes", getting rid of the range mode?

tmorehouse:
Although people should be deleting rows from a copy, and then replace items with a new array, rather than manipulating the local items array directly. 

May have to use the computedItems watcher initially, and just list a caveat that showing and hiding details will clear the selections.

tmorehouse:
Yeah, multi and range modes could be merged... I was playing with a multi select select element and it supports shift click, and individual click (select/clear).

lianee:
TBH, i mimicked the modes from [datatables](https://www.datatables.net/extensions/select/), I used it since a long time, and I guess that if it is good since many years, it can't really be wrong...

tmorehouse:
But for now, we could leave all three modes as is, but later consider merging range and multi into just multi

tmorehouse:
Just was thinking that multi could emulate how native select inputs work. Although maybe it is different on different OS/browsers (was playing on windows with chrome)

lianee:
well, that is the way range mode works, similar to OS selection.

lianee:
multi is designed to work the way checkboxes work, click toggles (hence my suggestion of synonyms before), more intuitive than range, but less powerfull because range works like single by default, like multi when used with CTRL and add range selection with SHIFT, they really are 3 different beasts

tmorehouse:
Synonyms might be handy initially..

tmorehouse:
Bit I think the main thing is to deal with things like item deletion/adding and clearing the selected array

lianee:
so, that is range mode

> Personally, I would like to be able to click to select and clear individual items, and use click, followed by shift click to select and arbitrary range, although not sure about range clearing, if that would be weird for users...

lianee:
yes bugs first, I'll have to do a couple of check with remote data too, then I'll open a new PR for bugs

tmorehouse:
What do you think about storing the actual selected items in data, and then watching it for changes to trigger the emit row-selected?

tmorehouse:
Watching the prop items might be a good start. And also could clear the selection when the provider runs... or could watch localItems for changes and trigger the clear too..

lianee:
that was my first take on that thing, but that was not satisfyng because:
- it mutates data
- is cleared with remote data
- is less performant with big arrays
- maybe other things I forgot...

tmorehouse:
Although, localItems and items changes typically would be things that would cause computedzitems to change.

I do have code that I was going to use for filter data change detection that filters out the item props using reserved property names, and then does a row by row loose equals compare until the first different row is found, and trigger an event. It would only need to fun against computed items, which will be only as large as the pagination size.. it's not overly fast, as when no changes happen it has to scan all the computedItems, but may be a solution.

Only time it would have issues is if there are duplicate rows without unique values.

tmorehouse:
I'll post the code for the loose watcher here tomorrow.. just getting ready to head to bed for the night.

You can take a look at it and see if we can just use it on the computedItems watcher to detect when to clear the selection. We could make the check conditional too, so that it doesn't run if the table is not in selectable mode.

PR #2267, opened 2018-12-11
fix(table): Clear selection when data change

lianee:

for array data, change is detected with the items watcher, works ok when changes are huge, added or deleted items for example.
localItems watcher is to be avoided because it is triggered for details
But for remote data (provider function), the watcher is not called, but as we need to call the refresh method anyway, clear selection is made there.
This refresh call could also be used for array data if need greater control.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] When resolving a specific issue, it's referenced in the PR's title (i.e. `fixes #xxxx[,#xxxx]`, where "xxxx" is the issue number)
- [ ] The PR should address only one issue or feature. If adding multiple features or fixing a bug and adding a new feature, break them into separate PRs if at all possible.

### DIFF
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -746,6 +746,7 @@ export default {
   watch: {
     // Watch props for changes and update local values
     items (newItems) {
+      this.clearSelected()
       if (this.hasProvider || newItems instanceof Function) {
         this.$nextTick(this._providerUpdate)
       } else if (isArray(newItems)) {
@@ -1272,6 +1273,7 @@ export default {
         let selected = !this.selectedRows[index]
         switch (this.selectMode) {
           case 'single':
+          case 'radio':
             this.selectedRows = []
             break
           case 'range':
@@ -1385,6 +1387,7 @@ export default {
         // Can't force an update when busy
         return false
       }
+      this.clearSelected()
       if (this.hasProvider) {
         this.$nextTick(this._providerUpdate)
       } else {
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -772,16 +772,6 @@ export default {
         this.clearSelected()
       }
     },
-    perPage (newVal, oldVal) {
-      if (oldVal !== newVal) {
-        this.clearSelected()
-      }
-    },
-    currentPage (newVal, oldVal) {
-      if (oldVal !== newVal) {
-        this.clearSelected()
-      }
-    },
     // Update .sync props
     localSortDesc (newVal, oldVal) {
       // Emit update to sort-desc.sync
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -1402,6 +1402,7 @@ export default {
         // Can't force an update when busy
         return false
       }
+      this.clearSelected()
       if (this.hasProvider) {
         this.$nextTick(this._providerUpdate)
       } else {
--- a/src/components/table/table.js
+++ b/src/components/table/table.js
@@ -746,7 +746,6 @@ export default {
   watch: {
     // Watch props for changes and update local values
     items (newItems) {
-      this.clearSelected()
       if (this.hasProvider || newItems instanceof Function) {
         this.$nextTick(this._providerUpdate)
       } else if (isArray(newItems)) {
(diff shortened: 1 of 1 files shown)


tmorehouse:
Here is the code I was mentioned the other day of for the `computedItems` watcher:

```js
  watch: {
    // ...
    computedItems (newVal, OldVal) {
      // Watch for changes on computedItems and update the v-model
      this.$emit('input', newVal)
      let equal = false
      if (this.selectable && this.selectedRows.length > 0) {
        // Quick check against array length
        equal = isArray(newVal) && isArray(oldVal) && newVal.length === oldVal.length
        for (let i = 0; equal && i < newVal.length; i++) {
          // Look for the first non-loosely equal row, after ignoring reserved fields
          equal = looseEqual(sanitizeRow(newVal[i]), sanitizeRow(oldVal[i]))
        }
      }
      if (!equal) {
        this.clearSelected()
      }
    },
    selectable (newVal, oldVal) {
      // Clear selection if prop selectable changes
      this.clearSelected()
    }
    // ...
  }
```

And this should be the only two places we need to trigger `this.clearSelected()`

tmorehouse:
Was thinking about changes such as show details, or any other dynamic change to the reserved item properties

lianee:
well, there is a good chance that a change in details won't trigger my code, in any case, that's not a performance issue, as your code will run on a few items only, I was just concerned with the length of code, but you surely know better, and if you think it's better, I see no reason for not using that.

PR updated

lianee:
just thinking, even with your checks, there is still a chance that we run into case where changes are not detected (many similar items). In that case, would it hurt to also put a clearSelected in refresh?

tmorehouse:
Was just thinking that it would reduce the number of places that we would need to add in the `this.clearSelected()` call, although it won't hurt to have it in the various watchers (as it will set selectedItems array length to 0, which the computedItems watcher checks first before doing the larger "deep" check)

tmorehouse:
It definitely should be called in the refresh method as you suggest.

lianee:
I'm writing a more comprehensive test case, with details and similar records and I'll play a little with it, and yes, it's likely there are now useless clears, I should tidy it up a bit

lianee:
I also thought at first it is problem, but if the editing produces a sort change (change the last name while sorting on that value), that might goes awok.
need more tests

lianee:
ok, so potential problem is only in array items, since provider always clears.
the edit case I observed was when there was no sorting on the table, with sorting, the selection is cleared (I couldn't find a case when it did not)
so, it seems it is not that big a problem

tmorehouse:
Ok, we can merge, and if any edge case issues crop up, we can create a new issue.

tmorehouse:
In the edge case where there are multiple identical rows, we should suggest that the users include a unique row ID to ensure uniqueness.

lianee:
not even sure of that, when the user will make a change at one of these identical rows, it'll will then be different. I couldn't break it that way

tmorehouse:
I was just thinking of when a row is not edited and sorting changes (or possibly an identical row is deleted), but results in an identical computed items due to the duplicates, the selections may not get cleared.

lianee:
I think I tried these cases already, but I might be wrong:
- identical row deleted: selection is cleared if the resulting rows are different, selection not cleared otherwise, but it is an identical row
- sort change: didn't find a way when that fails
About sorting, I found an interesting thing in my tests: if you try to sort on an object item (like name in the complete table example), that won't work, because toString of an object is undefined. I know that object order is not stable, but JSON.stringify gives much better results than toString in that case

tmorehouse:
We might need to update out recToString/testing methods

lianee:
I stand corrected: I based this remark on the documentation on "Sort-Compare routine", the actual sort routine is stronger... but still gives incorrect results for objects

tmorehouse:
I'll have to take a look at the stringification method (and maybe make it sort property names first)

lianee:
don't waste your time, seems I need a rest, sort works ok on the complete example, so it must be something on my side.
Above all, this has nothing to do with this PR, I'll get back on that if I have something more serious, sorry.

tmorehouse:
No problem.. if we do need to tweak the stringification, we can open a new issue or pr.

So sage to merge this?

lianee:
yes please, before I think of any other stupid thing. I'll then open a PR if needed

tmorehouse:
Crap... I did a merge commit instead of a squash and merge

PR #2360, opened 2018-12-28
Fix(table): slot details when using tbody-tr-class="d-flex"

williamsantc:


src/components/table/table.js
```diff
@@ -1338,7 +1338,7 @@ export default {
         )
         // Row Details slot
         if (rowShowDetails) {
-          const tdAttrs = { colspan: String(fields.length) }
+          const tdAttrs = { colspan: String(fields.length), class: 'col-12' }
```
tmorehouse: This will probably add unexpected padding on the spanned column (`td`) which people may not be expecting.

It should be conditional, or we should add an option to apply additional classes to the details `td`

tmorehouse: Also note that not everyone uses 12 columns in their custom bootstrap set up (I.e. 10 columns).

tmorehouse:
This could also be addressed by using custom css that targets the row-details cell and sets a display/width values.

Could you first open an issue on this so that it can be discussed first, as this PR is an "always on" fix for one edge case and may introduce issue for other users use cases or styling.

tmorehouse:
Closing as there would be a better way to handle this than hard-coding a specific class (which may have conflicts with custom bootstrap column counts).

Issue #1967, opened 2018-07-25
Elements with :to attribute use 'active' class instead of vue-router classes


claviska:
When using vue-router, the following produces dropdown items with the `active` class instead of the appropriate router classes:

```html
<b-dropdown-item to="/">Home</b-dropdown-item>
<b-dropdown-item to="/about">About</b-dropdown-item>
```

Assuming the current path is `/about`:

Result:
```html
<a href="/" class="dropdown-item active" role="menuitem" target="_self">Home</a>
<a href="/about" class="dropdown-item active" role="menuitem" target="_self">About</a>
```

Expected:
```html
<a href="/" class="dropdown-item router-link-active" role="menuitem" target="_self">Home</a>
<a href="/about" class="dropdown-item router-link-exact-active" role="menuitem" target="_self">About</a>
```

Fiddle: https://jsfiddle.net/xpjweqg0/

This appears to be a side effect inherited from the [link component](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/link/link.js), which sets both `activeClass` and `exactActiveClass` to the same value. However, the `active` class probably isn't desirable when router links are used in dropdown items, as it mimics the hover/focus styles.

I realize these can be overridden by specifying `active-class` and `exact-active-class` attributes on each dropdown item, but I expected the router's classes would be default.

Is this a bug or by design?

claviska:
I just noticed that this appears to affect `<b-button>` and likely all elements that utilize the link component.

It's actually as simple as this: https://jsfiddle.net/xpjweqg0/3/

In this example, the button with `to="/"` always has an active state because the link component uses 'active' for both `activeClass` and `exactActiveClass`.

tmorehouse:
Bootstrap uses the class `active` to highlight active components, which is why, if you want vue-router to control the active state, you will need to override vue-router's `active-class` prop on components that you are using `to` props to define the destination.

tmorehouse:
Working as expected.

PR #2388, opened 2019-01-04
fix(router-link): remove default values for active-class and exact-active-class (Fixes #2387)

### LINKED ISSUE #2387: bug[b-link]: Global active-class and exact-active-class overridden by prop defaults
When a global `active-class` and `exact-active-class` are defined in `vue-router` or `nuxt.js`, `<b-link>` (and all components that inherit from `<b-link>`) is overriding these by specifically defaulting the props to the Bootstrap V4 class `active`.

Related issue #1967 

Changing the props to default to `undefined` would remedy this, but might be constituted as a breaking change

tmorehouse:

Removes the default values (setting to `undefined`) for props `active-class` and `exact-active-class`, so that the globally configured values can be used by default.

Affects `b-link`, `b-button`, `b-dropdown-item` and other components that build on `b-link` (i.e. have a `to` prop).

Fixes #2387, #1967

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Components that accept a `to` prop will no longer default to using `active` as the class when the router-link is in the active state, but will instead use the globally configured values from vue-router (or nuxt) if the props `active-class` and `exact-active-class` are not set.

**While this change does not affect functionality of the component(s), it _may affect_ the appearance if the user has not provided values to the props.**

### DIFF
--- a/docs/components/nav.vue
+++ b/docs/components/nav.vue
@@ -46,11 +46,11 @@
         <b-nav-item
           to="/docs"
           exact>Docs</b-nav-item>
-        <b-nav-item to="/docs/components">Components</b-nav-item>
-        <b-nav-item to="/docs/directives">Directives</b-nav-item>
-        <b-nav-item to="/docs/reference">Reference</b-nav-item>
-        <b-nav-item to="/docs/misc">Misc</b-nav-item>
-        <b-nav-item to="/play">Play</b-nav-item>
+        <b-nav-item to="/docs/components" active-class="active">Components</b-nav-item>
+        <b-nav-item to="/docs/directives" active-class="active">Directives</b-nav-item>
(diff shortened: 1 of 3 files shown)


docs/components/nav.vue
```diff
@@ -60,9 +60,6 @@
           right>
           <b-dropdown-item href="http://dev.bootstrap-vue.netlify.com">Development</b-dropdown-item>
```
jackmu95: Shouldn't we use `https://` here because there is a redirect anyway.
A SSL cert for this subdomain would be ideal.

tmorehouse: Yeah, although I don't believe this link actually works atm, and will need to be updated before release.

PR #2444, opened 2019-01-12
chore: Add `prettier` formatting for all files

jackmu95:

This PR adds `prettier` formatting for all `.js`, `.vue`, `.scss` and `.md` files.
Eslint is integration is given and all test pass.

## PR checklist:

- [ ] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [x] Other, please describe:

- [ ] Yes
- [x] No


docs/components/codemirror.vue
```diff
@@ -61,19 +62,19 @@ export default {
       default: false
     }
   },
-  data () {
+  data() {
```
tmorehouse: I wish prettier would get off their butts and add in the option/switch to allow the space here, as it makes searching for the function definition so much easier knowing that the def has the space, while usage does not.

docs/components/componentdoc.vue
```diff
@@ -1,106 +1,148 @@
...
+          <code>&lt;{{ kebabCase(alias) }}&gt;</code>
+        </li>
       </ul>
     </article>
 
     <article v-if="props_items && props_items.length > 0">
-      <h4 :id="`comp-ref-${componentName}-props`">Properties</h4>
+      <h4 :id="`comp-ref-${componentName}-props`">
+        Properties
+      </h4>
```
tmorehouse: The only thing about this, is that all that whitespace (inside the `<h4>`) is stored in the compiled render function.

jackmu95: I think this is related to the [`vue/singleline-html-element-content-newline`](https://eslint.vuejs.org/rules/singleline-html-element-content-newline.html) ESlint rule.

docs/components/codemirror.vue
jackmu95: Yeah, I wish this would be configurable too.

docs/components/componentdoc.vue
tmorehouse: Might be.  Things like inline elements (`<span>`, `<strong>`, etc) should all be allowed to be inlined without a line break, to reduce final render function size (unless tehre is a rule to strip whitespace, but then that can affect things like `<pre>` etc.

The following will not render the same and will have slightly different spacing on screen:

```html
<span><input type="checkbox"><label>Foobar</label></span>
```

vs.

```html
<span>
  <input type="checkbox">
  <label>Foobar</label>
</span>
```

The latter will introduce spacing between the label and the input (via whitespace), while the former will have night tight spacing.  Inspecting the rendered output in chrome dev tools will show the extra whitesapce.

image

So I don't always agree to breaking tags across lines, specifically when one doesn't want to introduce side effects of adding spaces between tags.  There may be times where we want to have no spaces between the tags (images and outer containers is usually a big one where the white space affects image alignment).

Is there a way to override this?

tmorehouse: The other issue is in the live examples in the README docs.  We try to reduce the "height" of the code (i.e. not breaking some tags across multiple lines), so that the user doesn't have miles to scroll on the screen to get to the next section of docs.  If an example can be written on one line, it should stay that way (i.e. `<b-button @click="foo">Bar</b-button>` should remain as a single line and not broken across three lines).

docs/components/codemirror.vue
pi0: The entire point with prettier is using an optionized preset.

docs/pages/play.vue
```diff
@@ -15,149 +14,222 @@
...
+          <form
+            class="d-inline-block ml-2 mr-0 p-0 float-right"
+            method="post"
+            action="https://jsfiddle.net/api/post/library/pure/"
+            target="_blank"
+          >
+            <input
+              type="hidden"
+              name="html"
+              :value="fiddle_html"
```
tmorehouse: I still prefer that things like this could stay on a single line (if htey fint on a single line).  As now six lines on screen have changed to 30 lines on screen.  Making the user (i,e. me :stuck_out_tongue: ) have to scroll up/down a lot more (when viewing or editing the file) in order to see what that component is doing.

Is there ea rule that can say if it fits in 80-100 characters, leave it alone?

jackmu95: I currently don't know why increasing `vue/max-attributes-per-line` to 3 doesn't help here :/

tmorehouse: Is there a maximum line width rule coming into play here?

jackmu95: I don't thin so. They are < 100 chars.

tmorehouse: Weird...

docs/components/codemirror.vue
tmorehouse: Prettier has a discussion going on about adding a flag/option to set space between method name and the opening parenthesis (specifically when using object shorthand to define a method, but also for regular function definitions), and no space when calling a method.

src/directives/scrollspy/scrollspy.class.js
```diff
@@ -100,8 +107,14 @@ function typeCheckConfig (componentName, config, configTypes) /* istanbul ignore
...
-          componentName + ': Option "' + property + '" provided type "' +
-                    valueType + '", but expected type "' + expectedTypes + '"'
+          componentName +
+            ': Option "' +
+            property +
+            '" provided type "' +
+            valueType +
+            '", but expected type "' +
+            expectedTypes +
+            '"'
```
tmorehouse: We could probably switch this to use interpolated string:
```js
warn(
  `${componentName}: Option "${property}" provided type "${valType}" but expected type "${expectedTypes}"`
)
```

Although this may exceed maximum ling length, but it is more readable.

docs/components/codemirror.vue
tmorehouse: If they add in the option in a future release, we can always enable it to get the space back.

PR #2474, opened 2019-01-18
prevent infinite looping in table.js' toString

btoo:

When `b-table` recursively converts an object to a string, it needs to make sure it doesn't get caught in an infinite loop if an object has a property that is a reference to itself

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


jackmu95:
@tmorehouse Is this really something we should handle or would a note in the docs be the better alternative.

tmorehouse:
The current solution in this PR looks only at one level of keys for duplicates in the object during stringification.

What might b better is to flatten the object(s) into an array of objects, and before concatenating a new value onto the array, see if the object exists in the array (excluding strings/numbers, as they can be duplicated without issue), and then convert the array into a concatenated string.

But a warning in the docs (for now) might suffice (in the filtering/sorting sections of b-table docs).

btoo:
i'm worried this may sometimes be an unsolvable issue in user-land when the `toString`ed data comes from a library which uses such objects as the one described in the PR - so i think this is something that would have to be eventually accounted for in the component itself

@tmorehouse does the fact that `toString` is being used recursively within itself mean it isnt a shallow (aka only at one level) stringification but a deep one?

tmorehouse:
@btoo the `toString()` method is indeed recursive, as table data entries can be an object.

So it needs to take into account possible nested objects that may appear in the same row or cell.

Imagine if someone placed a component reference, or even an HTML element, as a data item (which have circular references).  toString would cause an infinite loop.

One possible solution would be to add a limit to the recursion depth... to prevent stack overflows.

The easiest thing is just to warn users of circular references in the item row data (per row and per cell)

PR #2477, opened 2019-01-20
feat(security): strip html tags [master]

### LINKED ISSUE #1974: Script tag is not escaped when using :option="" in form-select
I was trying to do some XSS in my project and I got some un-shown text. Which I found out that it was the cause of the script tag that was not escaped.. So I tried doing a v-for loop inside an `<option> `tag and use 
`{{  }}` (Double brackets) to escaped the script tag. I would like to make a pull request sadly bootstrap_vue is using a render less components (I don't really know but it's using render function) which I don't have much knowledge about. So I just submitted an Issue here.

by the way you can replicate the problem using the Live documentation on Bootstrap_vue site.

My browser: Opera
Operating System: Fedora LXDE

pi0:

Clarification letter: https://gist.github.com/pi0/674d49d5f1c2ccfe20d3b1e29ae7b497

fixes #1974.

This PR disables implicit use of any `innerHTML` and removes all HTML tags in other places. setting `global/window.BV_UNSAFE_HTML` value to `true` enables legacy approach. (Nuxt users can easily set module option `unsafeHTML`)

## PR checklist:

- [x] Content prop
- [x] Stript Tags
- [ ] Security section docs
- [ ] Notes about prefering Slots for HTML (credits to @mosinve)

- [x] Security

- [x] Yes
- [ ] No

2.x RC users that use custom markups in text props should prevent that or enable a special flag to enable unsafe `innerHTML`. As we are still in the RC phase, this change should worth.


src/components/dropdown/dropdown.js
```diff
@@ -26,7 +27,7 @@ export default {
             click: this.click
           }
         },
-        [this.$slots['button-content'] || this.$slots.text || this.text]
+        [this.$slots['button-content'] || this.$slots.text || stripTags(this.text)]
```
mosinve: Maybe `this.$slots.text` also does need to strip tags?

mosinve: Okay

PR #2591, opened 2019-02-15
fix(modal): stop event propagation on modal content click

kalvenschraut:

With the latest release (v2.0.0-rc-12), clicking on a button in a modal that triggers a content render change (a button that changes a variable that then hides the button via v-if) triggers the backdrop close event handler. This occurs because the markup no longer exists in the modal when contains is called so the code thinks the click didn't occur in the modal content. I added a click handler to the modal content div that stops click events from propagating up to the backdrop div, preventing the modal from closing.

codepen showing this case: https://codepen.io/Kalvens/pen/VgERdj
I am using chrome: 69.0.3497.100 (Official Build) (64-bit)
OS: linux

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


tmorehouse:
This might be considered a breaking change if people are using event delegation to listen for events higher up in the DOM.

A work around would be to use `v-show` rather than `v-if` to hide the elements, but keep them in DOM.

mosinve:
@tmorehouse But modal windows designed to block any interaction with parents. 
From  modal windows wiki:
> In user interface design for computer applications, a modal window is a graphical control element subordinate to an application's main window. It creates a mode that disables the main window but keeps it visible, with the modal window as a child window in front of it. Users must interact with the modal window before they can return to the parent application. This avoids interrupting the workflow on the main window.

tmorehouse:
@mosinve, that is true, but someone may be using event delegation in document to listen for events.

kalvenschraut:
A v-show instead of a v-if is not a viable option right now at least since I have a component using the b-popover in a modal that has a button in it, when that button is clicked the popover is closed. From what I can tell the main content of the popover is not in the DOM when it is not visible.

mosinve:
@tmorehouse, I agree, this could be a BC.

tmorehouse:
Popovers (and tootips) are mounted in the modal (cant recall off hand if it is modal content or the modal wrapper), when they detect that they are inside a modal.

jackmu95:
Adding the ".stop" modifier to the click event fixes the problem.

I'am not sure if preventing propagation for all events in the modal is a good idea.

A note in the docs with a concrete example would be the better option IMHO.

jackmu95:
@tmorehouse @mosinve: I've got another idea... What about adding the `onClickOut()` listener to the backdrop itself? That's also the way Bootstrap does in in their JS files.

tmorehouse:
@jackmu95 That might work, since the modal and backdrop are siblings.

tmorehouse:
Listening to backdrop element click won't work, as it is "under" the modal, and the modal has a tabindex (which makes it clickable, hijacking the click on backdrop).

PR #2608 handles the case where the clicked element is removed from document before the onClickOut handler runs (by checking that the evt.target still exists in DOM).

Tested the code from the above codepen in the deploy preview playground and the original issue has been resolved.

PR #2640, opened 2019-02-18
fix(utils): Make `looseEqual()` util handle File's comparison correctly

### LINKED ISSUE #2638: b-form-file: model doesn't change if another file is selected
When changing the selected file, the model isn't updated and still contains the first selected file.

On https://bootstrap-vue.js.org/docs/components/form-file/ 
First, select a file and notice how the "Selected file:" text now shows the selected filename. Then, select another file. The "Selected file:" text doesn't change and still shows the first filename. I would expect the newly selected filename to be shown here.

Tested on Firefox 65.0, Bootstrap version bootstrap-vue@eb07b196f3762

jackmu95:

This PR fixes a bug in the `looseEqual()` util that returned `true` if two [File](https://developer.mozilla.org/de/docs/Web/API/File) objects were passed because `Object.keys()` returns an empty array.

Closes #2638.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


src/utils/loose-equal.js
```diff
@@ -16,44 +19,27 @@ function isObject(obj) {
...
-        )
-      } else {
-        /* istanbul ignore next */
+  if (a === b) {
+    return true
+  }
+  if (isObject(a) && isObject(b)) {
+    // Using for loop over `Object.keys()` here since some class
+    // keys are not handled correctly otherwise
+    for (const key in a) {
```
tmorehouse: Should we not do a quick comparison of the number of keys in each object for speed?

i.e. the original Vue version did a quick check on `Object.keys(a).length !== Object.keys(b).length` and returned `false` if not equal.

tmorehouse: i.e. if we had the following two objects with the same shape, esxcept for one additional key in the b object:

```js
const a = { a: 1, b: 2, c: 3 }
const b = { a: 1, b: 2, c: 3, d: 4 }
```

jackmu95: We can do that before the for loop for sure.

src/utils/loose-equal.js
```diff
@@ -16,44 +19,27 @@ function isObject(obj) {
...
+      const bHasKey = b.hasOwnProperty(key)
+      if ((aHasKey && !bHasKey) || (!aHasKey && bHasKey) || !looseEqual(a[key], b[key])) {
         return false
       }
-    } catch (e) {
-      /* istanbul ignore next */
-      return false
     }
-  } else if (!isObjectA && !isObjectB) {
-    return String(a) === String(b)
```
tmorehouse: We may want to add this before the last `return false` as a catch all comparison.

jackmu95: Which exact case should this handle?

jackmu95: When there is a case for it we can simply replace the last `return false` with `return String(a) === String(b)`.

tmorehouse: Possibly a custom object/class/etc that may have a toString method. Vue placed it in their original code for some reason.

tmorehouse: Yeah, that may work.

We might want to add in a test spec for this util (based on the tests that Vue has for their version of `looseEqual()`

jackmu95: Yeah. Still not sure about the need of the `String(a) === String(b)`.

tmorehouse:
Just wondering if we just added in a test to see if the object is a file object, and then comparing equality (`===`) between the objects, similar to how the test for Date objects was done, and leave the rest of the code the same as the original Vue `looseEqual` code?  Just to ensure that it works the same as Vue's (except for the additional test for File object instance)?  Since looseEqual is used in many of the form components and others.

tmorehouse:
i.e.:

```js
function looseEqual(a, b) {
  if (a === b) return true
  const isObjectA = isObject(a)
  const isObjectB = isObject(b)
  if (isObjectA && isObjectB) {
    try {
      const isArrayA = isArray(a)
      const isArrayB = isArray(b)
      if (isArrayA && isArrayB) {
        return (
          a.length === b.length &&
          a.every((e, i) => {
            return looseEqual(e, b[i])
          })
        )
      } else if (a instanceof File && b instanceof File) {
        return a === b
      } else if (a instanceof Date && b instanceof Date) {
        return a.getTime() === b.getTime()
      } else if (!isArrayA && !isArrayB) {
        const keysA = keys(a)
        const keysB = keys(b)
        return (
          keysA.length === keysB.length &&
          keysA.every(key => {
            return looseEqual(a[key], b[key])
          })
        )
      } else {
        /* istanbul ignore next */
        return false
      }
    } catch (e) {
      /* istanbul ignore next */
      return false
    }
  } else if (!isObjectA && !isObjectB) {
    return String(a) === String(b)
  } else {
    return false
  }
}
```

jackmu95:
But I'am not sure if File objects are the only case where the current implementation makes problems. I'am going to create some unit tests to back this all up and then we can create a PR for Vue.

src/utils/loose-equal.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    const obj2 = { foo: 'bar1' }
+    const nestedObj1 = { ...obj1, bar: [{ ...obj1 }, { ...obj1 }] }
+    const nestedObj2 = { ...obj1, bar: [{ ...obj1 }, { ...obj2 }] }
+    expect(looseEqual(obj1, obj1)).toBe(true)
+    expect(looseEqual(obj1, { ...obj1 })).toBe(false)
+    expect(looseEqual(obj1, obj2)).toBe(false)
+    expect(looseEqual(nestedObj1, nestedObj1)).toBe(true)
+    expect(looseEqual(nestedObj1, { ...nestedObj1 })).toBe(false)
+    expect(looseEqual(nestedObj1, nestedObj2)).toBe(false)
+  })
```
tmorehouse: Maybe a test for objects defined with different key orders (as key order is typically retained in the order they were defined):

```js
const obj3 = { a: 1, b: 2, c: 3 }
const obj4 = { b: 2, c: 3, a: 1 }
expect(looseEuqal(obj3, obj4)).toBe(true)
```

src/utils/loose-equal.js
```diff
@@ -16,44 +20,31 @@ function isObject(obj) {
...
+      return false
+    }
+
+    // Using for loop over `Object.keys()` here since some class
+    // keys are not handled correctly otherwise
+    for (const key in a) {
+      const aHasKey = a.hasOwnProperty(key)
+      const bHasKey = b.hasOwnProperty(key)
+      if ((aHasKey && !bHasKey) || (!aHasKey && bHasKey) || !looseEqual(a[key], b[key])) {
         return false
```
tmorehouse: This should be:
```js
   return (aHasKey && !bHasKey) || (!aHasKey && bHasKey) || !looseEqual(a[key], b[key])
```

As the test for object with same key/values but in different order fails.

src/utils/loose-equal.js
```diff
@@ -16,44 +20,31 @@ function isObject(obj) {
...
       }
-    } catch (e) {
-      /* istanbul ignore next */
-      return false
     }
```
tmorehouse: Should probably return `true` after the for loop.

src/utils/loose-equal.js
```diff
@@ -16,44 +20,31 @@ function isObject(obj) {
...
-  } else if (!isObjectA && !isObjectB) {
-    return String(a) === String(b)
-  } else {
-    return false
   }
+  if (isArray(a) && isArray(b)) {
```
tmorehouse: These two ifs will never be called, since testing for isObject excludes arrays and dates

src/utils/loose-equal.js
tmorehouse: Since this returns false, and the fallback at the end returns false, object equality (loose) will not return `true` reliably when two objects have the same shape (but different key orders).

src/utils/loose-equal.js
```diff
@@ -28,39 +28,35 @@ function looseEqual(a, b) {
...
+      : false
   }
-  if (isArray(a) && isArray(b)) {
-    return a.length === b.length && a.every((e, i) => looseEqual(e, b[i]))
-  }
-  if (isObject(a) && isObject(b)) {
-    if (keys(a).length !== keys(b).length) {
+  const validObjectsCount = [isObject(a), isObject(b)].filter(v => v === true).length
+  if (validObjectsCount > 1) {
+    if (validObjectsCount === 1 || keys(a).length !== keys(b).length) {
```
tmorehouse: `validObjectsCount === 1` will never be true as the wrapping if is ` validObjectsCount > 1`

src/utils/loose-equal.js
```diff
@@ -28,39 +28,35 @@ function looseEqual(a, b) {
   if (a === b) {
     return true
   }
-  if (typeof a !== typeof b) {
```
tmorehouse: This would still be a quick bail if the types are different

src/utils/loose-equal.js
```diff
@@ -29,21 +29,21 @@ function looseEqual(a, b) {
     return true
   }
   const validDatesCount = [isDate(a), isDate(b)].filter(Boolean).length
-  if (validDatesCount > 1) {
+  if (validDatesCount > 0) {
```
tmorehouse: could this just be `if (validDatesCount) {`

And wondering if we should use a `let validCount` to re-use the same var to prevent a new stack entry for each `const`?

jackmu95: 1. Don't like that loose checks
2. 👍

src/utils/loose-equal.js
```diff
@@ -31,30 +31,32 @@ function looseEqual(a, b) {
...
   }
-  const validArraysCount = [isArray(a), isArray(b)].filter(Boolean).length
-  if (validArraysCount > 0) {
-    return validArraysCount === 2
+  validTypesCount = [isArray(a), isArray(b)].filter(Boolean).length
+  if (validTypesCount > 0) {
+    return validTypesCount === 2
       ? a.length === b.length && a.every((e, i) => looseEqual(e, b[i]))
       : false
   }
```
tmorehouse: Could this just be:
```js
  if (validTypesCount === 2) {
    return a.length === b.length && a.every((e, i) => looseEqual(e, b[i]))
  }
```
and change all the `if`'s to `else if`? and let the final return false kick in?

tmorehouse: Err... never mind....

Would need to move the computation of the validTypesCount into the if condition

src/utils/loose-equal.js
```diff
@@ -16,44 +25,52 @@ function isObject(obj) {
...
+      return false
+    }
+    if (aKeysCount === 0 && bKeysCount === 0) {
+      return String(a) === String(b)
+    }
+    // Using for loop over `Object.keys()` here since some class
+    // keys are not handled correctly otherwise
+    for (const key in a) {
+      if (
+        [a.hasOwnProperty(key), b.hasOwnProperty(key)].filter(Boolean).length === 1 ||
```
tmorehouse: Since we know `a` will have `key`, can we just check to see if `b` doesn't have the `key`?

jackmu95: I think this is the safest and shortest way to handle this.

tmorehouse: ie:
```js
if ( !b.hasOwnProperty(key) || !looseEqual(a[key], b[key])) {
  return false
}
```

jackmu95: The `a.hasOwnProperty(key)` check is also required in the for loop.

tmorehouse: ok

Issue #2513, opened 2019-02-04
Tree shaking not working


cliffordh:
I have a minimal bootstrap-vue application. I basically just used the Vue CLI to create a new project, and added bootstrap-vue as follows:

```
import BootstrapVue from "bootstrap-vue";
import "bootstrap/dist/css/bootstrap.css";
import "bootstrap-vue/dist/bootstrap-vue.css";
Vue.use(BootstrapVue);

```
I also added font-awesome but that doesn't appear to take up that much space...

When I perform a build using the Vue UI the build size seems enormous:

```
File                                   Size              Gzipped

  dist/js/chunk-vendors.efb23010.js      422.49 KiB        131.86 KiB
  dist/js/app.7c82e299.js                7.65 KiB          2.82 KiB
  dist/js/about.3a94e576.js              0.44 KiB          0.31 KiB
  dist/css/chunk-vendors.f1000a76.css    171.00 KiB        24.30 KiB
  dist/css/app.03795e02.css              0.53 KiB          0.31 KiB

```
I found another thread mentioning build size and it mentioned something about "esm" builds and I tried using a different import, but barely any difference:

`//import BootstrapVue from "bootstrap-vue/dist/bootstrap-vue.esm.min";`

Any suggestions? Thanks.

mikeerickson:
Curious, how about using a CDN release which will likely always use a cached version of the framework.

cliffordh:
Here's the visual output from source-map-explorer

<img width="1316" alt="screen shot 2019-02-04 at 9 31 51 am" src="https://user-images.githubusercontent.com/1755156/52214563-e88bf000-285f-11e9-8ef7-5cd8598fe604.png">

hertz1:
Try importing like this:

```js
import Navbar from 'bootstrap-vue/es/components/navbar'
Vue.use(Navbar)
```

cliffordh:
Thanks for responding. Unfortunately that does not make a difference. I'm using 2.0rc11. I'm going to try and register components individually to see if this helps. Could this be a bug in 2.0rc11?

cliffordh:
Individually importing the components and registering them results in a massive reduction in the amount of code imported, from 160k to about 10k. I wouldn't call this issue closed because I shouldn't have to do this, but given these results I think I can still proceed with using bootstrap-vue.

```
import bNavbar from "bootstrap-vue/es/components/navbar/navbar";
[...]

PR #2654, opened 2019-02-19
feat(nuxt-module): add tree-shaking support to Nuxt module

pimlie:

In response to @tmorehouse's answer in #2513, when using Nuxt.js we can easily support tree-shaking within the Nuxt module. This pr adds that.

A remark, there is no validation to check if the items listed in `plugins` are really BootstrapVue plugins, but that shouldnt be a big problem. We cant make everything foolproof and if you'd write your own imports there is also no validation except for the error you get when trying to build your project. If you'd really would like that anyway, then I'd prefer to ship a component/directive map with each release so we dont have to create that mapping during runtime.

I think it would also be very useful if we would have a `suggest` feature, where we do a grep on `<b-`, map every component we find to the corresponding plugin and print the list of suggested plugins for your project. Then you just use full bootstrap-vue during development and once thats ready and you want to optimize the project you can quickly add the list of plugins you need (although most often you'd also want to limit the list of globally installed components and which are only used on some pages). Let me know what you think and where/how I should add that.

## PR checklist:

- [ ] Bugfix
- [x] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

- [x] Includes documentation updates (including updating the component's `package.json` for slot and
      event changes)

description above includes:**


nuxt/index.js
```diff
@@ -1,39 +1,64 @@
...
-
-  // Transpile src
-  this.options.build.transpile.push('bootstrap-vue/src')
-
-  // Register plugin, pasing options to plugin template
-  this.addPlugin({
-    src: resolve(__dirname, 'plugin.template.js'),
-    fileName: 'bootstrap-vue.js'
+    const kebabCase = str => str.replace(/([_\s]+([a-zA-Z])|([A-Z]))/g, (m, $1, $2, $3, o) => (o ? '-' : '') + ($2 || $3 || '').toLowerCase())
+    const pascalCase = str => str.replace(/(^|[-_\s]+)(.)/g, (m, $1, $2) => $2.toUpperCase())
```
pimlie: I included these here because I think a oneliner is preferable above an external dependency

nuxt/index.js
```diff
@@ -1,39 +1,64 @@
 const { resolve } = require('path')
 
-function pickFirst(...args) {
-  for (const arg of args) {
-    if (arg !== undefined) {
-      return arg
+module.exports = function nuxtBootstrapVue(moduleOptions = {}) {
+  this.nuxt.hook('build:before', () => {
+    function pickFirst(...args) {
```
pimlie: Including this function in the hook scope should make for easier gc

nuxt/index.js
```diff
@@ -1,39 +1,64 @@
 const { resolve } = require('path')
 
-function pickFirst(...args) {
-  for (const arg of args) {
-    if (arg !== undefined) {
-      return arg
+module.exports = function nuxtBootstrapVue(moduleOptions = {}) {
+  this.nuxt.hook('build:before', () => {
```
pimlie: pi0 has said that he likes that nuxt-modules only run when they need to, unless I miss something I think this module only needs to run during build

nuxt/index.js
```diff
@@ -1,39 +1,64 @@
...
+    }
+
+    const bootstrapCSS = pickFirst(options.bootstrapCSS, options.bootstrapCss, options.css)
+    if (bootstrapCSS) {
+      // Add Bootstrap CSS
+      this.options.css.unshift('bootstrap/dist/css/bootstrap.css')
+    }
+
+    // Transpile src
+    this.options.build.transpile.push('bootstrap-vue/src')
```
pimlie: Is this still needed when we import everything from `bootstrap-vue/es` ?

tmorehouse:
As well,  Nuxt modules use lodash templates to create the plugin code.

Using the above mention two options, one could create a loop to generate the appropriate import code. This template gets built only one at the start of the nuxt build landbecomes static until the start of the next build.

This would be a better solution as webpack4 and terser can easily shake this static code plugin

pimlie:
> Using the above mention two options, one could create a loop to generate the appropriate import code. This template gets built only one at the start of the nuxt build landbecomes static until the start of the next build.
> 
> This would be a better solution as webpack4 and terser can easily shake this static code plugin

Could you please elaborate on this? Not sure I fully understand what you are trying to convey

tmorehouse:
Modules (which the nuxt plugin is) can generate a plugin code at build time.  Modules can accept options from nuxt.config.js, and pass the them to the template that builds the plugin code). Plugin templates are written as lodash templates. 

https://nuxtjs.org/guide/modules#provide-plugins

Basically the template would render a bunch of:

```js
import PluginName from 'bootstrap-vue/es/components/PluginName'
Vue.use(PluginName)
```

Once rendered, it will be included in nuxt build.

This imports just the list of plugins needed.

pimlie:
Thanks, maybe I still misunderstand but isnt that what this pr does?

Eg here is an example output I used with testing (this is the file which is saved in `.nuxt/bootstrap-vue.js`):
```js
import Vue from 'vue'

import Card from 'bootstrap-vue/es/components/card'
import FormInput from 'bootstrap-vue/es/components/form-input'
import InputGroup from 'bootstrap-vue/es/components/input-group'

Vue.use(Card)
Vue.use(FormInput)
Vue.use(InputGroup)
```

tmorehouse:
Also don't set the options name to `plugins` as directive plugin names conflict with conponent plugin names. Separate this into two option properties. `componentPlugins` and `directivePlugins`

Plugin names shouldn't be named the same as the component tag.

I.e. all layout related components made available by the layout plugin should be a plugin called `Layout`... not col or row etc.

pimlie:
The plugins in the example are not listed by the component tag but by the name of the plugin dir, which according to my deductions seems to be the kebab-case version of the plugin name (which in itself is the pascal-case version of the plugin dir). I already convert from/to both, so I think I can just update the example to include the plugin names.

Do you have a proposal what to do when somebody eg lists both `Popover` as component plugin as directive plugin? Should I prefix or ignore the directive (at that should already be loaded by the component I guess)?

tmorehouse:
That is why I am proposing changing the options definiton. Don't use

```js
BootstrapVue:
  plugins: [...]
}
```

Instead two properties
```is
BootstrapVue: {
  componentPlugins: [...],
  directivePlugins: [...]
}
```

And in the plugin template, iterate over the two arrays to produce the import code. This gets around duplicate plugin names between directives and components)

pimlie:
Sorry for the misunderstanding, but I was talking about the names in `plugin.template.js`. I have now added a prefix `v` to directivePlugin imports so they dont conflict with componentPlugin imports. But as mentioned in my previous comment I think we should also be able to just ignore directivePlugins when a componentPlugin with the same name exists.

Let me know if you want me to change something else

tmorehouse:
Plugins should be `PascalCase` without a `B` in front of them. I would still prefer _different_ config properties for component plugins vs directive plugins. The current naming convention used is confusing with the component names and directive names.

docs/markdown/intro/README.md
```diff
@@ -114,6 +114,27 @@ In your app main entry point include the single custom SCSS file (when using `sa
...
+If you wish to reduce your bundle size because you only use a subset of the available
+BootstrapVue `plugins`, you can configure the list of BootstrapVue `plugins` you want
+to globally install in your Nuxt.js project.
+
+
+```js
+{
+  modules: ['bootstrap-vue/nuxt'],
+  bootstrapVue: {
+    plugins: [
```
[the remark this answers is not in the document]
pimlie: I think you reviewed outdated code by accident cause I already split it: https://github.com/bootstrap-vue/bootstrap-vue/pull/2654/files#diff-8a4b8394f503f644ca14e802018a9cadR127

jackmu95:
@tmorehouse Is this ready to merge?

nuxt/index.js
tmorehouse: I'm think it is not needed, as it should be using the `es` files by default, and not `src`.  But we can check with @pi0 to be sure.

nuxt/index.js
tmorehouse: Yeah, I think that is correct, as the config options are defined at built time (and should not change after build)

nuxt/index.js
pi0: Sometimes users may want to directly include a specific component. This just instructs nuxt to transpile if such pattern **seen**. So is safe to be here.

nuxt/index.js
jackmu95: @pi0 @tmorehouse: Shouldn't we use the `lodash` utils here or is this JS going into a browser bundle?

tmorehouse: It is only ever run server side during nuxt build

pimlie: Using lodash is against the efforts that @pi0 put in #2523 

lodash has a lot of extra code with regards to unicode chars which you dont need here, as we know that plugin names are ascii chars only so lodash seemed not the correct choice to me. Also there didnt seem to be other use cases in bootstrap-vue for these conversions so thats why I choice not to create more dependency trees and keep the plugin as flat as possible as the nuxt plugin is already something of an oddball in this repo :)

kstadler:
Does Tree Shaking also support adding subcomponents (files instead of folders) to the whitelist?

tmorehouse:
The tree shaking option for the Nuxt plugin module allows you to specify which component group plugins to use (I.e. Layout, which includes b-container, b-row, b-col, etc). This allows you to while list which plugins are imported.

kstadler:
I've implemented the use of the tree shaking feature on component group level now and the size of the javascript was almost halved.
I suppose it would go down even more if it was also possible to include only single comonents instead of whole component groups. Would such a feature be possible?
Also, are there any references on how to cut down the size of the bootstrap css file?

pimlie:
As the author of this PR, I looked into that but supporting to include separate components is imo not worth the effort for the masses as it differs too much from the current bootstrap-vue exports. Also it wouldnt help that much, most component plugins already only consists of components you require anyway (eg breadcrumb, normally you need all those components anyway). Some plugins like nav and card could maybe benefit from this, but I dont think the savings are worth the effort or the additional complexity atm.

tmorehouse:
If one needs finer granularity with Nuxt, you can just import the individual components where you need them (i,e, in your page or custom components).

i.e. to just import `b-button`, don't use the Nuxt Plugin, but rather just do the following in you page:

```html
<!-- some page .vue file -->
<tempalate>
  <div>
     <p>Hello World</p>
     <b-button>Click Me</b-button>
</template>

<script>
import BButton from `bootstrap-vue/es/components/button/button'
export default {
  components: {
    BButton
  },
  methods: {
    @onClick() {
      // do something
    }
  }
}
</script>
```

Or use the required components in your main app entry or a custom plugin to make them available to all pages:

```js
import Vue from 'vue'
import BButton from `bootstrap-vue/es/components/button/button'
Vue.component('b-button', BButton)
```
Don't for get to import the bootstrap Vue CSS into your project.

kstadler:
Thanks for your suggestion, I'll take a look at it!

PR #2802, opened 2019-03-08
feat(table): option to disable footer sorting

iliveoncaffiene:

Option to disable sorting on the footer columns.

In all my use cases of the footer I am aggregating columns and never need to display column headers (but they are default shown with `:foot-clone="true"`). I'm OK with having to specify blank footer cells to clear out the copied headers - but in doing so I still have the sort arrows and "clickability" in those cells (which is not desired).

Idea: Maybe instead it would be a good idea to create an option that just enables the footer (i.e. `footer-enabled`) rather than requiring cloning the headers. If that's more desirable, it would eliminate the need for this PR

## PR checklist:

- [ ] Bugfix
- [x] Feature
- [ ] ARIA accessibility
- [x] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

- [x] Includes documentation updates (including updating the component's `package.json` for slot and
      event changes)

description above includes:**


iliveoncaffiene:
Also leaving the question out there what I should be doing about the headClicked processing. Should we pass through an "is footer" value? Pass through a "enable sorting" value (since no longer determined by just the field)? Or are we OK with the side-effect that if this option is enabled, we won't trigger head-clicked events in the footer.

src/components/table/table.js
```diff
@@ -1177,27 +1181,28 @@ export default {
           // We need to add a hint about what the column is about for non-dighted users
           ariaLabel = startCase(field.key)
         }
-        const ariaLabelSorting = field.sortable
+        const sortable = field.sortable && (!isFoot || !this.noFooterSorting)
```
tmorehouse: This could be simply written as:

```js
const sortable = field.sortable && !(isFoot && this.noFooterSorting)
```

Which is a bit more clear with it's intention

src/components/table/table.js
```diff
@@ -1207,9 +1212,15 @@ export default {
           },
           on: {
             click: evt => {
+              if (!sortable) {
+                return
+              }
               this.headClicked(evt, field)
```
tmorehouse: Head Clicked is used for more than just sorting...

Which is why the tests are failing.

You may need to pass `sortable` as the third argument to the `this.headClicked(...)` call, and then put some logic around the sorting routines in the handler (using the `sortable` argument passed)

tmorehouse:
Maybe passing the `isFoot` variable as well as `sortable` to the headClicked hanlder would be a good idea, and then in the emitted `head-click` event, we could pass the isFoot as the last argument to `$emit`.

This will also need documentation updates (both mentioning in the `README.md` and updating the component's event meta data in it's `package.json` file.

src/components/table/table.js
```diff
@@ -1026,22 +1030,24 @@ export default {
...
       }
-      if (field.sortable) {
-        if (field.key === this.localSortBy) {
-          // Change sorting direction on current column
-          this.localSortDesc = !this.localSortDesc
-        } else {
-          // Start sorting this column ascending
-          this.localSortBy = field.key
+      if (sortable) {
+        if (field.sortable) {
```
tmorehouse: This second `if` isn't really needed since it is determine in the render function.  so just checking to see if `sortable` is true is enough.

tmorehouse: Although....

What is happening now:

```
table › sortable header th should emit a sort-changed event with context when clicked and sort changed

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      [{"apiUrl": "", "currentPage": 1, "filter": "", "perPage": 5, "sortBy": "age", "sortDesc": false}]
    But it was not called.

      532 |           } else {
      533 |             if (sortBy) {
    > 534 |               expect(spy).toHaveBeenCalledWith(vm.context)
          |                           ^
      535 |               expect(vm.context.sortBy).toBe(null)
      536 |               sortBy = null
      537 |             } else {
```

Is because clicking on a non-sortable header cell (when sorting is enabled on the table), the cell should emit a `sort-changed` event in order to _clear_ the sort.

So right now, the `sortable` check prevents the header cell from emitting the `sort-changed` clear event.

Will need a bit of new logic to ensure that on hte header, the `sort-changed` event is always triggered (when the table has sorting enabled), and in the footer, only if fotter sorting is enabled, then the non-sortable `tfoot > th`'s should emit the sort changed.

tmorehouse: Fixed... tests are now passing.

tmorehouse:
Hmmm it appears that sorting is not working at all at the moment.

PR #2825, opened 2019-03-12
feat(utils): es6ify functions wherever possible

jackmu95:

This PR es6ifies function calls in the utils wherever possible.

## PR checklist:

- [ ] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [x] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


src/utils/bv-event.class.js
```diff
@@ -26,7 +26,7 @@ export default class BvEvent {
     // Create a private variable using closure scoping.
     let defaultPrevented = false
     // Recreate preventDefault method. One way setter.
-    this.preventDefault = function preventDefault() {
+    this.preventDefault = function() {
```
tmorehouse: The one thing about giving functions names, is that the name will appear in stack traces, rather than appearing as `anonymous function`.  Although that is mainly handy for larger inline functions.

src/utils/copy-props.js
```diff
@@ -3,14 +3,14 @@ import { isObject } from './object'
...
- * are also cloned as new references to prevent possible mution of original
+ * Copies props from one array/object to a new array/object. Prop values
+ * are also cloned as new references to prevent possible mutation of original
  * prop object values. Optionally accepts a function to transform the prop name.
  *
  * @param {[]|{}} props
  * @param {Function} transformFn
  */
-export default function copyProps(props, transformFn = identity) {
+export default (props, transformFn = identity) => {
```
tmorehouse: Similar for util functions, it is handy for debugging to have function names appear in stack traces

src/utils/html.js
```diff
@@ -1,9 +1,5 @@
...
 
-export function stripTags(text = '') {
-  return String(text).replace(stripTagsRegex, '')
-}
+export const stripTags = (text = '') => String(text).replace(stripTagsRegex, '')
 
-export function htmlOrText(innerHTML, textContent) {
-  return innerHTML ? { innerHTML } : { textContent }
-}
+export const htmlOrText = (innerHTML, textContent) => (innerHTML ? { innerHTML } : { textContent })
```
tmorehouse: This helper I would like to tweak a bit, as it kind of assumes that on or the other argument is present, but doesn't downgrade nicely when the second argument is undefined (which can/will be handy in some situations where we only want domProps for innerHTML and not innerText)

src/utils/copy-props.js
tmorehouse: When ES6 is transpiled into ES5:

```js
const someVar = (a,b,c) => {
  // some stuff
}
// becomes (sorta...)
someVar = function /* anonymous */ (a, b, c) {
  // some stuff
}.bind(undefined) // or what ever the `this` context of the parent is
```

So any stack trace from an exception will never show a function name, where as providing a named function can be really handy in a debugging sense.

I'm not against ES6 function shorthands... but they do have their downfalls.

jackmu95: Yeah. Code size is another thing we should take a look at.

src/utils/html.js
jackmu95: Ok, feel free to apply you changes to this PR!

tmorehouse: Tweaked... it now fallbacks to an empty object if both `innerHTML` and `textContent` are falsey.

src/utils/copy-props.js
jackmu95: What way do we go about this now?

tmorehouse: For most of the one-liners (or ones with few lines), I think the ES6 shorthand is fine.

For ones that do something more complex (where there is a higher chance something could go wrong), I think maybe named functions might be useful.

PR #2826, opened 2019-03-12
feat: add BOOTSTRAP_VUE_NO_WARN environment variable to hide warnings

gitlab-winnie:

This uses an environment variable `BOOTSTRAP_VUE_NO_WARN` to hide warnings logged by Boostrap Vue. This is useful in a test environment when incomplete test setup leads to false positives or in a production environment.

A second iteration of this would be to add a config option similar to [`Vue.config.warnHandler`](https://vuejs.org/v2/api/#warnHandler) to allow making tests fail.

## PR checklist:

- [ ] Bugfix
- [x] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


gitlab-winnie:
I'm happy to provide documentation and tests (if wanted) in case this is an acceptable way to solve the problem.

In case there are other ways to provide a global config, I'm happy to hear and change this approach.

src/utils/warn.js
```diff
@@ -4,6 +4,13 @@
  */
 /* istanbul ignore next */
 function warn(message) {
+  // eslint-disable-next-line no-undef
+  if (VueBootstrap && VueBootstrap.warnHandler) {
```
tmorehouse: The proper name is `BoostrapVue`

And this assumes that the name is global (i.e. assigned as `window.BootstrapVue`).

It would be better to pass this as an option when `Vue.use(BootstrapVue, { /*options object */})`

And have the plugin check for the second argument when instantiating (it would need to be defined as a property on the main install routine and bound to `window` or `global`, depending on the environment it is running in (i.e. client or SSR)

tmorehouse: Another option would be to optionally compile a no-op function depending on DEV or PROD mode of the final project.

gitlab-winnie: @tmorehouse Thank you for the amazingly fast review here! 🚀 

> The proper name is BoostrapVue

I found `VueBootstrap.warnHandler` closer to `Vue.config.warnHandler` but also don't mind reversing it to match the project name. 😃 

> It would be better to be able to optionally pass this as an option via Vue.use(BootstrapVue, { /*options object */})

Totally agree—but it does not work for the case where you don't register BootstrapVue components globally, correct?

> depending on DEV or PROD mode of the final project.

you mean like

```javascript
if (process.env.NODE_ENV === 'development') {
  console.warn(`[BootstrapVue warn]: ${message}`)
}
```

?

That still wouldn't allow failing tests for warnings but would be a very helpful first step. 👍

tmorehouse: The other option would be maybe a minification process... if in the minified version, `no-op` the method out, and in the non-minified version it would warn to the console. (i.e. setting global vars for terser: https://github.com/terser-js/terser#conditional-compilation during minification to remove the call to console.warn, without removing the warn utility)

But is most cases, the user should be fixing any issues that cause errors to be logged out to the console, rather than hiding them ;-)

gitlab-winnie: > But is most cases, the user should be fixing any issues that cause errors to be logged out to the console, rather than hiding them ;-)

Completely agree—in our case however we have more than 300 of these warnings and fixing them will take a while. In the meantime, I'd appreciate less output when running tests. 😃

gitlab-winnie: Would a

```javascript
if (process.env.BOOTSTRAP_VUE_NO_WARN) {
  return;
}

console.warn(`[BootstrapVue warn]: ${message}`)
```

work for you?

tmorehouse: Assuming that it can be handled well if `process && process.env` are tested for existence first, as not all users use a bundler to compile their projects (i.e. users using script tags to load BootstrapVue into their browser page).

tmorehouse: And warning are there for a reason... not correcting them may cause unpredictable results.

gitlab-winnie: I have updated this pull request now. I will squash the commits and add docs tomorrow.

> And warning are there for a reason... not correcting them may cause unpredictable results.

Thanks for your concerns! ❤️ As far as I can tell, the warnings only happen in our test environment (due to not waiting for components to be properly setup). While reducing the confidence in our test suite, at least there seem to be no consequences for our users.

In any case: my wish was actually to be able to fail our newly written tests in case warnings happen, so they cannot be ignored. Maybe can get that implemented in a follow-up?

Alternatively we could introduce a `BOOSTRAP_VUE_WARN=ignore` / `BOOSTRAP_VUE_WARN=throw` to solve the problem:

```javascript
const { BOOSTRAP_VUE_WARN } = process && process.env || {};
if (BOOSTRAP_VUE_WARN === 'ignore') {
  return;
}

if (BOOSTRAP_VUE_WARN === 'throw') {
  throw new Error(message);
}

console.warn(`[BootstrapVue warn]: ${message}`)

Comment (tmorehouse) [reply src/utils/warn.js]: Throw might be a bit extreme in some cases ;-) and the stack trace might not be overly helpful due to some anonymous function.

Comment (gitlab-winnie) [reply src/utils/warn.js]: Not sure what happened with my comment here. 🤔 

Let's stick with hiding the warnings then. I will write some docs later today and try to come up with tests.

**Comment by gitlab-winnie** [comment]
@tmorehouse I have squashed the commits now and added tests. I'm not quite sure where the docs would go though. Do you have a suggestion?

File: src/utils/warn.spec.js [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+    it('does not call console.warn()', () => {
+      warn(dummyWarning)
+
+      expect(console.warn).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('without BOOTSTRAP_VUE_NO_WARN environment variable set', () => {
+    beforeAll(() => {
+      global.process = null
```
gitlab-winnie: This seems quite unfriendly. Do we want

```suggestion
      process.env.BOOTSTRAP_VUE_NO_WARN = undefined
```

instead?

tmorehouse: This would assume `process.env` exists (assigning to a nested property on an object that might not exist).

I would also switch to `beforeEach`, and create an `afterEach` to restore the value to what ever it was originally.

```js
describe('...', () => {
  process.env = process.env || {}
  oldEnv = { ...process.env }
  beforeEach(() => {
    // Alter the environment
    process.env.BOOTSTRAP_VUE_NO_WARN = undefiend
  })
  afterEach(() =>
    // Restore the environment
    process.env = oldEnv
  })
  it('...', async () => {
    // test here
  })
})
```

And if you need to test with both set and unset environment, create two describe sections, each with their own beforeEach and afterEach sections

gitlab-winnie: > and create an afterEach to restore the value to what ever it was originally.

Makes sense! 👍 I'll restore the complete `process` object though (because restoring `process.env` would fail after `process = null`).

tmorehouse: Note, you will need to do a deep clone, as nested objects in process would be affected.

gitlab-winnie: I'm mocking the process object completely, so I don't think we need to clone the original one. I have just pushed my changes, so please let me know if you disagree. 😃

gitlab-winnie: > I don't think we need to clone the original one

...in addition as far as I know, the `process` object is not JSON serializable, so deep cloning would be difficult.

src/utils/warn.js
```diff
@@ -3,11 +3,10 @@
  * @param {string} message
  */
 /* istanbul ignore next */
-function warn(message) {
+const warn = message => {
   if (process && process.env && process.env.BOOTSTRAP_VUE_NO_WARN) {
     return
   }
```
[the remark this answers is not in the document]
jackmu95: Sounds like a good idea 👍

tmorehouse:
@gitlab-winnie I think this is ready for merging.

I'm not sure where this should be documented.  I think maybe a new reference section page on miscellaneous settings?

tmorehouse:
Added some rudimentary documentation.

src/utils/env.js
```diff
@@ -8,3 +10,7 @@ export const hasTouchSupport =
   inBrowser && ('ontouchstart' in document.documentElement || navigator.maxTouchPoints > 0)
 
 export const hasPointerEvent = inBrowser && Boolean(window.PointerEvent || window.MSPointerEvent)
+
+// Getters
+
+export const getNoWarn = () => process && process.env && process.env.BOOTSTRAP_VUE_NO_WARN
```
jackmu95: `getNoWarn()` seems a bit odd to me. What about `shouldWarn()`?

```js
export const shouldWarn = () => process && process.env && process.env.BOOTSTRAP_VUE_NO_WARN !== true
```

tmorehouse: That would be a good option...

tmorehouse: The only thing is... `get` implies a getter function, where `shouldWarn` kind of implies a constant (like all the other constants above it in `env.js`, and future coders might miss actually calling shouldWarn() by omitting the parens)

jackmu95: Yeah you are right about the `get` prefix.
Since `getShouldWarn()` isn't that better than `getNoWarn()` I'am also fine with that.

gitlab-winnie:
Thank you @tmorehouse and @jackmu95 for completing this! ❤️

PR #2844, opened 2019-03-14
fix(carousel): fix touchmove handler to re-enable swipe gestures

Anaphase:

In the carousel component, there's an [event handler](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/carousel/carousel.js#L403) that references `event.originalEvent`, which appears to always be undefined in several (all?) mobile browsers. Since [line 405](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/carousel/carousel.js#L405) isn't checking for the existence of `originalEvent` first, an error is thrown for every touch.

Maybe someone more familiar with the codebase can explain why `event.originalEvent` is being used here? It seems like `event.originalEvent` is a jQuery-specific thing, so maybe it's just a hold over from some Bootstrap code? Either way, I tried to make this PR as least destructive as possible - so if `event.originalEvent` exists, it'll be used but it will fall back to just using `event.touches`. Maybe this is unnecessary and `event.originalEvent` should just be removed all together?

To replicate this issue, simply use a touch-supporting device to load [a page with a carousel component](https://bootstrap-vue.js.org/docs/components/carousel/) and tap anywhere on the carousel. Note the errors thrown in the console. I see this behavior in Mobile Safari 12, Chrome Mobile iOS 72.0.3626 & 73.0.3683, and FireFox 65 Android.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

### DIFF
--- a/src/components/carousel/carousel.js
+++ b/src/components/carousel/carousel.js
@@ -402,7 +402,7 @@ export default {
     },
     touchMove(evt) /* istanbul ignore next: JSDOM doesn't support touch events */ {
       // ensure swiping with one touch and not pinching
-      if (evt.touches && (evt.originalEvent || evt).touches.length > 1) {
+      if (evt.touches && evt.touches.length > 1) {
         this.touchDeltaX = 0
       } else {
         this.touchDeltaX = evt.touches[0].clientX - this.touchStartX
--- a/src/components/carousel/carousel.js
+++ b/src/components/carousel/carousel.js
@@ -402,7 +402,7 @@ export default {
     },
     touchMove(evt) /* istanbul ignore next: JSDOM doesn't support touch events */ {
       // ensure swiping with one touch and not pinching
-      if (evt.touches && evt.originalEvent.touches.length > 1) {
+      if (evt.touches && (evt.originalEvent || evt).touches.length > 1) {
         this.touchDeltaX = 0
       } else {
         this.touchDeltaX = evt.touches[0].clientX - this.touchStartX


Anaphase:
NOTE: it appears as if this bug is actually preventing swipes from working at all, so maybe this PR title should change from:

```
fix(carousel): prevents touchmove handler from throwing errors when originalEvent is undefined
```

to

```
fix(carousel): fixes touchmove handler to re-enable swipe gestures
```

Thoughts?

jackmu95:
@Anaphase Yeah, please change the title.

The `touchMove()` method seems to be a port of the code Bootstrap itself uses here:
https://github.com/twbs/bootstrap/blob/9ba80d31d479cdd432e47586f35d0be0149da8bf/js/src/carousel.js#L289

We can drop the `.originalEvent` handling at all and work with the `evt` itself.

src/components/carousel/carousel.js
```diff
@@ -402,7 +402,7 @@ export default {
     },
     touchMove(evt) /* istanbul ignore next: JSDOM doesn't support touch events */ {
       // ensure swiping with one touch and not pinching
-      if (evt.touches && evt.originalEvent.touches.length > 1) {
+      if (evt.touches && (evt.originalEvent || evt).touches.length > 1) {
```
jackmu95: We can just use `if (evt.touches && evt.touches.length > 1) {`

tmorehouse: Yeah, this is a carryover from the Boostrap Without jQuery code (as they also have customized events in the non-jQuery version).

Anaphase: @tmorehouse Cool, that's kinda what I figured. Thanks!

jackmu95:
Thanks @Anaphase!

PR #2867, opened 2019-03-18
feat(table): add TypeScript definitions for table fields

mattgrande:

I had basically already created these TS definitions within my own project, before realising I should probably create this as a PR.

I'm not super familiar with TS + Vue, so if there's a better way to declare what I'm doing on line 142 for Vue Components, please let me know.

## PR checklist:

- [ ] Bugfix
- [x] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)
- [x] ARIA Accessibility has been taken into consideration (does it affect screen reader users or

description above includes:**


src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
 
   export default VuePlugin
 
+  type VueVariant = 'active' | 'success' | 'info' | 'warning' | 'danger'
```
tmorehouse: There are actually quite a few more variants, depending on where they are used, such as `primary`, `secondary`, `light` and `dark`... and people can defined their own variants in CSS/SCSS as well.

See the Bootstrap Standard variants at: https://bootstrap-vue.js.org/docs/reference/color-variants

So I am not sure if restricting the values the the above 5 types is goo.. (as well, `active` is not an official variant, as it is a class unto it's own)

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
...
+    headerTitle?: string
+    headerAbbr?: string
+    class?: string | string[]
+    formatter?: string | FormatterCallback
+    sortable: boolean
+    sortDirection: SortDirection
+    tdClass: string | string[] | FormatterCallback
+    thClass: string | string[]
+    thStyle: mixed
+    variant: VueVariant
```
tmorehouse: Different components have different variants available to them... so I don't think this type should be called `VueVariant`... but maybe `TableVariant` (and even then, people can defined their own variants to use, so maybe this should be `string`, or add the ability to use the defined variants, plus a catch all string to say it could be custom)

src/index.d.ts
```diff
@@ -115,6 +139,7 @@ declare module 'bootstrap-vue' {
   export interface Progress extends Vue {}
   export interface Table extends Vue {
     refresh: () => void
+    fields: TableFieldObject | TableFieldArray
```
tmorehouse: Maybe: `fields: TableFieldObject | TableFieldArray | string` ?

Although the fields definition can be a mix of string entries and object entries... does the `TableFieldArray` type handle mixed value arrays (i.e. strings and objects)?

tmorehouse: Also, we discourage the use of an object being used to describe the fields, as field order cannot be guaranteed, and may cause SSR rehydration issues on some browsers.

tmorehouse: Fields can also be `null`, which means that `b-table` will sample the first row of data to generate the field names.

So maybe it should be `fields?: TableFieldObject | TableFieldArray | string[]` ?

iliveoncaffiene: The way I do it is:
```ts
type TableFieldObject = { key: string } | string; // expand out the object - just here for example
type TableFieldArray = Array<TableFieldObject>;
```

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
...
+    thClass: string | string[]
+    thStyle: mixed
+    variant: VueVariant
+    tdAttr: mixed | ((value: mixed, key: string, item: mixed) => mixed)
+    isRowHeader: boolean
+  }
+  export interface TableFieldObject {
+    [key: string]: TableField
+  }
+  export interface TableFieldArray extends Array<string | {key: string & TableField}> {}
```
iliveoncaffiene: `{key: string & TableField}` implies that `key` itself needs to be an object
Shouldn't this be:
`({key: string} & TableField)`

src/index.d.ts
mattgrande: According to the docs for `variant`, here: https://bootstrap-vue.js.org/docs/components/table#field-definition-reference

> Apply contextual class to all the `<th>` and `<td>` in the column - `active`, `success`, `info`, `warning`, `danger`. These variants map to classes thead-${variant} (in the header), table-${variant} (in the body), or bg-${variant} (when table prop dark is set).

So, is the documentation here incorrect? If that's the case, we should update the docs as well (which I'm willing to do in a further PR).

mattgrande: Oh, I just read your comment below... Perhaps I'll rename this to `TableVariant`, and define `variant` as `variant: TableVariant | string`. Would that solve our issue?

src/index.d.ts
mattgrande: `TableFieldArray` is defined as `Array<string | {key: string & TableField}>` to handle arrays of strings or objects (since, as you said, they can be mixed).

If you're discouraging objects being used, I'm willing to remove it, but then it comes down to a question of are we typing all supported options, or just the ones we prefer?

I will also add handling for `fields` to be `null`.

src/index.d.ts
mattgrande: Correct, thank you for that!

src/index.d.ts
tmorehouse: Ah, right `table-active` is an actual variant.  The variant reference docs should be updated to include the extra `active` variant.  And maybe the table docs should have a link back to the variant docs.

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
 
   export default VuePlugin
 
+  type TableVariant = 'active' | 'success' | 'info' | 'warning' | 'danger'
+  type SortDirection = 'asc' | 'desc'
+  type FormatterCallback = ((value: mixed, key: string, item: mixed) => string)
```
iliveoncaffiene: @tmorehouse I know of instances where I use the formatter and return values that are not strings (which are converted to strings)
Is it wrong to force this to be a string return rather than `mixed` (or `any`)?

tmorehouse: Hmmm... yeah, that might be good. Not sure of the difference between `mixed` or `any`

src/index.d.ts
tmorehouse: and `variant: TableVariant | string` would be good.

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
...
+    headerTitle?: string
+    headerAbbr?: string
+    class?: string | string[]
+    formatter?: string | FormatterCallback
+    sortable: boolean
+    sortDirection: SortDirection
+    tdClass: string | string[] | FormatterCallback
+    thClass: string | string[]
+    thStyle: mixed
+    variant: TableVariant
```
tmorehouse: Maybe update to the following:

```suggestion
    variant: TableVariant | string
```

src/index.d.ts
tmorehouse: Probably `any` is safer than `mixed`, as it has a looser typing than `mixed`.

src/index.d.ts
```diff
@@ -115,6 +139,7 @@ declare module 'bootstrap-vue' {
   export interface Progress extends Vue {}
   export interface Table extends Vue {
     refresh: () => void
+    fields?: TableFieldObject | TableFieldArray
```
tmorehouse: Should `fields` be up in `export interface TableField {`? and we are missing `items`, which could be `array[object]` | function`, where function is a provider function?

iliveoncaffiene: Not to jump in but, the first request doesn't make sense: `fields` is a property on a table is it not? Not a property on the "field" itself.

In general, I don't see the use case for having anything other than methods that might be called externally via a reference. In 99% of uses you're going to be managing your fields / items (or provider method) passed into this component, and shouldn't need to reference them from the table - they should instead be modified/referenced via whatever is consuming the `b-table` component.
Ex:
```html
<b-table :items="myItems" :fields="myFields" ref="table1">....</b-table>
```
The code that this is in will have `myItems` and `myFields`... in what cases would I need to do:
`(this.$refs.table1 as Table).fields` or `(this.$refs.table1 as Table).items`? Other than reading them, modifying them goes against Vue conventions and recommended practices

tmorehouse: Ah, right... I was thinking it was the table interface definition, not the fields definition (jumping back and forth between too many things at the moment)

mattgrande:
I think I've taken care of everything mentioned above; Please let me know if I've missed anything!

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
 
   export default VuePlugin
 
+  type TableVariant = 'active' | 'success' | 'info' | 'warning' | 'danger' | 'primary' | 'secondary' | 'light' | 'dark'
+  type SortDirection = 'asc' | 'desc'
```
tmorehouse: Actually this one supports three values:

```js
type SortDirection = 'asc' | 'desc' | 'last'
```

src/index.d.ts
```diff
@@ -14,6 +14,30 @@ declare module 'bootstrap-vue' {
...
+  type FormatterCallback = ((value: mixed, key: string, item: mixed) => any)
+
+  export interface TableField {
+    label?: string
+    headerTitle?: string
+    headerAbbr?: string
+    class?: string | string[]
+    formatter?: string | FormatterCallback
+    sortable: boolean
```
tmorehouse: Should all of these have `?:`

src/index.d.ts
tmorehouse: So why is `fields` mentioned here, but `items` is not? should we define all the props, or is that just overkill?

src/index.d.ts
mattgrande: Hmm, most likely. Will edit.

src/index.d.ts
mattgrande: I just added `fields` because it's what I already had right now, and I thought it should go upstream.

tmorehouse: OK...  at some point in the future we should define the other props.

jackmu95:
Thanks @mattgrande!

mattgrande:
Thanks! :tada:

ldawkes:
@mattgrande What version of TypeScript are you using? Because as far as I'm aware TypeScript doesn't have the `mixed` keyword, so TypeScript 3.4 now fails the type check because this change introduces an invalid keyword.

You should use `any` if you're going to imply a mix of types that can be returned.

PR #3012, opened 2019-04-04
fix(form-textarea): improved computedHeight calculation when in auto resize mode

Istador:

`computedHeight` is used, when `max-rows` is given to an `b-form-textarea` to calculate the auto height.

**1. fix: not subtracting border width from scrollHeight**
`scrollHeight` contains only the padding and not the border.
This resulted in an incorrect calculation of `contentRows` (float instead of int) and lead to an visible vertical scroll bar, even when `max-rows` haven't been reached yet.

~**2. fix: respecting boxSizing in height calculation**~
~Depending on boxSizing, the height of the element contains the padding and border (`box-sizing: border-box`) or not (`box-sizing: content-box`).~ **EDIT (by TM):** _Bootstrap CSS always sets every element to `border-box`, so this extra test would only generate overhead that is not needed_.

~**3. fix: setting height temporary to `minHeight` instead of `'auto'`**~
~Setting it to `auto` might result in a height that is bigger than 2 content rows.
This is bad because when the content height is lower than this height, then `scrollHeight` is affected by this. It effectively results in the `auto`-height being the `minHeight`, therefore the textarea is bigger than it should be.~

---

I only tested chrome, firefox and waterfox on desktop. Because one of the lines I changed was marked as "browser dependant" it should better be tested thoroughly - especially on mobile browsers.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**

### DIFF
--- a/src/components/form-textarea/form-textarea.js
+++ b/src/components/form-textarea/form-textarea.js
@@ -54,7 +54,7 @@ export default {
       const styles = {
         // Setting `noResize` to true will disable the ability for the user to
         // manually resize the textarea. We also disable when in auto resize mode
-        resize: !this.computedRows || this.noResize ? 'none' : null,
+        resize: !this.computedRows || this.noResize ? 'none' : null
       }
       if (!this.computedRows) {
         // The computed height for auto resize.
--- a/src/components/form-textarea/form-textarea.js
+++ b/src/components/form-textarea/form-textarea.js
@@ -54,7 +54,10 @@ export default {
       const styles = {
         // Setting `noResize` to true will disable the ability for the user to
         // manually resize the textarea. We also disable when in auto resize mode
-        resize: !this.computedRows || this.noResize ? 'none' : null
+        resize: !this.computedRows || this.noResize ? 'none' : null,
+        // We always add a vertical scrollbar to the textarea when auto-resize is
+        // enabled so that the computed height calcaultion returns a stable value.
+        'overflow-y': this.computedRows ? 'scroll' : null
       }
       if (!this.computedRows) {
         // The computed height for auto resize.
--- a/src/components/form-textarea/form-textarea.js
+++ b/src/components/form-textarea/form-textarea.js
@@ -92,7 +92,7 @@ export default {
       const el = this.$el
 
       // Element must be visible (not hidden) and in document
-      // *Must* be checked after above checks
+      // Must be checked after above checks
       if (!isVisible(el)) {
         return null
       }
@@ -101,17 +101,14 @@ export default {
       const computedStyle = getCS(el)
       // Height of one line of text in px
       const lineHeight = parseFloat(computedStyle.lineHeight)
-      // calculate height of border and padding
-      const border = (
+      // Calculate height of border and padding
+      const border =
         (parseFloat(computedStyle.borderTopWidth) || 0) +
         (parseFloat(computedStyle.borderBottomWidth) || 0)
-      )
-      const padding = (
-        (parseFloat(computedStyle.paddingTop) || 0) +
-        (parseFloat(computedStyle.paddingBottom) || 0)
-      )
-      // calculate offset
-      const offset = (computedStyle.boxSizing === 'border-box' ? border + padding : 0 )
+      const padding =
+        (parseFloat(computedStyle.paddingTop) || 0) + (parseFloat(computedStyle.paddingBottom) || 0)
+      // Calculate offset
+      const offset = computedStyle.boxSizing === 'border-box' ? border + padding : 0
       // Minimum height for min rows (browser dependant)
       const minHeight = lineHeight * this.computedMinRows + offset
 
@@ -128,10 +125,9 @@ export default {
       // Calculate the required height of the textarea including border and padding (in pixels)
       const height = Math.max(Math.ceil(rows * lineHeight + offset), minHeight)
 
-      // noAutoShrink
+      // Computed height remains the larger of oldHeight and new height
+      // When height is `sticky` (no-auto-shrink is true)
       if (this.noAutoShrink && (parseFloat(oldHeight) || 0) > height) {
-        // Computed height remains the larger of oldHeight and new height
-        // When height is `sticky` (no-auto-shrink is true)
         return oldHeight
       }
 
--- a/src/components/form-textarea/form-textarea.js
+++ b/src/components/form-textarea/form-textarea.js
@@ -5,7 +5,7 @@ import formStateMixin from '../../mixins/form-state'
 import formTextMixin from '../../mixins/form-text'
 import formSelectionMixin from '../../mixins/form-selection'
 import formValidityMixin from '../../mixins/form-validity'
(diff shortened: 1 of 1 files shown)


src/components/form-textarea/form-textarea.js
```diff
@@ -97,36 +97,39 @@ export default {
...
+        (parseFloat(computedStyle.borderBottomWidth) || 0)
+      )
+      const padding = (
         (parseFloat(computedStyle.paddingTop) || 0) +
         (parseFloat(computedStyle.paddingBottom) || 0)
-      // Calculate content height in "rows"
-      const contentRows = Math.max((el.scrollHeight - offset) / lineHeight, 2)
+      )
+      // calculate offset
+      const offset = (computedStyle.boxSizing === 'border-box' ? border + padding : 0 )
```
tmorehouse: Actually, Bootstrap CSS always has everything in `border-box` mode... so this test isn't really needed.

tmorehouse: https://github.com/twbs/bootstrap/blob/c6c9e322f7a30d380dbcbef77f31ced85df790c8/scss/_reboot.scss#L11-L23

```scss
// Document
//
// 1. Change from `box-sizing: content-box` so that `width` is not affected by `padding` or `border`.
// 2. Change the default font family in all browsers.
// 3. Correct the line height in all browsers.
// 4. Prevent adjustments of font size after orientation changes in IE on Windows Phone and in iOS.
// 5. Change the default tap highlight to be completely transparent in iOS.


*,
*::before,
*::after {
  box-sizing: border-box; // 1
}
```

src/components/form-textarea/form-textarea.js
```diff
@@ -97,36 +97,39 @@ export default {
...
+      // Minimum height for min rows (browser dependant)
+      const minHeight = lineHeight * this.computedMinRows + offset
+
+      // Probe scrollHeight by temporarily changing the height to the minimum
+      const oldHeight = el.style.height
+      el.style.height = minHeight + 'px'
```
jackmu95: `el.style.height = `${minHeight}px`

src/components/form-textarea/form-textarea.js
```diff
@@ -97,36 +97,39 @@ export default {
...
-        (parseFloat(computedStyle.borderBottomWidth) || 0) +
+        (parseFloat(computedStyle.borderBottomWidth) || 0)
+      )
+      const padding = (
         (parseFloat(computedStyle.paddingTop) || 0) +
         (parseFloat(computedStyle.paddingBottom) || 0)
-      // Calculate content height in "rows"
-      const contentRows = Math.max((el.scrollHeight - offset) / lineHeight, 2)
+      )
+      // calculate offset
```
jackmu95: `// Calculate offset`

src/components/form-textarea/form-textarea.js
```diff
@@ -97,36 +97,39 @@ export default {
...
-
       // Get current computed styles
       const computedStyle = getCS(el)
       // Height of one line of text in px
       const lineHeight = parseFloat(computedStyle.lineHeight)
-      // Minimum height for min rows (browser dependant)
-      const minHeight = parseInt(computedStyle.height, 10) || lineHeight * this.computedMinRows
-      // Calculate height of content
-      const offset =
+      // calculate height of border and padding
```
jackmu95: `// Calculate height of border and padding`

src/components/form-textarea/form-textarea.js
```diff
@@ -97,36 +97,39 @@ export default {
...
       const height = Math.max(Math.ceil(rows * lineHeight + offset), minHeight)
 
-      // Place old height back on element, just in case this computed prop returns the same value
-      el.style.height = oldHeight
-
-      // Value of previous height (without px units appended)
-      const oldHeightPx = parseFloat(oldHeight) || 0
-
-      if (this.noAutoShrink && oldHeightPx > height) {
+      // noAutoShrink
```
jackmu95: Please remove this comment, we explain exactly what's happening here in the if statement.

src/components/form-textarea/form-textarea.js
tmorehouse: As well, brackets are not needed around this computation

src/components/form-textarea/form-textarea.js
```diff
@@ -92,43 +92,42 @@ export default {
...
       // Element must be visible (not hidden) and in document
-      // *Must* be checked after above checks
+      // Must be checked after above checks
       if (!isVisible(el)) {
         return null
       }
 
-      // Remember old height (includes `px` units) and reset it temporarily to `auto`
-      const oldHeight = el.style.height
-      el.style.height = 'auto'
```
tmorehouse: Part of the reason for using `auto` was that it would trigger a reflow of the `<textarea>`.

tmorehouse:
@Istador do you have an example of where `#3`, in your original post above, is occurring?  What browser, and can you provide a minimum reproduction of that issue?

`auto` was used because of it automatically triggering a reflow calculation, where as setting it to a specific height may wait until the next animation frame occurs to update the returned scroll height. The next animation frame may happen after the routine pulls the `scrollHeight` from the `<textarea>`.

tmorehouse:
Here is an example using the new code from this PR, with min rows set to `2` (in Chrome 72):

image

And the old code with the same text typed in (Also on chrome 72):

image

Note the extra blanks line in the first example using the new code that sets to `minHeight`, where as the second example is using `auto` as the temporary height.

tmorehouse:
The real issue with the extra line (or lack of in some situations) is due to the effect the scroll bar has on the maximum width of the content.  i.e. when calculating the required height, there is no scrollbar taken into account.

By adding `overflow-y: scroll` (rather than the default of `overflow-y: auto`), the effect the scrollbar has on the content is always consistent (because the inner width does not change).

Istador:
> @Istador do you have an example of where `#3`, in your original post above, is occurring? What browser, and can you provide a minimum reproduction of that issue?
> 
> `auto` was used because of it automatically triggering a reflow calculation, where as setting it to a specific height may wait until the next animation frame occurs to update the new `scrollHeight`. The next animation frame may happen _well after_ the routine pulls the (possibly incorrect) `scrollHeight` from the `<textarea>`.

`#1`: Scrollbar present when not needed
https://jsfiddle.net/vthwn6gq/


`#2`: Too big at the end (the user might overwrite box-sizing to content-box for whatever reason):
https://jsfiddle.net/g7k6txoj/


`#3`: Too big if underfull:
https://jsfiddle.net/byjdxr8f/
Works in Chrome 73.0.3683.103 but not in Firefox 66.0.2 or 60.6.1esr

tmorehouse:
For `#2`, the user should never overwrite `box-sizing: border-box` as that would completely break bootstrap V4 layout which depends on sizes including the border, since form-controls rely on this accurate sizing (for widths of elements).

PR #3017, opened 2019-04-05
fix(modal): stacked modal z-index calculations (closes #3015)

### LINKED ISSUE #3015: Z-index of Modal
Modal (element `.modal`) is appended to `<div id="modal-1___BV_modal_outer_" style="position: absolute; z-index: 2000;">` after opening. So.. Modal with correct z-index value (scss: `$zindex-modal: 1050 !default;`) is inside container with `z-index: 2000` and it is problem for me (Modals cover my toast notifications).

Tested with
bootstrap-vue: 2.0.0-rc.16
bootstrap: 4.3.1

jackmu95:

This PR makes the `z-index` offset used for stacked modals from `2000` to `1040` which is the default from Boostrap for the modal backdrop.

Automatically detects if the modal backdrop z-index has been changed from the default.

Now uses an increment  value of 1 for each successively opened stacked modal, rather than using the previous value of 2000.

Introduced a helper module to handle keeping track of stacked modals (and their display order), which includes adjusting their z-index values as modals are opened and closed, as well as handles any adjustments needed to the document body and it's contents.

Closes #3015.

## PR checklist:

- [x] Bugfix
- [ ] Feature
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other, please describe:

- [ ] Yes
- [x] No

      from these messages.**

      event changes)

description above includes:**


src/utils/config.js
```diff
@@ -76,6 +76,8 @@ const DEFAULTS = {
     blankColor: 'transparent'
   },
   BModal: {
+    // Defaults to the Boostrap default for the modal backdrop
```
tmorehouse: I would slightly change the comment to
```
// Defaults to the same value as the bootstrap Backdrop z-index
```

As the backdrops, modals containers, etc, all still have their bootstrap defined z-indexes, ans this z-index is used only for stacking order (as modals can be in-document anywhere, in any order)

The stacker DIV just creates a _new_ stacking context (where the content of the div retains it's stacking definition within that container. the content inside is stacked relative to each other, and then relative to the other stacker divs.

Once we move to portal-vue for modals (so that they all appended at the bottom of the document in a single container when open), the staking z-index trickery will go away, as the modals will always be in the portal-target in the order they will be stacked.

docs/plugins/bootstrap-vue.js
```diff
@@ -1,4 +1,7 @@
 import Vue from 'vue'
 import BootstrapVue from '../../src'
 
-Vue.use(BootstrapVue)
+Vue.use(BootstrapVue, {
+  // Make sure modals in the docs are above anything else
+  BModal: { zIndexOffset: 2000 }
+})
```
tmorehouse: Any reason why we don't use the new default in the docs?

tmorehouse:
Originally, I had the first modal stacker use a z-index if 0 (i.e. the default for any element without a specific z-index), and the rest that were opened on top used the z-index offset calculation . to make sure the modal backdrop of the stacked modal was above the previously opened modal.  

But I think we ran into some issues and defaulted the first one to use offset as it's base z-index.

The issue problem of the modal being over the toasts will sill be there is more than one modal is open at a time.

One workaround it to increase the toast's default z-Index (in CSS) to be some rather high value like `999999999` which would ensure it would be placed over any stacked modals.

docs/plugins/bootstrap-vue.js
tmorehouse: If the new default doesn't work in the docs, then it may not work in production for others.

tmorehouse:
@jackmu95 I'm going to try setting the first opened modal to `z-index` of `0` and see if all works OK.

Tests may fail initially due to the checks for the default stacking z-index being applied.

tmorehouse:
After a bit of playing, I think we need two config values:

- zIndexBase (default to 1040)
- zIndexIncrement (default 2000)

The base would be the first modal opened z-index and the increment would be what to use for the next modal's z-index would be incremented by (added onto the initial base Value).

The first one would be used to make sure the modal is above any user elements that may have a high z-index (for example the docs top navbar), and then the increment would be used to make sure the next modal is opened above any previous modals.

The increment has to be higher than the highest z-index in the lower modals.

Issue #3126, opened 2019-04-17
<b-modal> @change event not triggering if the modal visible closed immediately


chenxeed:
### Describe the bug

There is a changed behavior in `<b-modal>` after the release of `v2.0.0-rc.12` and perhaps specifically because of [this PR](https://github.com/bootstrap-vue/bootstrap-vue/pull/2270/files#diff-2f6544a135bbefb71664d04f6e67d3d0R555).

When we need to directly hide the modal on `@change` event, it was working before. For example:

(note: You can try this script on the [modal example](https://bootstrap-vue.netlify.com/docs/components/moda)

```
<template>
  <div>
    <b-button @click="modalShow = true">Open Modal {{modalShow}}</b-button>
    <b-modal :visible="modalShow" @change="modalShow = false">Hello From Modal!</b-modal>
  </div>
</template>

<script>
  export default {
    data() {
      return {
        modalShow: false
      }
    }
  }
</script>
```

On first click the button, it will set the `modalShow` to `true` and directly back to `false`. On the second click and next after, it will set `modalShow` to `true` and never set back to `false` anymore.

By right, the `modalShow` should always be `false` whenever we click the button, because when the modal show and `@change` event triggered, it will directly set to `false`. This is working back in `v2.0.0-rc.11` before.

I found out the culprit is because on the 2nd click, the flag `is_closing` is still true therefore it prevents the show function from running further.

![Screen Shot 2019-04-17 at 2 47 57 PM](https://user-images.githubusercontent.com/3530355/56266883-df50b680-611f-11e9-8dda-794824f0f255.png)

### Steps to reproduce the bug

1. Copy paste the script I add above to the modal example
2. Click the "Open Modal" button.
3. On first click, the `showModal` flag will be set back to `false`
4. On second click, the `showModal` will be set to `true` and never be `false` anymore.

### Expected behavior

The flag `showModal` should be back to `false`, as the `@change` event should be triggered.

### Versions

**Libraries:**

- BootstrapVue: 2.0.0-rc.12
- Bootstrap: 4.3.1
- Vue: 2.6.0

**Environment:**

- Device: Mac
- OS: MacOS
- Browser: Chrome
- Version: 73

### Demo link

Try the script pasted above in the [bootstrap vue website modal example](https://bootstrap-vue.netlify.com/docs/components/modal)

tmorehouse:
It should be scheduling a new show() event once it receives a `hidden` event.

[...]

PR #3131, opened 2019-04-18
fix(modal): ensure that v-model is updated when show or hide is canceled

tmorehouse:
### Describe the PR

Addresses https://github.com/bootstrap-vue/bootstrap-vue/issues/3126#issuecomment-484347044

Specifically: When the v-model is used to open or hide the modal, and the corresponding show or hide event was cancelled, the v-model was not being reset.

### PR checklist

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -362,7 +362,7 @@ export default Vue.extend({
   },
   methods: {
     updateModel(val) {
-      if (val !== !!this.visible) {
+      if (val !== this.visible) {
         this.$emit('change', val)
       }
     },
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -361,6 +361,11 @@ export default Vue.extend({
     }
   },
   methods: {
+    updateModel(val) {
+      if (val !== !!this.visible) {
+        this.$emit('change', val)
+      }
+    },
     // Public Methods
     show() {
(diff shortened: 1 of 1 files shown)


src/components/modal/modal.js
```diff
@@ -361,6 +361,11 @@ export default Vue.extend({
     }
   },
   methods: {
+    updateModel(val) {
+      if (val !== !!this.visible) {
```
jackmu95: Why the `!!`? The `visible` prop should always be a boolean.

tmorehouse: "Should" be is the keyword, except when someone uses a non Boolean on the prop (and has dev mode off which won't warn the user)

tmorehouse: The other option was to use:
```js
if (val === !this.visible)
```

jackmu95: Hmm, but we expect the type at most other places... But I'am also fine with using it.

tmorehouse: I suppose we should probably emit it regardless of type... that way the model is updated correctly

PR #3170, opened 2019-04-23
feat(toast): add additional options to global config (closes #3169)

### LINKED ISSUE #3169: Setting custom <b-toast> defaults

When using toats on demand (using the ```this.$bvToast.toast()```), I always use my same custom behavior that is different from the defaults (such as position, title , duration...) but currently there's no way to set them initially and I have to re-set them in every call.

I would like to be able to make a call in my application loading code in order to set my deaults.

### Describe alternatives you've considered

If this is actually possible, I haven't found it in the documentation page.

### Additional context

This toast are soooo cool! :)

tmorehouse:
### Describe the PR

Adds additional values to global config for `b-toast`, `b-toaster` (and `this.$bvToast.toast()` as it uses the previous two).

Closes #3169

### PR checklist

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -32,7 +32,7 @@ export const props = {
   },
   toaster: {
     type: String,
-    default: () => String(getComponentConfig(NAME, 'toaster'))
+    default: () => getComponentConfig(NAME, 'toaster')
   },
   variant: {
     type: String,
@@ -159,6 +159,9 @@ export default Vue.extend({
       // Minimum supported duration is 1 second
       return Math.max(parseInt(this.autoHideDelay, 10) || 0, MIN_DURATION)
     },
+    computedToaster() {
+      return String(this.toaster)
+    },
     transitionHandlers() {
       return {
         beforeEnter: this.onBeforeEnter,
@@ -212,7 +215,7 @@ export default Vue.extend({
     })
     // Make sure we hide when toaster is destroyed
     this.listenOnRoot('bv::toaster::destroyed', toaster => {
-      if (toaster === this.toaster) {
+      if (toaster === this.computedToaster) {
         this.hide()
       }
     })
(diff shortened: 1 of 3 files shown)


src/components/toast/toast.js
```diff
@@ -32,7 +32,7 @@ export const props = {
   },
   variant: {
     type: String,
-    default: null
+    default: () => getComponentConfig(NAME, 'variant') || null
```
jackmu95: I think we should omit the fallbacks, since we don't do that for the other components.

tmorehouse: True...

src/components/toast/toast.js
```diff
@@ -30,14 +30,14 @@ export const props = {
     type: String,
     default: null
   },
-  variant: {
-    type: String,
-    default: null
-  },
   toaster: {
     type: String,
     default: () => getComponentConfig(NAME, 'toaster') || 'b-toaster-top-right'
```
jackmu95: Please remove that too.

tmorehouse: Although a toaster name is required... and users could set it to null or an empty string by mistake.

jackmu95: But it adds additional bundle size, which de did with the config stuff already enough.

tmorehouse: I could force it to a string

jackmu95: This will prevent the prop type warning for invalid value in DEV-mode.
Not sure about that change.

tmorehouse: I'm not sure the prop type check runs on the `default` value or not.

tmorehouse: But it will prevent PortalTarget from crashing ;-)

tmorehouse: Unless we add a computed prop that stringifies the name, and we pass that computed prop toPortal and PortalTarget

PR #3172, opened 2019-04-23
fix(form-input): properly handle out-of-sync values (closes #2657)

### LINKED ISSUE #2657: <b-form-input> becoming out of sync with value if using formatter
When a form-input has a formatter on it stripping out characters in the case of trying to only allowing numerial values in a text input, the input value can sometimes not match up with the emitted value.

codepen: https://codepen.io/Kalvens/pen/Lqabrj (type any nonnumerical character in the input)
browser: Chrome Version 69.0.3497.100 (Official Build) (64-bit)
OS: linux

I believe the issue comes from 
https://github.com/bootstrap-vue/bootstrap-vue/blob/a14245cc27751e395f34b90fba20db16869a3484/src/mixins/form-text.js#L107-L123

the value is already formatted at that point so if the formatter strips out all of the characters leaving the value to be an empty string, nothing will happen even though the input still has a value set for it

I was playing around with your source code and doing
```javascript
if (this.localValue !== value) {
    ...
} else if (this.$el.value !== value)  {
  this.$el.value = value;
}
```
seems to fix my issue, but I am not sure if that would be affecting anything else using the mixin.

jackmu95:
### Describe the PR

This PR adds an additional check to to the `updateValue()` method to check if the given value is different from the actual input value, even when the `localValue` hasn't changed.
This can be caused by the `formatter` function.

Closes #2657.

### PR checklist

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**

### DIFF
--- a/src/mixins/form-text.js
+++ b/src/mixins/form-text.js
@@ -103,7 +103,7 @@ export default {
     getFormatted(value, evt, force = false) {
       value = this.stringifyValue(value)
       if ((!this.lazyFormatter || force) && isFunction(this.formatter)) {
-        return this.formatter(value, evt)
+        value = this.formatter(value, evt)
       }
       return value
     },
--- a/src/mixins/form-text.js
+++ b/src/mixins/form-text.js
@@ -140,6 +140,7 @@ export default {
       // Exit when the `formatter` function strictly returned `false`
       // or prevented the input event
       if (formatted === false || evt.defaultPrevented) {
+        /* istanbul ignore next */
         evt.preventDefault()
         return
       }
@@ -157,6 +158,7 @@ export default {
       // Exit when the `formatter` function strictly returned `false`
       // or prevented the input event
       if (formatted === false || evt.defaultPrevented) {
+        /* istanbul ignore next */
         evt.preventDefault()
         return
       }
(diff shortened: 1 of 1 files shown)


src/mixins/form-text.js
```diff
@@ -100,38 +100,44 @@ export default {
...
-          // Emulate .trim modifier behaviour
+          // Emulate `.trim` modifier behaviour
           value = value.trim()
         }
         // Update the v-model
         this.$emit('update', value)
+      } else if (value !== this.$refs.input.value) {
+        // When the `localValue` hasn't changed but the actual input value
+        // is out of sync, make sure to change it to the given one
+        this.$refs.input.value = value
```
tmorehouse: One thing to note about this is that it can cause the user's cursor/insert point to jump.  Although in general, formatters will have that effect as well.

jackmu95: Yeah, I think we can/have to live with that.

tmorehouse: May want to add an `/*istanbul ignore next */` above the line, as it will be difficult to test this scenario

PR #3229, opened 2019-05-02
feat(config): Add option in config to set global tooltip and popover boundary

AlanRezende:
### Describe the PR

A clear and concise description of what the pull request does.

### PR checklist

- [ ] Bugfix
- [x] Feature
- [X] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [X] No
- [ ] Yes (please describe)

- [X] It's submitted to the `dev` branch, **not** the `master` branch
- [X] It should address only one issue or feature. If adding multiple features or fixing a bug and adding a new feature, break them into separate PRs if at all possible.
- [X] The title should follow the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention (i.e. `fix(alert): not alerting during SSR render`, `docs(badge): update pill examples, fix typos`, `chore: fix typo in README`, etc). **This is very important, as the `CHANGELOG` is generated from these messages.**

description above includes:**


tmorehouse:
This should probably be made to work for both the directive and the component versions (using the same `BTooltip` key in the config for both directive and component versions).  As well, there should be a similar option for popovers (using the `BPopover` as they key).

src/utils/config.js
```diff
@@ -142,6 +142,9 @@ const DEFAULTS = {
     ariaLive: 'polite',
     ariaAtomic: 'true',
     role: null
+  },
+  vBTooltip: {
```
tmorehouse: Change this to `BTooltip` so that it will apply to both the directive version and the component versions.

src/directives/tooltip/tooltip.js
```diff
@@ -20,8 +23,10 @@ const validTriggers = {
 // Arguments and modifiers take precedence over passed value config object
 /* istanbul ignore next: not easy to test */
 const parseBindings = bindings => /* istanbul ignore next: not easy to test */ {
-  // We start out with a blank config
-  let config = {}
+  // We start out with a basic config
+  let config = {
+    boundary: String(getComponentConfig(NAME, 'boundary'))
```
tmorehouse: Explicitly use `BTooltip` here instead of a NAME variable.

AlanRezende:
Hi, I will add support for the components now. Since the popover and tooltip use a Mixin, It will be best to implement in the mixin `toolpop` or in each component?

tmorehouse:
@AlanRezende The prop `boundary` is defined in the mixin at https://github.com/AlanRezende/bootstrap-vue/blob/dev/src/mixins/toolpop.js#L62-L67

But the Miin is shared between BTooltip and BPopover, so in order to get the config value default unique for each, the individual components will need to have their own `boundary` prop defined (and then remove the `boundary` from the Mixin props.

AlanRezende:
That was what I figured out and didi exactly what you describe. But I have a problem.

With the directives  all works fine, I`m using nuxt to set new values in config.

With the components, not so much, if in dev I set new values in my dev config and rebuild all, it updates, but if I set in nuxt config nothing changes.

Is there some aditional code to make the directives work with the given configs?

tmorehouse:
Ah, that is because the default value is being evaluated when the modal is loaded, not at runtime.

I'll suggest a fix in a few moments.

tmorehouse:
Could you provide an example of your nuxt config you are using?

For the components, make sure you are using:

```js
    boundary: {
      // String: scrollParent, window, or viewport
      // Element: element reference
      type: [String, HTMLElement],
      default: () => getComponentConfig(...)
    },
```

and also include the HTMLElement polyfil (needed for SSR, as Node doesn't support `HTMLElement`) at the top:

```js
import { HTMLElement } from '../utils/safe-types'
```

AlanRezende:
All working now wi the nuxt config:

```
bootstrapVue: {
    bootstrapCSS: false, // Or `css: false`
    bootstrapVueCSS: false, // Or `bvCSS: false`
    config: {
      BTooltip: {
        boundary: 'window'
      },
      BPopover: {
        boundary: 'window'
      }
    }
  },
```

Care to explain why on component I had to use `() => getComponentConfig(...)` while in directive `getComponentConfig(...)` worked fine?

is it better to always use `() => getComponentConfig(...)`?

tmorehouse:
In the directives, the config is generated on demand in the `bind` function

For props, if you didn't use hte `() => ...` nomenclature, the result of `getComponentConfig` runs (is evaluated) immediately when the module is loaded/imported.  Setting the default to a function allows it to be evaluated lazily on first use of the component, to ensure the config is loaded before the component definition is evaluated.

Since in the directive version everything is wrapped in functions, we didn't need to use `() => ...`, as they are lazy evaluated on first use.

src/components/popover/popover.js
```diff
@@ -1,8 +1,12 @@
 import Vue from '../../utils/vue'
 import PopOver from '../../utils/popover.class'
 import warn from '../../utils/warn'
+import { getComponentConfig } from '../../utils/config'
 import normalizeSlotMixin from '../../mixins/normalize-slot'
 import toolpopMixin from '../../mixins/toolpop'
+import { HTMLElement } from '../../utils/safe-types'
```
tmorehouse: Could you move this up under `getComponentConfig`? we keep all the imports from the same directories together.

src/components/tooltip/tooltip.js
```diff
@@ -1,12 +1,16 @@
 import Vue from '../../utils/vue'
 import ToolTip from '../../utils/tooltip.class'
 import warn from '../../utils/warn'
+import { getComponentConfig } from '../../utils/config'
 import normalizeSlotMixin from '../../mixins/normalize-slot'
 import toolpopMixin from '../../mixins/toolpop'
+import { HTMLElement } from '../../utils/safe-types'
```
tmorehouse: Same here, Could you move this up under `getComponentConfig`? we keep all the imports from the same directories together.

src/mixins/toolpop.js
```diff
@@ -59,12 +59,6 @@ export default {
       type: String,
       default: null
     },
-    boundary: {
-      // String: scrollParent, window, or viewport
-      // Element: element reference
-      type: [String, HTMLElement],
-      default: 'scrollParent'
-    },
     boundaryPadding: {
```
tmorehouse: Just thinking, since `boundaryPadding` is also related to `boundary`, it should probably be moved into the config as well.

AlanRezende: I don't  see the boundaryPadding in the Bootstrap documentation, and it event isn`t to much information about it in bootstrap-vue.

I try to use it but didn`t see the effect.

I believe since, its not a oficial bootstrap config option, we could leave it as it is.

I try to find, but didn`t understand the use case for it.

tmorehouse: The boundary padding option makes sure the toolti/popover is inside the boundary element by # pixels.  say if the boundary element was window, and boundary padding was set to 0, then the tooltip could touch the edge of the window.  The current default is 5 (5px), so it will try to stay away from edge of the boundary element by 5px.

tmorehouse: It is a popper config option (which you can pass via the config normally) bootstrap v4 provides a link to Popper options in their docs.

tmorehouse: I've added in the boundaryPadding to the config.

AlanRezende:
I can confirm boundaryPadding is working based on config values (nuxt config used)

AlanRezende:
Do we need to change anything else?

tmorehouse:
I think we are good to go for a merge.

PR #3447, opened 2019-06-04
feat(b-table-lite): new `<b-table-lite>` lightweight table component

tmorehouse:
### Describe the PR

New `<b-table-lite>` table component. using shared _helper_ code of `<b-table>`, but tailored for just rendering arrays of items, without enhanced features.

New component **does not** support:
- Filtering
- Sorting
- Pagination
- Provider functions
- Busy state
- Selectable rows
- Fixed top / bottom rows

This PR refactored the `<b-table>` helper code-base to support lightweight table components.

When importing individual components, this will help consumers reduce their bundle size if they don't need the full `b-table` feature set.

### PR checklist

- [ ] Bugfix
- [x] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [x] Other (please describe) code refactor

- [x] No
- [ ] Yes (please describe)

description above includes:**

### DIFF
--- a/src/components/table/helpers/mixin-items.js
+++ b/src/components/table/helpers/mixin-items.js
@@ -51,8 +51,10 @@ export default {
       }, {})
     },
     computedItems() {
-      // Fallback if mixins not provided
-      return (this.paginatedItems || this.sortedItems || this.filteredItems || this.localItems || []).slice()
+      // Fallback if various mixins not provided
+      return (
+        this.paginatedItems || this.sortedItems || this.filteredItems || this.localItems || []
+      ).slice()
     },
     context() {
       // Current state of sorting, filtering and pagination props/values
--- a/src/components/table/helpers/mixin-items.js
+++ b/src/components/table/helpers/mixin-items.js
@@ -19,6 +19,13 @@ export default {
       // If provided the value in each row must be unique!!!
       type: String,
       default: null
+    },
+    value: {
+      // v-model for retrieving the current displayed rows
+      type: Array,
+      default() {
+        return []
+      }
     }
   },
   data() {
@@ -44,6 +51,17 @@ export default {
     computedItems() {
       // Fallback if mixins not provided
       return this.paginatedItems || this.sortedItems || this.filteredItems || this.localItems || []
+    },
+    context() {
+      // Current state of sorting, filtering and pagination props/values
+      return {
+        filter: this.localFilter,
+        sortBy: this.localSortBy,
+        sortDesc: this.localSortDesc,
+        perPage: parseInt(this.perPage, 10) || 0,
+        currentPage: parseInt(this.currentPage, 10) || 1,
+        apiUrl: this.apiUrl
+      }
     }
   },
   watch: {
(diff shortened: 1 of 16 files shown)


src/components/table/helpers/mixin-items.js
```diff
@@ -52,6 +76,21 @@ export default {
         /* istanbul ignore next */
         this.localItems = []
       }
+    },
+    // Watch for changes on computedItems and update the v-model
+    computedItems(newVal, oldVal) {
+      this.$emit('input', newVal)
```
jackmu95: Don't we need a `looseEqual()` check here too?
Otherwise we should remove the unused `oldVal` argument.

tmorehouse: `computedItems` always returns a new array reference when it changes

tmorehouse: So yeah, we could remove the unused argument var.

tmorehouse: Removed the unused argument

src/components/table/table-lite.js
```diff
@@ -0,0 +1,39 @@
...
+
+// Main table renderer mixin
+import tableRendererMixin from './helpers/mixin-table-renderer'
+
+// b-table-lite component definition
+// @vue/component
+export default Vue.extend({
+  name: 'BTableLite',
+  // Order of mixins is important!
+  // They are merged from left to fight, followed by this component.
```
jackmu95: `left to fight` => `top to bottom` or `first to last`

tmorehouse: Top to bottom because of prettier ;-)

jackmu95:
@tmorehouse Not to sure about this naming scheme. I have never been a fan of the "overload" of features of the current `<b-table>` component. But I do see that some people need the advanced functionality.

I would suggest to rename `<b-table-lite>` to `<b-table>` and the current `<b-table>` to `<b-table-advanced>`.

tmorehouse:
@jackmu95 reversing the names would be a rather large breaking change though

jackmu95:
@tmorehouse The thing is when we don't do it before v2.0.0 we can't do it before v3.0.0.
And I don't think it is too much effort to run a search and replace.

Would it make sense to give the users a notice to stick with one of the two components to safe some bundlesize?

tmorehouse:
Maybe the migration to Vue 3.x  and maybe Bootstrap V5, can trigger the component name flip

tmorehouse:
@jackmu95 Maybe we can get an opinion from @pi0 on the component names, and see what he thinks.

pi0:
I don't know the number of users that individually import components, but my guess is a single digit percentage. Making it by default is rather a breaking change and including the lite version increases bundle size for (the majority) of users that simply use plugin or module.

My suggestion is making this new component optional for those who individually import by name (and not by plugin)

tmorehouse:
@pi0 The code base addition for the `b-table-lite` component is quite small, since the majority of it (actually **all** of it) is shared code between the two components (`b-table` and `b-table-lite`), as it is all mixin based. The only bundle overhead is the new `table-lite.js` file , which imports a reduced set of mixins compared to `table.js`

pi0:
@tmorehouse OK let's bundle it by default :)

PR #3448, opened 2019-06-05
chore(config): updates to config handling

tmorehouse:
### Describe the PR

Updates to config handling, separating the `setConfig` handling from the config get routines for better tree-shaking

Performance wise, this PR reduces the number of times the `cloneDeep` utility is called in any one method improving both startup time and speeding up calls to `getComponentConfig(...)` (which is called each time a component with configurable prop defaults is instantiated)

The default configuration object is now deep frozen which will prevent any mutations, removing the need to deep clone it when requesting the default config.

### PR checklist

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [x] Other (please describe) code refactor

- [x] No
- [ ] Yes (please describe)

description above includes:**


src/bv-config.js
```diff
@@ -1,12 +1,10 @@
...
 // Utility Plugin for setting the configuration
 //
-import { setConfig } from './utils/config'
+import { installFactory } from './utils/plugins'
 
 const BVConfigPlugin = {
-  install(Vue, config = {}) {
-    setConfig(config)
-  }
+  install: installFactory({})
```
jackmu95: Would suggest to add a default value to `installFactory` itself so we can just call `installFactory()` here.

src/utils/config-defaults.js
```diff
@@ -0,0 +1,151 @@
...
+//
+//  props: {
+//    variant: {
+//      type: String,
+//      default: () => getConfigComponent('BAlert', 'variant')
+//    }
+//  }
+
+// prettier-ignore
+export default deepFreeze({
```
jackmu95: Do we really need to freeze the default config?
Wouldn't it be enough to make sure that we don't modify it anywhere?

tmorehouse: Just in case someone manually imports the default in thier app

As well... I want to remove the multiple calls to `cloneDeep` (cloning default, then cloning again afterwards) to improve installation performance.

src/bv-config.js
tmorehouse: yeah was thinking that...

tmorehouse: Checkout `utils/plugins.js` for defaults

jackmu95: Yeah, `{ components, directives, plugins } = {}` is enough.

tmorehouse: Done

PR #3542, opened 2019-06-17
feat(modal): submit and hide modal on enter

dubchenko:
### Describe the PR

Browser `alert` and `confirm` modal windows are submitted on Enter. `Bootstrap-vue` doesn't support such behaviour. This PR fixes this.
### PR checklist

- [ ] Bugfix
- [x] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [ ] No
- [x] Yes (please describe) pressing enter inside a modal will close it with OK status by default.

description above includes:**

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -77,6 +77,10 @@ export const props = {
     type: Boolean,
     default: false
   },
+  noSubmitOnEnter: {
+    type: Boolean,
+    default: false
+  },
   noEnforceFocus: {
     type: Boolean,
     default: false
@@ -600,10 +604,14 @@ export const BModal = /*#__PURE__*/ Vue.extend({
     onClose() {
       this.hide('headerclose')
     },
-    onEsc(evt) {
+    onKeyDown(evt) {
       // If ESC pressed, hide modal
       if (evt.keyCode === KeyCodes.ESC && this.isVisible && !this.noCloseOnEsc) {
         this.hide('esc')
+      }
+      // If Enter pressed, submit modal
+      if (evt.keyCode === KeyCodes.ENTER && this.isVisible && !this.noSubmitOnEnter) {
+        this.onOk()
       }
     },
     // Document focusin listener
@@ -850,7 +858,7 @@ export const BModal = /*#__PURE__*/ Vue.extend({
             'aria-hidden': this.isVisible ? null : 'true',
             'aria-modal': this.isVisible ? 'true' : null
           },
-          on: { keydown: this.onEsc, click: this.onClickOut }
+          on: { keydown: this.onKeyDown, click: this.onClickOut }
         },
         [modalDialog]
       )
--- a/src/components/modal/index.d.ts
+++ b/src/components/modal/index.d.ts
@@ -41,6 +41,7 @@ export interface BvMsgBoxOptions {
   noFade?: boolean
   noCloseOnBackdrop?: boolean
   noCloseOnEsc?: boolean
+  noSubmitOnEnter?: boolean
   headerBgVariant?: string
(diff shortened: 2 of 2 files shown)


src/components/modal/modal.js
```diff
@@ -600,11 +604,15 @@ export const BModal = /*#__PURE__*/ Vue.extend({
...
-    onEsc(evt) {
+    onKeyDown(evt) {
       // If ESC pressed, hide modal
       if (evt.keyCode === KeyCodes.ESC && this.isVisible && !this.noCloseOnEsc) {
         this.hide('esc')
       }
+      // If Enter pressed, submit modal
+      if (evt.keyCode === KeyCodes.ENTER && this.isVisible && !this.noSubmitOnEnter) {
+        this.onOk()
+      }
```
tmorehouse: This would close the modal even in there are cancel buttons, etc.

tmorehouse:
It probably would be better to handle this in the app logic rather than inside the modal component.

i.e. listening for `enter` keypress on an input, or replacing the default buttons with an `OK` button with `autofocus` prop set.

This PR assumes that `OK` is the default, when users may want the default action to be `Cancel`

src/components/modal/modal.js
dubchenko: Would it ok to pass the default action via props?

src/components/modal/modal.js
```diff
@@ -77,6 +77,10 @@ export const props = {
     type: Boolean,
     default: false
   },
+  noSubmitOnEnter: {
+    type: Boolean,
+    default: false
+  },
```
tmorehouse: This also assumes people want to close modal on <kbd>ENTER</kbd>, which is currently not the default behavior and would be considered a breaking change.

dubchenko:
msgBoxConfirm and msgBoxOk doesn’t provide access to buttons via $ref

src/components/modal/modal.js
tmorehouse: It would probably be better to have some option(s) to specify which default rendered buttons have focus when the modal opens (defaulting to no buttons having focus, as autofocusing an input or button on a modal will usually only announce the button/input but not the modal content when using screen readers.

A more robust approach is to have users listen for the `shown` event, and then focus the element they want (be it a button, input, or other focusable content control).

The `shown` event includes a reference to the modal element in the bvEvent object (the `target` property).  Once shown, you can then use `bvEvent.target.querySelector('some selector criteria`)` to find an element and focus it.

src/components/modal/modal.js
tmorehouse: Also, `submit` assumes a form is being used...  so its probably not the best name for the prop, as modals don't "submit"

tmorehouse:
If you pass an ID to the msgBox methods, you can then listen for `shown` events on `$root`, and match the ID passed, and then use the bvEvent.target to focus a prop.

tmorehouse:
If you had a msgBox with an OK and CANCEL button, and the user had CANCEL focused (i.e. they tabbed to CANCEL), your enter handler would trigger the OK action, not the cancel action (which is what the user would be expecting)

tmorehouse:
What would be better is a prop called `autofocus-button` which accepts a string of either `ok` or `cancel`, which modal could then use it's own internal `$refs` to focus the appropriate button (if the $ref exists and is visible).

It would need tests for coverage, and documentation on the prop, and note that it only works with the default built in footer buttons.

dubchenko:
@tmorehouse thank you!

DragosPeta:
When will we have this feature?

PR #3581, opened 2019-06-27
feat(b-modal): added a listener for backspace button 

gjdipietro:
### Describe the PR

This PR added a listener for the backspace button that will close the modal if the user has configured escape to close the modal and prevent default. Default behavior on some browsers for the
backspace button will navigate back. On single-page apps this can lead to errors as the modal would remain open but no longer exist.

### PR checklist

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**


src/components/modal/modal.js
```diff
@@ -600,11 +600,26 @@ export const BModal = /*#__PURE__*/ Vue.extend({
...
         this.hide('esc')
       }
+
+      // if BACKSPACE is pressed and we are not inside a textarea or input
+      // prevent default behavior of navigating to the previous page and hide the modal
+      if (
+        evt.keyCode === KeyCodes.BACKSPACE &&
+        (evt.target || evt.srcElement).tagName !== 'TEXTAREA' &&
+        (evt.target || evt.srcElement).tagName !== 'INPUT' &&
+        this.isVisible
```
tmorehouse: User could also have a content editable area (non form-input, such as a WYSIWYG editor) which would close the modal if the user is backspacing over a character in the content editable div (or other element).

tmorehouse:
Modal's have a `$route` watcher (when using Vue Router (or Nuxt, or similar based on Vue Router), which automatically closes the menu on a route change.

Chrome doesn't co to previous page on backspace, but Firefox does.  testing on Firefox using the current docs, and opening a modal (that doesn't have a text input focused), and pressing backspace, the modal automatically close when the route changes.

What browsers are you specifically seeing this issue in where the modal remains open?

gjdipietro:
Thanks for the additional information.  I am seeing the modal linger when backspace is hit in my codebase but I may have some additional code that causes it,  I'll check around the route.  I'm showing the modal based on a query param, its possible that the backspace is removing that query param and the route watcher isn't getting fired.

tmorehouse:
The route watcher should react to any changes in the route (including path, query params, and route params), as we watch `this.$route` which gets a new object created if any parameter is changed.

src/components/modal/modal.js
gjdipietro: Good catch, thank you!

src/components/modal/modal.js
```diff
@@ -608,18 +608,25 @@ export const BModal = /*#__PURE__*/ Vue.extend({
...
+        const target = evt.target || evt.srcElement
+        if (
+          target.tagName !== 'TEXTAREA' &&
+          target.tagName !== 'INPUT' &&
+          target.isContentEditable === false
+        ) {
+          return false
+        }
+        return true
+      }
```
tmorehouse: This could be simplified to:

```js
const el = evt.target || evt.srcElement
return el && (el.tagName === 'TEXTAREA' || el.tagName === 'INPUT' || el.isContentEditable)
```

tmorehouse:
I'm just trying to think of other situations where someone may expect backspace to work as expected.

elements that have a tabindex set (i.e. 0 or greater) which then become focusable, and may expect backspace to do something like delete a `tag` entry in a custom tag input.

I think this may need to be set behind a flag prop which defaults to `false` (current behaviour), and when set to `true` enables this new behaviour for backspace.

tmorehouse:
Also, tests would be required to provide coverage, and to make sure it works as epected.

tmorehouse:
As an alternative, a user could also use Vue Router route guards to prevent the page from changing while modal is open, or to close the modal if it is open.

alidavid0418:
Is this PR being out-of-date yet?

tmorehouse:
@alidavid0418 this PR need the suggested changes (so that it is opt-in, not always "on"), and needs test cases.

PR #3674, opened 2019-07-11
fix(b-form-file): ensure drag/drop handles `required` prop correctly (except IE11), add drag/drop accept type handling, fix drop in IE11 (closes #3673)

### LINKED ISSUE #3673: Dragged file input still shows required alert
### Describe the bug

When `required` is used in `b-form-file` and a file is dragged, it still shows select a file alert on submit.

It works fine with selecting file from the popup file system.

1. Go to https://codesandbox.io/s/form-required-drag-0uccq
2. Drag a file in the Choose file form.
3. Submit form, it shows alert.
4. Now click on `Choose a file` and select a file from the selector.
5. Submit form, it works.

### Expected behavior
1. Go to https://codesandbox.io/s/form-required-drag-0uccq
2. Drag a file in the Choose file form OR click on `Choose a file` and select a file from the selector.
3. Submit form. It should work for both methods.

tmorehouse:
### Describe the PR

**Note: New PR required**

Closes #3673 (except for IE 11, sorry)

Drag and drop files bypasses the file input altogether, which makes native browser validation constraints (i.e. `required`) not work, as the input does not reflect the dropped files.

Chrome, Firefox, Safari, Edge support dropping files directly on a file input, but IE11 and some older browser versions do not, so we can't take advantage of the native file input drop features (which would simplify the code drastically, and retain proper validity state).

All modern browsers support setting `InputElement.files` to a `FileList` array (which will satisfy `required` constraint), but again IE 11 doesn't support this (could be placed in a try/catch to trap/mask errors).

**What's new and changes:**

- Drag and drop now attempts to set the files array on the native file input (flattened `FileList`, filtered by accept criteria), so that `required` prop works with drag/drop (_except on IE_, the attempt to set the files is wrapped in a try/catch to suppress the error)
- Drag/drop now respects the `accept` criteria. Converts the `accept` prop to an array of regular expressions for testing the file type/extension of dropped files
- Drag and drop did now respects the `directory` setting (Except IE, which doesn't support directory mode)
- Uses a negative `z-index` on the opaque file input so that it doesn't interfere with drag/drop on IE (and other browsers). Clicking the custom input label will still open the native file browser (as the label is associated with the input via the `for` attribute).
- Added class `overflow-hidden` to custom label to prevent filenames from "breaking out" of custom file control
- filename array and files array that is passed to `scoped` slot are now always flattened (even in directory mode), but includes a new scoped property which provides the files in traversed mode
- Also adds a few fixes to the form-file docs

TODO:

- [x] `accept` file matching tests
- [x] Get drag/drop to work correctly on IE 11
- [x] Set `dropEffect` to `'none'` when file(s) can't be dropped (works on all but IE 11, as files are not known until the drop event happens)
- [ ] Add optional prop/slot for prompt when file's can't be dropped - or make drop-placeholder slot scoped with a `valid` scope prop (not applicable to IE 11, as files aren't known until dropped)
- [ ] Add notes in docs about limitations of required prop with drag/drop on IE11
- [ ] document requirement of `Promise` support for drag and drop mode
- [x] Add slots for `placeholder` prompt, and `drop-placeholder` prompt, as an alternative to the props.
- [ ] document new slots
- [x] add new slots to package.json meta
- [ ] minor custom SCSS updates to handle overflowing selected file name(s)
- [ ] additional test suites

Notes: Bootstrap v5 has a new structure for the custom label, which allows for better flexibility in handling overflowing names. We may want to consider using this newer format with Bootstrap v4.x (with a bit of custom SCSS, with variable sniffing) to take advantage of the new structure (and prepare for v5 when it is released)

### PR checklist

- [x] Bugfix
- [ ] Feature
- [x] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

- [ ] It should address only one issue or feature. If adding multiple features or fixing a bug and adding a new feature, break them into separate PRs if at all possible.

description above includes:**


src/components/form-file/form-file.js
```diff
@@ -199,29 +199,31 @@ export const BFormFile = /*#__PURE__*/ Vue.extend({
...
-      evt.stopPropagation()
-      if (this.noDrop || !this.custom) {
+      if (this.noDrop || this.disabled) {
+        evt.preventDefault()
+        evt.stopPropagation()
         return
       }
       this.dragging = true
-      evt.dataTransfer.dropEffect = 'copy'
+      if (evt && evt.dataTransfer) {
```
jackmu95: `evt` check is unnecessary, we would have problems otherwise above too...

tmorehouse:
One way we could get around the required prop (for IE) would be to include a second opaque input (type text), _after_ the native file input, and populate it with the file names.  Rather than placing the required attribute on the file input, we would place it on the text input. (when in custom mode, and drag/drop is enabled) It would be kind of hacky... but it might get around the issue.

jackmu95:
@tmorehouse Yeah, sounds a bit hacky but it could work out.
Would you implement this behavior only for IE or for all browsers?

tmorehouse:
If implemented that way, it probably should be targeted for IE only... But after some thought it might interfere with validator plugins (that may want to check file types)

PR #3773, opened 2019-07-26
Shamus03 patch 1

Shamus03:
### Describe the PR

Add a default value of `false` for the `centered` prop.  I find myself almost always setting this to `true`, so it would be nice to have the ability to set it as a default in my projects, but I cannot without a default value in `config-defaults.js`.

### PR checklist

- [ ] Bugfix
- [x] Feature
- [ ] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -40,7 +40,7 @@ export const props = {
   },
   centered: {
     type: Boolean,
-    default: false
+    default: () => getComponentConfig(NAME, 'centered')
   },
   scrollable: {
     type: Boolean,
--- a/src/utils/config-defaults.js
+++ b/src/utils/config-defaults.js
@@ -89,6 +89,7 @@ export default deepFreeze({
   BModal: {
     titleTag: 'h5',
     size: 'md',
+    centered: false,
     headerBgVariant: null,
     headerBorderVariant: null,
     headerTextVariant: null,


tmorehouse:
The only concern I have with this, is that if someone uses a 3rd party component that makes use of modal, users now need to explicitly use the `v-bind` method to make `centered` false (i.e. `:centered="false"`), as this is a Boolean prop (not one that accepts a string or numeric value). Centered modals have issues with overflowing content on short screens.

Config was mainly meant for things like variants, default labels, etc, and not really for boolean props (which now loose there meaning - i.e. `no-this` or `that`).

Also note that each prop that uses a function as a default slows down the instantiation of the component due to the function overhead processing.

An easier way to achieve what you want would be to create your own extended b-modal component and override the prop:

```js
import { BModal } from 'bootstrap-vue'
export default {
  name: 'MyModal',
  extends: BModal,
  props: {
    centered: {
      type: Boolean,
      default: true
    }
  }
}
```

Shamus03:
Good point.  I was even going to ask why not just make every prop part of the config, but given your concerns I think the drawbacks outweigh the benefits.

I had considered forcing all modals to be centered via css to enforce stylistic consistency, but I hadn’t considered just making an extended component.  I’ll do that instead.

Shamus03:
Unfortunately extending the component doesn't work for the `$bvModal.msgBoxOk` and `$bvModal.msgBoxConfirm` helpers.  That was the main reason I wanted a default!  Any suggestions for how to apply a default to that helper?

tmorehouse:
One could create a default msgBox config object that could be imported into your components, and then merged with any new options:

A file to contain the defaults:
```js
export const msgBoxDefaults = {
  centered: true,
  size: 'lg'
}
```

And in your component:
```js
import { msgBoxDefaults } from 'path/to/defaults'

export default {
  // ...
  methods: {
    showMsgBoxOk() {
      return this.$bvModal.msgBoxOk('Message', {
        ...msgBoxDefaults,
        title: 'MsgBox Title',
      })
    }
  }
}
```

Or you could create a helper method that does the merging for you.

Or create a [Vue.mixin](https://vuejs.org/v2/api/#Vue-mixin) that adds your own instance method:

```js
import Vue from 'vue'

Vue.mixin({
  methods: {
    $myMsgBoxOk(message = '', options = {}) {
       return this.$bvModal.msgBoxOk(message, {
        ...msgBoxDefaults,
        ...options
      })     
    }
  }
})
```
And do this in the main entrypoint of your app. this will add a new $myMsgBoOk method to every Vue instance in your app.

PR #4169, opened 2019-09-29
feat(b-form-input, b-form-textarea): add `lazy` modifier prop to update v-model on change/blur event

tmorehouse:
### Describe the PR

Adds a new prop `lazy` to `<b-form-input>` and `<b-form-textarea>`, to only update the v-model on `change`/`blur` event, emulating the Vue `.lazy` modifier

Target release: v2.1.0

To Do:

- [x] Add unit tests for `lazy` prop

### PR checklist

- [ ] Bugfix
- [x] Feature
- [ ] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

description above includes:**


jackmu95:
@tmorehouse Our current implementation of the `trim`, `number` and the new `lazy` props behaved differently then the Vue.js `v-model` modifiers. My recent commit brought them in line and simplified the logic a lot.

This is the test setup I used in the playground:

```vue
<div>
  <input
    v-model.trim="text1"
    class="form-control mb-2"
    @input="onInput('text1', $event.target.value)"
    @change="onChange('text1', $event.target.value)"
  >
  
  <input
    v-model.trim.number="text2"
    class="form-control mb-2"
    @input="onInput('text2', $event.target.value)"
    @change="onChange('text2', $event.target.value)"
  >
  
  <input
    v-model.lazy.trim="text3"
    class="form-control mb-2"
    @input="onInput('text3', $event.target.value)"
    @change="onChange('text3', $event.target.value)"
  >
 
  <b-form-input
    v-model="text4"
    class="mb-2"
    @input="onInput('text4', $event)"
    @change="onChange('text4', $event)"
  ></b-form-input>
 
  <b-form-input
    v-model="text5"
    trim
    lazy
    @input="onInput('text5', $event)"
    @change="onChange('text5', $event)"
  ></b-form-input>
</div>
```

```js
{
  data() {
    return {
      text1: '',
      text2: '',
      text3: '',
      text4: '',
      text5: '',
    }
  },
  methods: {
    onInput(name, value) {
      const localValue = this[name]
      console.log('input', name, value, localValue, value === localValue)
    },
    onChange(name, value) {
      const localValue = this[name]
      console.log('change', name, value, localValue, value === localValue)
    }
  }
}
```

src/mixins/form-text.js
```diff
@@ -175,28 +151,20 @@ export default {
...
         evt.preventDefault()
         return
       }
-      this.updateValue(formatted, false)
-      this.$emit('change', formatted)
+      this.localValue = formattedValue
+      this.$emit('update', formattedValue)
+      this.$emit('change', formattedValue)
     },
     onBlur(evt) {
```
tmorehouse: Lazy formatting should be performed here on Blur, and not on change, as change can be triggered by character composition (for Chinese IME composition), mobile browser auto-complete or auto-correct.  Normally when blur happens a change event is fired first ,follewed by the blur event.

src/mixins/form-text.js
```diff
@@ -105,68 +106,43 @@ export default {
...
-        /* istanbul ignore next */
+      /* istanbul ignore next */
+      if (formattedValue === false || evt.defaultPrevented) {
         evt.preventDefault()
         return
       }
-      this.updateValue(formatted, this.lazy)
-      this.$emit('input', formatted)
+      this.localValue = value
+      if (!this.lazy) {
```
tmorehouse: And typically the formatted value (when formatted) should update the actual value of the input (this.$input.value = formatted).  This will work if the user has a v-model applied, but it should also work when v-model isn't explicitly used.

src/mixins/form-text.js
```diff
@@ -105,68 +106,43 @@ export default {
...
+        this.$emit('update', formattedValue)
+      }
```
tmorehouse: And typically the formatted value (when formatted) should update the actual value of the input (this.$input.value = formatted).  This will work if the user has a `v-model` applied, but it should also work when `v-model` isn't explicitly used, so the input's DOM prop value should be updated.

tmorehouse:
Trim should leave the value of the input alone... if the v-model is equal to the trimmed value of the input (local value or this.$refs.input.value) then the input visually should still have spaces on either end while the v-model has them trimmed off  (which ishow Vue does it).

When leaving the input (blur/change), the displayed text in the input gets is leading and trailing spaces remove

jackmu95:
@tmorehouse Sorry I misunderstood how the formatter should work.
I made some adjustments and the formatter should behave as before now and the modifier props act like Vue.js.

jackmu95:
@tmorehouse The only thing that still bothers me is that the `v-model` is updated on `blur` and not on `change` when in `lazy` mode. We could change that but the question is which value we should emit with the `change` event then?

IMHO we should emit the value the input will receive on `blur`.

src/mixins/form-text.js
```diff
@@ -158,26 +166,31 @@ export default {
...
-        /* istanbul ignore next */
+      /* istanbul ignore next */
+      if (formattedValue === false || evt.defaultPrevented) {
         evt.preventDefault()
         return
       }
-      this.updateValue(formatted)
-      this.$emit('change', formatted)
+      this.localValue = formattedValue
+      this.updateValue(formattedValue)
```
tmorehouse: @jackmu95 do you think we should use the `force` option of `updateValue` on change?

tmorehouse: Added force true for the change event, to better mimic Vue's lazy modifier

PR #4364, opened 2019-11-09
fix(v-b-modal): open modal using `ENTER` key on non-button elements for A11Y

markcarver:
### Describe the PR

When the `v-b-modal` directive is used on non-link elements (`a`), users can only open it using the spacebar. This adds inconsistency where other implementations (because they're link elements) can be opened using the `Enter` key.

From https://www.w3.org/WAI/GL/wiki/Making_actions_keyboard_accessible_by_using_keyboard_event_handlers_with_WAI-ARIA_controls#Example_1:_Using_Space_to_activate_a_button:

> Adding the ARIA role "button" makes it clear that this is a button control and not a link. Because we are using an anchor element, the control is focusable and the browser will automatically call the onclick handler for the Enter keystroke. Since users expect to be able to activate buttons with either Enter or Space, a keyboard handler provides the support for activating the button via the Space character.

> **Note that supporting SPACE in addition to ENTER is not required by WCAG 2.0; this control would still be keyboard accessible without the key event handler for SPACE. However, supporting platform keyboard conventions for controls is strongly encouraged.**

### PR checklist

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement
- [x] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [x] No
- [ ] Yes (please describe)

- [x] Includes any needed TypeScript declaration file updates

description above includes:**

### DIFF
--- a/src/directives/modal/modal.js
+++ b/src/directives/modal/modal.js
@@ -10,6 +10,7 @@ import {
 } from '../../utils/dom'
 import { isString } from '../../utils/inspect'
 import { keys } from '../../utils/object'
+import KeyCodes from '../../utils/key-codes'
 
 // Emitted show event for modal
 const EVENT_SHOW = 'bv::show::modal'
@@ -25,7 +26,7 @@ const getTarget = ({ modifiers = {}, arg, value }) => {
 }
 
 const getTriggerElement = el => {
-  // If root element is a dropdown item or nav item, we
+  // If root element is a dropdown-item or nav-item, we
   // need to target the inner link or button instead
   return el && matches(el, '.dropdown-menu > li, li.nav-item') ? select('a, button', el) || el : el
 }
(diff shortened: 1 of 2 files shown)


src/directives/modal/modal.js
```diff
@@ -47,7 +47,7 @@ const bind = (el, binding, vnode) => {
       if (!isDisabled(currentTarget)) {
         const type = evt.type
         // Open modal only if trigger is not disabled
-        if (type === 'click' || (type === 'keydown' && evt.keyCode === 32)) {
+        if (type === 'click' || (type === 'keydown' && (evt.keyCode === 13 || evt.keyCode === 32))) {
```
tmorehouse: We should import the KeyCode object from `./../utils/key-kodes` rather than using literal numeric values.

```js
import KeyCodes from '../../utils/key-codes'

// ...

const keyCode = evt.keyCode
if (type === 'click' || (type === 'keydown' && (keyCode === KeyCodes.SPACE || keyCode === KeyCodes.ENTER))) 

```

tmorehouse: Tweaked the code a bit to use the key-codes utility and linted

tmorehouse:
@markcarver Thanks!

I've tweaked the code a bit to use KeyCode constants utility, as well as added in two unit tests to ensure both <kbd>ENTER</kbd> and <kbd>SPACE</kbd> triggers work.

PR #4734, opened 2020-02-06
fix(b-table): sticky header borders (closes #4123)

### LINKED ISSUE #4123: row is visible behind sticky table header
### Describe the bug

When using the sticky header I can see when scrolling part of the background row, and it looks kind of bad.

My table is defined as: 

`<b-table primary-key="id" :fields="fields" :items="nodes" :filter="filter" head-variant="light" sticky-header="77.3vh" show-empty hover bordered>`

image

~~Also, setting the height to something like `calc(100vh-100px)` does not work.~~

### Expected behavior

The background row should be completely hidden behind the header. I only found this to be an issue on chrome.

### Versions

**Libraries:**

- BootstrapVue: ^2.0.0-rc.28
- Bootstrap: ^4.3.1
- Vue: ^2.6.10

**Environment:**

- OS: Windows 10
- Browser: Chrome 76.0.3809.132

alexandreDavid:
### Describe the PR

The table sticky header has 1px which displays the `<tr>` when the users scrolls.
It's because of the top border on the `<th>`.

This PR removes the 1px border on `<b-table>` sticky header.

Closes #4123.

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement
- [ ] ARIA accessibility
- [ ] Documentation update
- [ ] Other (please describe)

- [ ] No
- [ ] Yes (please describe)

- [x] The title should follow the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention (i.e. `fix(alert): not alerting during SSR render`, `docs(badge): update pill examples`, `chore(docs): fix typo in README`, etc). **This is very important, as the `CHANGELOG` is generated from these messages.**


src/components/table/_table.scss
```diff
@@ -110,6 +110,7 @@
     // Default `max-height` before a scrollbar will show
     // We don't use `height` as table could be shorter than this value
     max-height: $b-table-sticky-header-max-height;
+    border-top: $table-border-width solid $table-border-color;
```
tmorehouse: This is assuming users haven't set a `table-{variant}` class on the table (via the `table-variant` pop), which sets the color of the borders as well. this would override that.

src/components/table/_table.scss
```diff
@@ -125,6 +126,7 @@
     // Positioning of sticky headers
     .b-table-sticky-header > .table.b-table > thead > tr > th {
       // Header cells need to be sticky on top
+      border-top: 0;
```
tmorehouse: This assumes people are using `no-border-collapse` at the same time, which may not be the case.

src/components/table/_table.scss
tmorehouse: probably better written as:

```scss
    border-top-width: $table-border-width;
```

To just affect the border width only, and not the color of the border

tmorehouse: Otherwise border colors will need to be generated for each possible variant color.

tmorehouse:
I think we will wait for Bootstrap v5 as it adds better border and background inheritance to the `.table` class.

tmorehouse:
Closing in favour of Bootstrap v5 table border improcements

PR #5344, opened 2020-05-11
feat(b-nav-item-dropdown): improve default handling of dropdown toggle link (closes #3942)

### LINKED ISSUE #3942: Add an href property to b-nav-item-dropdown

Before the javascript has loaded, clicking on the b-nav-item-dropdown causes the page to change and there's no way to control what page the b-nav-item-dropdown goes to because it's hardcoded to '#'.

Have b-nav-item-dropdown support a new href property that would allow changing the href while keeping the default as '#'.

The href value of '#' is hardcoded here:
https://github.com/bootstrap-vue/bootstrap-vue/blob/21fab353f2bbaa4a5698bc646bfb99213117a916/src/components/nav/nav-item-dropdown.js#L48

jackmu95:
### Describe the PR

This PR improves the link handling of `<b-nav-item-dropdown>` by assining the user provided ID as the toggle link `href` has link.

When `id` prop isn't set (defaults to `null`) the `href` for the `<b-link>` i set to `#`. If an `id` is provided, then the `href` for `<b-link>` will be set to `#{id}`. This way, before the page is hydrated, the link will just scroll to the A tag (which should already be in view and hence not scroll) rather than scrolling to the top of the page.

Closes #3942.

### PR checklist

- [x] Feature (adds a new feature to BootstrapVue) - `feat(...)`, requires a minor version update
- [x] Enhancement (augments an existing feature) - `feat(...)`, requires a minor version update
- [x] ARIA accessibility (fixes or improves ARIA accessibility) - `fix(...)`, requires a patch or minor version update
- [ ] Other (please describe)

- [x] No

- [x] Includes documentation updates
- [x] CodeCov for patch has met target (all changes/updates have been tested)

description above includes:**


jackmu95:
@scyclops Can you create a CodeSandbox based on this [template](https://codesandbox.io/s/goofy-forest-tb3jl) with a minimal reproduction of your Nuxt.js setup to confirm that the problem is resolved?

tmorehouse:
I just checked in the PR preview, and checked the SSR rendered source for the version dropdown in the docs nav, and it renders the following (before hydration):

```html
a aria-haspopup="true" aria-expanded="false" href="#" target="_self" class="nav-link dropdown-toggle mr-md-2"><span>Pull Request</span></a>
```

So the `href="#"` is still there before Vue kicks in.

Perhaps we could try and use a `<button>` element for the dropdown toggle in `b-nav-item-dropdown` and see how the Bootstrap v4 styling work on it.

jackmu95:
@tmorehouse Or we just expose `href` as prop as suggested in the issue.
It works fine for most of the users and for those who need to adjust it would have the possibility.

tmorehouse:
We would still need to make sure event.preventDefault() is called... `<b-link>` doesn't do this automatically when HREF is anything other than `#`.

Another option would be to set an ID on the `<a>` and set href to point to this ID (unfortunately we cant use `safeId()` auto-ID as it only runs client side)

jackmu95:
The `toggle()` method in `dropdownMixin` calls `event..preventDefault()` which the `click` ebent on the `<b-link>` is assigned to. So we should be fine.

tmorehouse:
Hmm, we should be adding `role="button"` to the rendered `<a>` tag that way an HREF won't confuse screen reader users.

How about we add an ID prop.  This ID will be added to the `<a :id="id" :href="'#' + id'" role="button">` and then used as the `href`, that way only hash HREFs are generated.

`<a

src/components/nav/nav-item-dropdown.js
```diff
@@ -41,23 +47,29 @@ export const BNavItemDropdown = /*#__PURE__*/ Vue.extend({
...
+
     const $button = h(
       BLink,
       {
-        ref: 'toggle',
         staticClass: 'nav-link dropdown-toggle',
         class: this.toggleClasses,
-        props: { disabled: this.disabled },
+        props: {
+          href: this.href || `#${buttonId}`,
```
tmorehouse: I still don't think we should provide an `href` prop, as people may want to try and add `/some/path/here`, when it should only be a hashed based URL (i.e. `#somehash`)

tmorehouse: And having `href` work differently here than other places where it is normally used will add to user confusion.

src/components/nav/nav-item-dropdown.js
```diff
@@ -41,57 +47,62 @@ export const BNavItemDropdown = /*#__PURE__*/ Vue.extend({
...
+
+    const $toggle = h(
       BLink,
       {
-        ref: 'toggle',
         staticClass: 'nav-link dropdown-toggle',
         class: this.toggleClasses,
         props: {
-          href: '#',
+          href: this.href || `#${toggleId || ''}`,
```
tmorehouse, line 58: use: 
```js
href: `#${this.id || ''}`,
```

or 

```js
href: `#${toString(this.id)}`,
```

tmorehouse: This way the `href` is only set to something other than `#` when the user supplied an ID.  The outer `<li>` will have this ID if a user provide one.

tmorehouse: And then we can get rid of the `href` prop altogether.

src/components/nav/nav-item-dropdown.spec.js
```diff
@@ -95,16 +95,89 @@ describe('nav-item-dropdown', () => {
...
+    await waitNT(wrapper.vm)
+
+    const $toggle = wrapper.find('.dropdown-toggle')
+    expect($toggle.find('span').exists()).toBe(true)
+    expect($toggle.text()).toEqual('bar')
+
+    wrapper.destroy()
+  })
+
+  it('should have correct toggle content from "text" slot', async () => {
```
tmorehouse, line 132: maybe add `(deprecated)` in the test description?

PR #5564, opened 2020-07-08
fix: properly handle special characters in user-provided IDs (closes #4927, #5561)

### LINKED ISSUE #4927: b-form-tags id cannot start with numbers
### Describe the bug

When using the `b-form-tags` component, if you use an `id` or `input-id` that starts with a number, it throws a `DOMException` when attempting to add a new tag.

Capture

1. Go to The tags docs (https://bootstrap-vue.js.org/docs/components/form-tags)
2. Click on the first live example, add a number to the start of the `id` prop (`1-tags-basic`)
```
<template>
  <div>
    <label for="tags-basic">Type a new tag and press enter</label>
    <b-form-tags input-id="1-tags-basic" v-model="value" class="mb-2"></b-form-tags>
    <p>Value: {{ value }}</p>
  </div>
</template>

<script>
  export default {
    data() {
      return {
        value: ['apple', 'orange']
      }
    }
  }
</script>
```

3. Type a new tag and press enter.
4. Note that the tag is not added. Open dev tools and observe the console error.

### Expected behavior

As per HTML5 specs (https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute), digits should be usable in the `id` of input elements. The component's `id` and `input-id` should allow for this.

### Versions

**Libraries:**

- BootstrapVue: 2.6.1
- Bootstrap: 4.4.1
- Vue: 2.6.10

### LINKED ISSUE #5561: Failed to execute 'querySelector' on 'Element': '#/my-id' is not a valid selector.
### Describe the bug

The `b-form-group` adds `aria-describedby` to the input specified, and looks for it based on the given `label-for`. When the id of the input uses special characters, which are valid HTML5 (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), a console error happens, and the `aria-describedby` is not added.

The reason for this is that the special characters are not escaped inside `querySelector`.

I encountered this bug when using an id that comes from a FHIR Questionnaire, which usually has ids like `/15478-3`. We are not in control of these ids, and it's kind of a standard practice to use this `/` character for them.

Just use `/my-id` for example:
```
<b-form-group ... label-for="/my-id">...</b-form-group>
```
See for a working example, and notice the console error, which makes the component "failing":  https://codesandbox.io/s/serene-dewdney-ydnnu

### Expected behavior

As these characters are allowed in HTML5, though have a compatibility warning for HTML4, this should just work, as a library like this must not be concerned with these warnings, as it's up to the user to decide if they want to be compatible.

### Versions

**Libraries:**

- BootstrapVue: 2.15.0
- Bootstrap: 4.5.0
- Vue: 2.6.11

**Environment:**

- Device: Windows laptop
- OS: Windows 10
- Browser: Edge
- Version: Version 84.0.522.28 

### Demo link

Thx for the playground and export links, this really makes it easy to have min. reproductions.

https://codesandbox.io/s/serene-dewdney-ydnnu

### Additional context

Like mentioned earlier, we are in no control of the ids being used, as our application is a consumer of these, and we have to use them.

---

The solution is just to use `CSS.escape` to escape the selector. I will see to create a PR to resolve this. Unless there are any concerns in doing this, as I don't know all internals of this library.

dietergeerts:
Describe the PR

Special characters are allowed in HTML5 (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) but need to be escaped when used in a selector for usage in e.g. `document.querySelector()`.

This PR makes sure to escape user-provided IDs (e.g. for `<b-form-group>` or a `<b-tags-input>`) when used with the `select()` util.

Closes #4927, #5561.

### PR checklist

- [ ] Other (please describe)

- [x] No

description above includes:**


dietergeerts:
I also would like to mention that for me, at first, it wasn't clear that when you use `label-for`, that the `id` on the input has to be provided. I though that the `b-form-group` would add this automatically. After reading the docs again, I see it's stated, but like a bit "hidden". I believe it would be beneficial to have it more prominent in the docs.

dietergeerts:
In the original post, I stated that `CSS.escape` should be used, but as it's experimental, and not supported by all browser supported by bootstrap-vue, I included a polyfill instead.

jackmu95:
@dietergeerts IMHO the escaping isn't wort it. Especially when we need to add an additional dependency.
Just make sure to only use IDs that are supported by `.querySelector()` out fo the box.

dietergeerts:
@jackmu95, like explained in the issue, we are not in control of the used id. The extra dependency was added because `CSS.escape` is still experimental and not supported by all browsers this library supports. So if this is not fixed, we'll be unable to use this library.

jackmu95:
@dietergeerts You have to make sure to assign a valid ID to the control of the `<b-form-group>` in the first place.
https://codesandbox.io/s/nifty-brook-kztwz?file=/App.vue

dietergeerts:
> @dietergeerts You have to make sure to assign a valid ID to the control of the `<b-form-group>` in the first place.
> https://codesandbox.io/s/nifty-brook-kztwz?file=/App.vue

Well `/1547-3` for example is a valid id in HTML5, and one that we get from a 3rd party system as unique id to be used in our forms.

jackmu95:
@dietergeerts I will revisit this later today.

`css.escape` is a polyfill and would cause side-effects. I think we should create a util for this ourselves which borrows most of the code.

There are also other places we need to escape a user-provided ID.

jackmu95:
Thanks @dietergeerts!

dietergeerts:
@jackmu95 , thx for finishing up and having it merged.

PR #5632, opened 2020-08-03
fix(b-form-input): fix debounce when value does not change

KarboniteKream:
### Describe the PR

This PR fixes the behavior of `debounce` on `b-input-form` when the initial and final value are the same.

Changing input like `value -> valu -> value` will cause the `update` event to receive `valu` instead of `value`. This is because `value` is the same as the initial value and will not update the triggered event.

I've modified the behavior to always replace the triggered event which then compares the initial and updated value before emitting an `update` event.

Now the flow `value -> valu -> value` will not emit an `update` event.

### PR checklist

**What kind of change does this PR introduce?**

- [x] Bugfix (fixes a boo-boo in the code)
- [ ] Feature (adds a new feature to BootstrapVue)
- [ ] Enhancement (augments an existing feature)
- [ ] ARIA accessibility (fixes or improves ARIA accessibility)
- [ ] Documentation update (improves documentation or typo fixes)
- [ ] Other (please describe)

**Does this PR introduce a breaking change?**

- [x] No
- [ ] Yes


jackmu95:
@KarboniteKream I made some slight adjustments to the codes so share the exact same update logic between immediate and debounced value changes.

It also makes sure to always clear the debounce when `updateValue()` get called.
This should get around any other problems with changed debounce values between value updates.

Can you please re-check if this solves your problem in the generated CodeSandbox wit the latest changes:
https://codesandbox.io/s/bootstrapvue-starter-project-tke9u

KarboniteKream:
@jackmu95 Thanks! I've verified that the code still works correctly.

jackmu95:
Thanks @KarboniteKream!

PR #5981, opened 2020-10-29
feat(config): improved defaults handling (closes #4507, #5138, #5291, #5459, #5958)

### LINKED ISSUE #4507: Application wide component settings

Many components have properties that are meant to provide strings that a user either sees or hears and as a result, needs to be translated.

I see we have the ability to configure some props via the plugin settings, but is there a pattern to how we choose which are included? Or is it just need-based? 

I wonder if there is a way we can make the available settings dynamic so that any prop on any component can be configured application wide without the need to code for it. Add a prop and it becomes available. My use case primarily centers around settings translated strings once for the application, but I feel that if being addressed, there is no reason not to address it for any prop.

### Describe alternatives you've considered

If this is not feasible, are PRs to add more settings typically accepted or are there specific requirements as to which props or types of props would be accepted?

### Additional context

None

### LINKED ISSUE #5138: BFormGroup labelClass configurable via plugin settings

We want to target the "primary" label element within form-group elements for styling (like bold font-weight) without having to be aware of all the other label elements that can be rendered within a BFormGroup (checkboxes, radios, etc).

We would like to be able to configure the labelClass prop for BFormGroup globally via the plugin settings.

### Describe alternatives you've considered

We have considered applying the labelClass for every field, but given we'd use the same class almost universally, configuring it at the plugin level would be idea.

Alternatively, if the library itself provided a built-in class to the primary label element, that would help enable the ability to theme it as well.

### Additional context

N/A

### LINKED ISSUE #5291: Ability to set a default for no-auto-hide for BToasts

I want to switch off auto-hide for all toasts, but I currently have to do this using the `no-auto-hide` prop on each `b-toast` individually, afaik.

I'd like to be able to default auto-hide to off using a setting.

### Describe alternatives you've considered

My current workaround is this:

```js
BToast: {
    // Can't default to no-auto-hide, so...
    // 99999999/1000/60/60 = 27.7777775 hours
    autoHideDelay: 99999999
  }
```

...which works, but it's a bit of a kludge?

### LINKED ISSUE #5459: Is it possible to have global settings for boolean props?

For instance, Modal supports global settings:
```js
"BModal": {
    "titleTag": "h5",
    "size": "md",
    "headerBgVariant": null,
    "headerBorderVariant": null,
    "headerTextVariant": null,
    "headerCloseVariant": null,
    "bodyBgVariant": null,
    "bodyTextVariant": null,
    "footerBgVariant": null,
    "footerBorderVariant": null,
    "footerTextVariant": null,
    "cancelTitle": "Cancel",
    "cancelVariant": "secondary",
    "okTitle": "OK",
    "okVariant": "primary",
    "headerCloseContent": "&times;",
    "headerCloseLabel": "Close"
},
```

But values like `noFade` or `hideBackdrop` do not work

> BvConfig - Unknown config property "BModal.noFade"

Being able to set properties like that in global config.

### Describe alternatives you've considered

Creating an extra object for config and passing it for each modal `v-bind="modalProps"`. Easy to forget.

### LINKED ISSUE #5958: Enhancement: More settings for BModal

Lots of BModal props are not globally configurable with the settings.  This result in lots of repeating code if, in example, you don't want the modal to be closed on backdrop click in your application.

Add the following properties to the BModal Settings:
* centered
* scrollable
* button-size
* no-stacking
* no-fade
* no-close-on-backdrop
* no-close-on-esc
* title-class
* header-class
* modal-class
* dialog-class
* content-class
* body-class
* footer-class
* hide-header
* hide-footer
* hide-header-close
* hide-backdrop
* ok-only
* ok-disabled
* cancel-disabled

### Describe alternatives you've considered

Create a custom modal that just call the BModal with the prefered default settings.

Hiws:
### Describe the PR

A rewrite of the config defaults, moving them to the separate components they belong to.
This should allow for better tree shaking, and bundle size allowing us to increase the amount of configurable props without hurting performance.

Closes #4507.
Closes #5138.
Closes #5291.
Closes #5459.
Closes #5958.
Replaces #5961.

- [x] Delete `config-default.js` and move all the defaults to their respective components.
- [x] Breakpoint configuration and defaults still works.
- [x] Rewrite config tests.
- [x] Test with Nuxt.
- [x] Fix `<b-form-datepicker>` fallback to `<b-date>` config.
- [x] Fix `<b-form-timepicker>` fallback to `<b-time>` config.
- [x] Fix `<b-form-timepicker>`  and `<b-time>`  fallback to `<b-form-spinbutton>` config.
- [x] Rewrite default config documentation.


jackmu95:
@Hiws Here is a CodeSanbox with Nuxt.js:
https://codesandbox.io/s/bootstrapvue-v2190-nuxtjs-1quvd

jackmu95 (approved):

PR #6116, opened 2020-12-01
chore: release v2.20.1

jackmu95:
### 🐛 Bug Fixes

- `general`
  - #6113 User supplied prop function detection
- `table`
  - c375ce90 Use original value for fallback when number parsing fails in `defaultSortCompare`


Hiws (approved):

dreysolano:
@jackmu95 - This change is causing inputs to break when no `formatter` is provided as default is `null` and thus `this.formatter.name` throws a "Cannot read property 'name' of null" error.
`
hasFormatter() {
    return this.formatter.name !== props.formatter.default.name
}
`

PR #6481, opened 2021-03-05
fix(b-modal): check for wether the body is overflowing

wparad:
### Describe the PR

Opening a modal incorrectly always pads the body and the modal even when there is no reason to because the scroll bar isn't needed.

image

### PR checklist

- [x] No

- [ ] It should address only one issue or feature. If adding multiple features or fixing a bug and adding a new feature, break them into separate PRs if at all possible.
- [ ] The title should follow the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention (i.e. `fix(alert): not alerting during SSR render`, `docs(badge): update pill examples`, `chore(docs): fix typo in README`, etc.). **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type (patch or minor).**


Hiws:
This is intentional though, to avoid having page shift in the background when opening/closing a modal.

Here's how it currently looks on the docs when opening a modal:

https://user-images.githubusercontent.com/7057431/110166155-3b945080-7df4-11eb-9ed5-fd671d133d83.mp4

Notice how the added padding makes it so the background doesn't shift.

Compared to the docs with your change applied:

https://user-images.githubusercontent.com/7057431/110166227-5666c500-7df4-11eb-9715-c0c371c79219.mp4

wparad:
I see, but then you have the opposite problem if the screen doesn't have a scrollbar, and the content is not as wide as the screen. What's the fix then?

Hiws:
> I see, but then you have the opposite problem if the screen doesn't have a scrollbar, and the content is not as wide as the screen. What's the fix then?

I'm not quite sure what you mean.
No padding should be applied if there's no scrollbar on the page.


https://user-images.githubusercontent.com/7057431/110167326-e2c5b780-7df5-11eb-98c3-67234fbeb0c5.mp4


<https://codepen.io/Hiws/pen/Pobdwzm>

wparad:
Maybe we just need to check that it is greater than 1px off, because I'm struggling with this:
image

Hiws:
I see what you mean in your codepen, but I'm not sure what can be done about it.
If you have a solution feel free to edit your PR accordingly :)

wparad:
I've made an update here, it now checks the body element by length instead of by width.

PR #6639, opened 2021-05-31
fix(b-table): selected table header text no longer prevents table row…

connorf25:
… selection

### Describe the PR

When b-table is in select-mode="range", and a row is already selected, and some text in the header is selected; any tbody row click events fail to update the selected rows. This fix only passes the tbody to the textSelectionActive function rather than the whole table to prevent this unexpected behaviour (and vice versa for the table head).

Steps for reproduction:
1. Go to [https://bootstrap-vue.org/docs/components/table](url) and navigate to the 'Row select support' section
2. In the interactive table dropdown, set the selection mode to range
3. Click on any row in the table's body to select the row
4. Highlight some text in the table header (thead)
5. Try to select another row in the table body, the click event will fail to propagate due to the textSelectionActive function returning true with the text in the header selected

### PR checklist

- [ ] Other (please describe)

- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "loader-utils": "^2.0.0",
     "lodash": "^4.17.21",
     "marked": "^2.0.3",
-    "nuxt": "^2.15.5",
+    "nuxt": "^2.15.6",
     "postcss": "^8.2.15",
     "postcss-cli": "^8.3.1",
     "prettier": "1.14.3",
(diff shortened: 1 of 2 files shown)


connorf25:
One thing I just noticed when testing is that the sticky-header functionality appears to be broken on the dev branch, however it appears that this was broken before my code changes.

PR #6748, opened 2021-08-18
perf(icon) Speed up slow icons module import

slaweet:
#### What was the problem?
I observed that doing a simple import of bootstrap-vue:

```js
import BootstrapVue from 'bootstrap-vue';
```

takes hundreds of milliseconds (give or take, depending on user's CPU).

I tried to do some profiling and found out that most of the time is spent in `icons` module, specifically in `makeIcon`:

Screenshot from 2021-08-18 15-10-48

When zoomed it, you can see that most of the time is spent in functions `omit` and `Vue.extend`:

Screenshot from 2021-08-18 15-11-24

Each call is just a few milliseconds, but there are hundreds of icons, so it takes hundreds of milliseconds. That's already a noticeable delay that I observed in these two situations:
- starting an app that uses bootstrap-vue (worse score in https://web.dev/vitals/)
- each re-run of tests (run with `jest --watch`) of code that uses bootstrap-vue

#### What is the fix?

Don't call `omit` in each `makeIcon` call, but only once.

This speeds things up roughly by half. The other half would be to somehow avoid repetitive cal of  `Vue.extend`, but I don't think that's possible. At least not as easily as with `omit`.

### PR checklist

- [x] Other (please describe) - performance improvement

- [x] No

### DIFF
--- a/src/icons/helpers/make-icon.js
+++ b/src/icons/helpers/make-icon.js
@@ -3,6 +3,8 @@ import { omit } from '../../utils/object'
 import { kebabCase, pascalCase, trim } from '../../utils/string'
 import { BVIconBase, props as BVIconBaseProps } from './icon-base'
 
+const iconProps = omit(BVIconBaseProps, ['content'])
+
 /**
  * Icon component generator function
  *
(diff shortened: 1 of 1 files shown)


xanf:
Nice and clean, thank you so much @slaweet

PR #6817, opened 2021-10-20
Update README.md

idontbyte:
Update include for bootstrap to use version 4.6 from node_modules in vue-bootstrap because otherwise it is using version 5 from base node_modules folder and breaks


xanf:
@idontbyte Thank you for your PR! Unfortunately we've decided to go with another fix to this problem!

PR #7021, opened 2022-08-01
feat(id) use universally unique ids for the id mixin (fixes #7004)

### LINKED ISSUE #7004: Labels and checkboxes break when using multiple Vue bundles, Vue's internal _uid is not unique enough.
### Describe the bug

Bootstrap Vue assigns IDs like __BVID__27 to link labels to their corresponding input.
This ID is generated by the ID mixin and is based on Vue's internal _uid.

If we have multiple Vue bundles with bootstrap-vue on the same page, they might use overlapping IDs which causes checkboxes and labels to not be interactive anymore, because the `<label for="__BVID__27" />` might point to a __BVID__27 which is also a Modal on the same page.

We get overlapping IDs because on a page with multiple bundles, each Vue instance keeps track of its' own _uid, which is just a counter and can easily collide with the _uids from the other Vue instances on the page.

It is hard to replicate as it requires a page with multiple vue bundles and overlapping BVIDs, but the resulting bug can be approximated by creating a template where 2 b-form-checkboxes have the same id:

https://codesandbox.io/s/bootstrap-vue-duplicate-id-org18h?file=/src/components/HelloWorld.vue:280-497

```
<b-form-checkbox id="__BVID__27" v-model="checked1" switch
  >b-form-checkbox 1</b-form-checkbox
>
<b-form-checkbox id="__BVID__27" v-model="checked2" switch
  >b-form-checkbox 2</b-form-checkbox
>
```

### Expected behavior

I expect that the safeID that bootstrap-vue's id mixin generates is safe, universally unique, and does not rely on Vue's internal _uid.
The author of vue himself says that relying on Vue's internal _uid for unique IDs is bad practice: https://github.com/vuejs/vue/issues/5886#issuecomment-308625735

### Versions

**Libraries:**

- BootstrapVue: 2.22.0
- Bootstrap: 4.6.1
- Vue: 2.6.14

**Environment:**

- Device: Macbook Pro
- OS: macOS
- Browser: Chrome
- Version: [e.g. 70]

### Demo link

https://codesandbox.io/s/bootstrap-vue-duplicate-id-org18h?file=/src/components/HelloWorld.vue:280-497

### Additional context

Obviously the best way to fix this is to simply not have multiple Vue bundles on the same page.
We are however migrating a massive legacy application to Vue, and we are gradually refactoring massive pages to Vue, and it can occur that we get multiple Vue apps on the same page.
This usecase is more common than one might think (google for sharing vuex stores between bundles) 

Bug as found on staging:

duplicate_bvid

stam:
### Describe the PR

Make the id mixin use a universally unique id. Previously vue's internal id was used. 
This is a locally unique id, but when there are multiple vue bundles on the same page it's not globally unique anymore. (#7004)

Because these ids are used to link labels to checkboxes or form elements, using local ids lead to some undefined behavior where checkboxes stopped working because they pointed to an id in use by another input element.

### PR checklist

- [x] Enhancement (augments an existing feature) - `feat(...)`, requires a minor version update
- [ ] Other (please describe)

- [x] No

- [ ] New/updated tests are included and passing (required for new features and enhancements)

description above includes:**

### DIFF
--- a/src/mixins/id.js
+++ b/src/mixins/id.js
@@ -1,7 +1,7 @@
 // SSR safe client-side ID attribute generation
 // ID's can only be generated client-side, after mount
 // `this._uid` is not synched between server and client
-import { COMPONENT_UID_KEY, Vue } from '../vue'
+import { Vue } from '../vue'
 import { PROP_TYPE_STRING } from '../constants/props'
 import { makeProp } from '../utils/props'
 


stam:
Does anyone know why the CI pipeline has been stuck for >2 weeks?

xanf:
@stam thank you for your contribution! Having UUID seems a bit of overkill for me (that will be leeeeeengthy IDs).

Also this will cause problems for tests which are using snapshots - because each time local id will be updated (yeah, we can solve it by writing smart serializer but I prefer not to break existing user experience unless really needed)

I can suggest two approaches, both are fine for me:

* Minimal change

Expose new method `setIdPrefix`. So if you have multiple Vue apps, you can call `setIdPrefix` in each app and solve it

* A bit bigger change

Just making a global counter in the module, which autoincrements each time :)
We could expose extra method `setLocalIdInitialValue` which will explicitly set this counter. This will serve two things: 

* if (for some reason) you will have multiple `bootstrap-vue` instances on the page you could do something like `setLocalIdInitialValue(1000000)` for first one and `setLocalIdInitialValue(2000000)` for second (ugly, but still works :))
* in tests we could call 

```js
beforeEach(() => setLocalIdInitialValue(0));
```

making all snapshot tests work reliably

WDYT?

VividLemon:
> Having UUID seems a bit of overkill for me

This is BV3's implementation:
```ts
/**
 * @param suffix
 * @returns
 */
export default (suffix = ''): string =>
  `__BVID__${Math.random().toString().slice(2, 8)}___BV_${suffix}__`
```
Generally, it's written as a composable to the same affect as the mixin you use.

xanf:
@VividLemon this is an approach I would like to avoid - it will generate totally new IDs on each run :)

VividLemon:
> @VividLemon this is an approach I would like to avoid - it will generate totally new IDs on each run :)

Seems like it would be desirable to avoid conflicts.

stam:
Hey @xanf
Before we work around the constraint of "not breaking snapshot tests", can we first verify that they are currently working?
The current bootstrap-vue behavior uses the incrementing internal ID of vue.
Which means that snapshot tests only succeed if they are being executed in the right order.

If you have snapshot test 1,2,3,4 & 5 and add test 2a, snapshot tests 3, 4 & 5 will break.
If you run test 2 individually, it will break.

I seriously doubt anyone is snapshot testing bootstrap-vue components without passing an id prop.

stam:
This was made 3 years ago, I'm not planning on continuing

